PR #1, opened 2022-02-21
Apply noexcept to appropiate std::move constructors

Skylion007:
I did some brief static analysis with clang-tidy and found some code improvements (a reason I suggested adding basic clang-tidy checks). Mainly it adds noexcepts which can make certain STL constructs able to use their move constructors. It also removed a useless std::move in the move constructor for src/error.cp since the base class std::exception apparently does not have a move constructor.


wjakob:
Wow, the first PR 🎉  -- thanks Aaron!

wjakob:
(I'm going to play with this clang-tidy thing soon, no worries.)

PR #2, opened 2022-02-28
Fix minor noexcept issue

Skylion007:
Just fixes the noexcept annoyance.

### DIFF
--- a/src/internals.h
+++ b/src/internals.h
@@ -166,7 +166,7 @@ inline func_record *nb_func_get(void *o) {
 }
 
 inline void *inst_data(nb_inst *self) {
-    void *ptr = (void *) (self + (intptr_t) self->offset);
+    void *ptr = (void *) (((char *)self) + (intptr_t) self->offset);
     return self->direct ? ptr : *(void **) ptr;
 }
 
--- a/src/internals.h
+++ b/src/internals.h
@@ -166,7 +166,7 @@ inline func_record *nb_func_get(void *o) {
 }
 
 inline void *inst_data(nb_inst *self) {
-    void *ptr = (void *) (((char *)self) + (intptr_t) self->offset);
+    void *ptr = (void *) ((intptr_t) self + self->offset);
     return self->direct ? ptr : *(void **) ptr;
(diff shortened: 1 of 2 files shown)


src/internals.h
```diff
@@ -166,7 +166,7 @@ inline func_record *nb_func_get(void *o) {
 }
 
 inline void *inst_data(nb_inst *self) {
-    void *ptr = (void *) ((intptr_t) self + self->offset);
+    void *ptr = (void *) (self + (intptr_t) self->offset);
```
wjakob, line 169: Hm, am I missing something? To me, `nb_inst*`-typed addition moves by `sizeof(nb_inst)*self->offset` bytes, whereas `uintptr_t`-typed addition moves by `self->offset` bytes.

wjakob: Note that libnanobind does at ton of such pointer manipulations. It is just like CPython code that uses pointer casts between unrelated types and pointer arithmetic all over the place. The traditional concern that one would have in such a situation is strict aliasing. The compiler might assume that two variables don't alias and perform an unsafe optimization.

That's why _nanobind_ is compiled with -fno-strict-aliasing (i.e. just like CPython). It must assume that any two pointers alias.

Skylion007: Well, that's an embarassing error, I suppose if we casted to a char* it would work? 

See https://www.ralfj.de/blog/2020/12/14/provenance.html for why it's an optimization fence.

wjakob: I read this writeup when it came out. AFAIK the discussion of the caveat explained there assumes an aliasing analysis.

Skylion007: Oh sorry, just saw the comment. Yeah, if strict aliasing is disabled this does nothing. I'll just trim the PR to the noexcept fix only.

wjakob: I don't see what is wrong with this `intptr_t` addition, and how a cast to `char *` would improve it. Given that this is basically CPython code, shouldn't it be held by similar standards?

wjakob:
Thanks!

PR #4, opened 2022-03-16
chore: some minor cleanups

henryiii:
This makes a few minor fixes detected and while looking at the code and via
some automated tooling. If this looks good, I'll probably have a few followups,
ending with a pre-commit file that will help keep this clean and tidy. :)

- docs: fix spelling
- chore: clean up some minor stylistic issues
- fix: always include build-backend when using pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,7 @@
 [build-system]
 requires = ["setuptools>=42", "wheel", "scikit-build", "cmake>=3.17", "ninja"]
+build-backend = "setuptools.build_meta"
+
 
 [tool.pytest.ini_options]
 testpaths = [ "tests" ]
--- a/tests/test_classes.py
+++ b/tests/test_classes.py
@@ -130,7 +130,6 @@ def test06_reference_internal(clean):
 def test07_big():
     x = [t.Big() for i in range(1024)]
     x2 = [t.BigAligned() for i in range(1024)]
-    assert x == x2
 
 
 def test08_inheritance():
--- a/tests/test_classes.py
+++ b/tests/test_classes.py
@@ -129,7 +129,8 @@ def test06_reference_internal(clean):
 
 def test07_big():
     x = [t.Big() for i in range(1024)]
-    x = [t.BigAligned() for i in range(1024)]
+    x2 = [t.BigAligned() for i in range(1024)]
+    assert x == x2
 
 
 def test08_inheritance():
@@ -151,20 +152,20 @@ def test09_method_vectorcall():
     def f(a, b, c, d, e):
         out.append((a, b, c, d, e))
 
-    class my_class:
+    class MyClass:
         def f(self, a, b, c, d, e):
             self.out = ((a, b, c, d, e))
 
     t.call_function(f)
 
-    i = my_class()
+    i = MyClass()
     t.call_method(i)
     assert out == [(1, 2, "hello", True, 4)]
     assert i.out == (1, 2, "hello", True, 4)
 
 
 def test10_trampoline(clean):
-    for i in range(10):
+    for _ in range(10):
         class Dachshund(t.Animal):
             def __init__(self):
                 super().__init__()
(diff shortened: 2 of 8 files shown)


tests/test_classes.py
```diff
@@ -129,7 +129,8 @@ def test06_reference_internal(clean):
 
 def test07_big():
     x = [t.Big() for i in range(1024)]
-    x = [t.BigAligned() for i in range(1024)]
+    x2 = [t.BigAligned() for i in range(1024)]
+    assert x == x2
```
henryiii, line 133: FYI, not sure what was intended here. Guessed maybe this?

wjakob:
These all look fine to me, except for the equality comparison thingy. I am not used to the pre-commit workflow, and there is a chance that I won't like it. I am mentioning this in case it's a lot of work for you to set up, in which case that work might be wasted.

Things that I don't want to do that in this project are:

- formatting everything with clang-format (it's fine to provide a .clang-format file for new people contributing code, but I don't want automation for this)
- adding lots of comments throughout the codebase to silence warnings produced by clang-tidy.

wjakob:
(I'm confused why the CI did not run for this PR. I had pressed a button to give permission, but then nothing seems to have happened.)

henryiii:
(GHA is out for a bit - https://www.githubstatus.com)

Not setting up clang-tidy (and that's not something you can put in pre-commit anyway). I'll avoid clang-format (though it works beautifully in pre-commit these days).

tests/test_classes.py
henryiii: Apparently that is not what was intended here. Is this just supposed to run and not check anything?

Skylion007:
@wjakob Happy to setup pre-commit workflow. It's not a lot of work at all.

tests/test_holders.py
```diff
@@ -139,7 +139,7 @@ def test07_uniqueptr_passthrough(clean):
     t.reset()
 
     with pytest.warns(RuntimeWarning, match=r'nanobind::detail::nb_relinquish_ownership()'):
-        with pytest.raises(TypeError) as excinfo:
+        with pytest.raises(TypeError):
             assert t.passthrough_unique(t.Example(1)).value == 1
```
Skylion007, line 143:
```suggestion
with pytest.warns(RuntimeWarning, match=r'nanobind::detail::nb_relinquish_ownership()'), pytest.raises(TypeError):
        assert t.passthrough_unique(t.Example(1)).value == 1
```

henryiii: You have to make the suggestion on the new line, not the deleted line, for it to be automatically apply-able.

henryiii:
I've already got the pre-commit file ready and will add it in a PR or two, but I think @wjakob meant running locally on his machine.

That's also really easy, and very worthwhile, IMO.

You can also set up pre-commit.ci and have it automatically run on all PRs, which is great (once there's a file here).

tests/test_holders.py
Skylion007, line 143:
```suggestion
with pytest.warns(RuntimeWarning, match=r'nanobind::detail::nb_relinquish_ownership()'), pytest.raises(TypeError):
        assert t.passthrough_unique(t.Example(1)).value == 1
```

tests/test_classes.py
henryiii:
```suggestion
```

wjakob (commented):
I left some comments but then never submitted them, sorry.

wjakob:
I will merge this -- still curious about the `build-backend` change.

pyproject.toml
```diff
@@ -1,5 +1,7 @@
 [build-system]
 requires = ["setuptools>=42", "wheel", "scikit-build", "cmake>=3.17", "ninja"]
+build-backend = "setuptools.build_meta"
```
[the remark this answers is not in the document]
henryiii: The default is `build-backend = "setuptools.build_meta:__legacy__"`, which has compatibility hacks to try to support a package that never expected to be build in isolation. Though it's still a different procedure, so it tends to work on both or fail on both. The main user-facing difference that I often see is the normal backend doesn't automatically inject the local directory into sys.path, so you can't import locally in `setup.py` without manually doing that. The reasoning was that importing a local package in `setup.py` was an antipattern and should not be encouraged.

PR #5, opened 2022-03-16
Apply clang-tidy fixes

Skylion007:
Applies various clang-tidy fixes that either follows best practices (ie. initializes pointers to nullptr) or uses modern C++ constructs (like using member initializers that can, reduce the amount of code, increase readability and make some constructors trivial. This also initializes some trivial types likes raw pointers, floats, and booleans since this makes debugging easier and reduces UB.

### DIFF
--- a/include/nanobind/nb_call.h
+++ b/include/nanobind/nb_call.h
@@ -59,7 +59,7 @@ NB_INLINE void call_init(PyObject **args, PyObject *kwnames, size_t &nargs,
         for (size_t i = 0, l = len(value); i < l; ++i)
             args[nargs++] = borrow(value[i]).release().ptr();
     } else if constexpr (std::is_same_v<D, kwargs_proxy>) {
-        PyObject *key, *entry;
+        PyObject *key = nullptr, *entry = nullptr;
         Py_ssize_t pos = 0;
 
         while (PyDict_Next(value.ptr(), &pos, &key, &entry)) {
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -233,7 +233,7 @@ template <typename Type_> struct type_caster_base {
     template <typename T>
     NB_INLINE static handle from_cpp(T &&value, rv_policy policy,
                                      cleanup_list *cleanup) noexcept {
-        Type *value_p;
+        Type *value_p = nullptr;
         if constexpr (is_pointer_v<T>)
             value_p = (Type *) value;
         else
--- a/include/nanobind/nb_error.h
+++ b/include/nanobind/nb_error.h
@@ -22,7 +22,7 @@ public:
     python_error();
     python_error(const python_error &);
     python_error(python_error &&) noexcept;
-    virtual ~python_error();
+    ~python_error() override;
 
     /// Move the error back into the Python domain
     void restore();
@@ -31,7 +31,7 @@ public:
     const handle value() const { return m_value; }
     const handle trace() const { return m_trace; }
 
-    virtual const char *what() const noexcept override;
+    const char *what() const noexcept override;
 
 private:
     object m_type, m_value, m_trace;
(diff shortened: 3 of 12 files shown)


include/nanobind/nb_call.h
```diff
@@ -59,7 +59,7 @@ NB_INLINE void call_init(PyObject **args, PyObject *kwnames, size_t &nargs,
         for (size_t i = 0, l = len(value); i < l; ++i)
             args[nargs++] = borrow(value[i]).release().ptr();
     } else if constexpr (std::is_same_v<D, kwargs_proxy>) {
-        PyObject *key, *entry;
+        PyObject *key = nullptr, *entry = nullptr;
```
wjakob, line 62: I don't think this change is needed. See the `PyDict_Next` documentation -- the loop is only entered when `key` and `entry` are correctly initialized. This change actually has a (small) cost.

include/nanobind/nb_error.h
```diff
@@ -22,7 +22,7 @@ class NB_EXPORT python_error : public std::exception {
     python_error();
     python_error(const python_error &);
     python_error(python_error &&) noexcept;
-    virtual ~python_error();
+    ~python_error() override;
```
wjakob, line 25: fascinating, wasn't aware of that.

include/nanobind/nb_cast.h
```diff
@@ -233,7 +233,7 @@ template <typename Type_> struct type_caster_base {
     template <typename T>
     NB_INLINE static handle from_cpp(T &&value, rv_policy policy,
                                      cleanup_list *cleanup) noexcept {
-        Type *value_p;
+        Type *value_p = nullptr;
```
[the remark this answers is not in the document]
Skylion007: The goal here is to make it more robust if the control flow is later changed so it's not initialized. I believe the compiler is smart enough to optimize the nullptr away in this case.

include/nanobind/nb_lib.h
```diff
@@ -19,10 +19,9 @@ struct NB_CORE cleanup_list {
 public:
     static constexpr uint32_t Small = 6;
 
-    cleanup_list(PyObject *self) {
-        m_size = 1;
-        m_capacity = Small;
-        m_data = m_local;
+    cleanup_list(PyObject *self) : m_size{1},
+		   m_capacity{Small},
```
wjakob, line 23: I am curious -- what's the point of brace init `{` here as opposed to the IMO more readable parenthesis-based init? I use the former mainly for initializing POD `struct` when really necessary (when they don't have a constructor). However, here it would seem that both are equally usable.

Skylion007: Main difference is that it prevents narrowing conversions I believe and is the recommended way to initialize types in C++11 or higher.

src/trampoline.cpp
```diff
@@ -64,8 +64,8 @@ PyObject *trampoline_lookup(void **data, size_t size, const char *name,
     }
 
     // Sill no luck -- perform a lookup and populate the trampoline
-    const char *error;
-    bool is_nb_func;
+    const char *error = nullptr;
```
Skylion007, line 67: I recommend keeping this one since it seems unlikely, but possible to reach goto fail without setting it.

include/nanobind/nb_cast.h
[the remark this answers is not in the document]
wjakob: I would say: let's fix _real_ issues that clang-tidy finds, and stuff like a missing `noexcept` move constructor is in that category. This is not that, it's a rather silly change. I can see how this is a good default at "Google-scale", but I don't see why a small framework like nanobind should be dominated by whatever an automated tool considers to be best practice of the month. It will just add a lot of grind. Perhaps it's just a matter of tweaking the defaults of this tool to get it to relax a bit.

PR #10, opened 2022-03-25
Clang tidy

schaumb:
- Template parameter is never used 
- Use of bitwise '&' with boolean operands
- Declaration uses identifier '__self', which is a reserved identifier
- Avoid repeating the return type from the declaration; use a braced initializer list instead
- Throw expression should throw anonymous temporary values instead
- Use auto when initializing with a cast to avoid duplicating the type name
- Use nullptr
- Narrowing conversion from unsigned to signed type is implementation-defined
- Global variable is never used

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -29,7 +29,7 @@ void raise(const char *fmt, ...) {
     if (size < sizeof(buf))
         throw std::runtime_error(buf);
 
-    auto allocated = std::make_unique<char[]>(size + 1);
+    auto allocated = std::unique_ptr<char[], decltype(&free)>{(char*) malloc(size + 1), &free};
     if (!allocated) {
         fprintf(stderr, "nb::detail::raise(): out of memory!");
         abort();
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -19,29 +19,27 @@ NAMESPACE_BEGIN(detail)
     [[noreturn]]
 #endif
 void raise(const char *fmt, ...) {
-    char buf[512], *ptr = buf;
+    char buf[512];
     va_list args;
 
     va_start(args, fmt);
-    size_t size = vsnprintf(ptr, sizeof(buf), fmt, args);
+    size_t size = vsnprintf(buf, sizeof(buf), fmt, args);
     va_end(args);
 
     if (size < sizeof(buf))
         throw std::runtime_error(buf);
 
-    ptr = (char *) malloc(size + 1);
-    if (!ptr) {
+    auto allocated = std::make_unique<char[]>(size + 1);
+    if (!allocated) {
         fprintf(stderr, "nb::detail::raise(): out of memory!");
         abort();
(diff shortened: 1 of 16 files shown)


src/common.cpp
```diff
@@ -19,29 +19,27 @@ NAMESPACE_BEGIN(detail)
...
-    size_t size = vsnprintf(ptr, sizeof(buf), fmt, args);
+    size_t size = vsnprintf(buf, sizeof(buf), fmt, args);
     va_end(args);
 
     if (size < sizeof(buf))
         throw std::runtime_error(buf);
 
-    ptr = (char *) malloc(size + 1);
-    if (!ptr) {
+    auto allocated = std::make_unique<char[]>(size + 1);
```
wjakob, line 32: This creates a dependence on `operator new`. My plan for the `libnanobind` library was to only use standard C memory allocation, and to eventually route this through `PyMalloc`.

src/nb_type.cpp
```diff
@@ -91,7 +91,7 @@ PyObject *inst_new_impl(PyTypeObject *tp, void *value) {
 
             *(void **) (new_alloc + nb_inst_size) = value;
             self = (nb_inst *) (new_alloc + gc_size);
-            self->offset = (uint32_t) basic_size;
+            self->offset = (int32_t) (uint32_t) basic_size;
```
wjakob, line 94: just do the `int32_t` cast?

src/trampoline.cpp
```diff
@@ -32,8 +32,8 @@ void trampoline_release(void **data, size_t size) noexcept {
         Py_XDECREF(data[i*2 + 2]);
 }
 
-extern char *cur_func;
-extern PyObject *cur_self;
+[[maybe_unused]] extern char *cur_func;
```
wjakob, line 35: delete those lines if really not used.

wjakob:
Looks all good to me, except for a few minor comments.

Skylion007 (approved):
Agree with @wjakob . Looks good and I've been meaning to submit a PR with these slightly more opinionated checks.

wjakob:
Thanks!

wjakob:
Hi @schaumb,

after thinking about it a bit more, I prefer not to have a unnecessary dependence on `<memory>` in the core lib (specifically using `std::unique_ptr`). Same for replacing simple integer casts with `static_cast` everywhere. Good old C is the way to go here, even if considered an anti-pattern by opinionated tools like Clang-Tidy. It's great when they find issues or improvements (like the other parts in this PR), but there are some suggestions where we will have to agree to disagree :).

I amended the commit to reflect this.



-Wenzel

schaumb:
Hi @wjakob !

The current solution not handles, if `std::runtime_error` constructor [may throws](https://en.cppreference.com/w/cpp/error/runtime_error) an exception. In that situation the ptr memory leaked.

PR #12, opened 2022-03-30
Apply more minor clang-tidy fixes

Skylion007:
This PR adds in a couple of fixes up:
1. Simplifies a few new line string literals into chars.
2. Moves default initialization out of the constructors if they are always the same.

### DIFF
--- a/include/nanobind/nb_accessor.h
+++ b/include/nanobind/nb_accessor.h
@@ -16,7 +16,7 @@ public:
 
     template <typename Key>
     accessor(handle obj, Key &&key)
-        : m_base(obj.ptr()), m_cache(nullptr), m_key(std::move(key)) { }
+        : m_base(obj.ptr()),  m_key(std::move(key)) { }
     accessor(const accessor &) = delete;
     accessor(accessor &&) = delete;
     ~accessor() {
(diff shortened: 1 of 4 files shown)


include/nanobind/nb_lib.h
```diff
@@ -46,7 +46,7 @@ struct NB_CORE cleanup_list {
     void expand() noexcept;
 
 protected:
-    uint32_t m_size;
+    uint32_t m_size{1};
```
wjakob, line 49: This change hurts readability. There is only one constructor, and keeping the initialization there is preferable to spreading it out between multiple places.

include/nanobind/nb_accessor.h
```diff
@@ -16,7 +16,7 @@ template <typename Impl> class accessor : public api<accessor<Impl>> {
 
     template <typename Key>
     accessor(handle obj, Key &&key)
-        : m_base(obj.ptr()), m_cache(nullptr), m_key(std::move(key)) { }
```
wjakob, line 19: This change hurts readability. There is only one constructor, and keeping the initialization there is preferable to spreading it out between multiple places.

wjakob:
Thanks! I am fine with moving the `nullptr` initialization to the fields, but it seems detrimental for the `cleanup_list::m_size` case. The change related to string -> char literals looks good.

PR #15, opened 2022-04-05
Update robin-map to v1.0.0

Skylion007:
* Robin-map recently released v1.0.0 in March 25th. Mostly has some CMake fixes to support newer versions and a few minor bugfixes. Full changelog can be found here: https://github.com/Tessil/robin-map/releases/tag/v1.0.0 . Would be good to update the submodule for the minor fixes.


wjakob:
I am subscribed to all PRs/changes in the robin-map repository and will update the submodule reference if there are any changes impacting this project (needlessly updating to stay "up to date" is just churn). We don't use any of the cmake stuff, and the code changes touch on parts that aren't used anywhere by nanobind.

PR #16, opened 2022-04-05
Nit fix ctor for Tensors

Skylion007:
* This makes one of the ctor for Tensors slightly more efficient by using member initializers (the handle won't default initialize it's value to a nullptr).

### DIFF
--- a/include/nanobind/tensor.h
+++ b/include/nanobind/tensor.h
@@ -233,12 +233,10 @@ public:
            const int64_t *strides = nullptr,
            dlpack::dtype dtype = nanobind::dtype<Scalar>(),
            int32_t device_type = device::cpu::value,
-           int32_t device_id = 0) {
-        m_handle =
-            detail::tensor_create(value, ndim, shape, owner.ptr(), strides,
-                                  &dtype, device_type, device_id);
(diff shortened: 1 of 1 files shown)


wjakob:
All supported compilers will be able to elide the default initialization, so the statement on efficiency is incorrect. I honestly prefer the previous version -- the sequencing order is very clear in two subsequent assignments. Whereas read access in an initializer list is on the fringes of the C++ spec -- AFAIK in some C++ versions the order of evaluation is actually tied to the object layout.

Skylion007:
> AFAIK in some C++ versions the order of evaluation is actually tied to the object layout, making such a change bug-prone.
True, but most compilers will issue warnings / errors if the order of evaluation order does not match the object layout. Static analysis tools like clang-tidy also will complain.

Just following the recommendations from https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

wjakob:
It working with "complex" objects where a default constructor will do something that a compiler won't be able to optimize away easily (e.g. a call to `operator new`). However, it's a moot point for scalar values like ints or pointers. I think that these guidelines are useful as a rough pointer, but I don't follow them religiously. In this case, I found the code easier to understand before the change, and the quality of the generated code is not an argument.

PR #21, opened 2022-04-07
add pragma once guards

clonker:
Hey guys,

thank you for the great project! I noticed that some header files are missing include guards, this PR adds them to all non-nb_* include files.

Cheers, Moritz


wjakob:
Good catch, thanks!

Skylion007 (approved):

PR #22, opened 2022-04-07
Implement `dynamic_attr`

inducer:
This was largely done by copy-paste from pybind11. It seems to do what it claims so far.

Still to do:
- [ ] Add some tests.
- [x] Think about whether the leaks are avoidable.
- [ ] Document this? (where?)
- [x] Mop up CI failures

### DIFF
--- a/include/nanobind/nb_class.h
+++ b/include/nanobind/nb_class.h
@@ -57,7 +57,12 @@ enum class type_flags : uint32_t {
     is_arithmetic            = (1 << 15),
 
     /// This type is an arithmetic enumeration
-    has_type_callback        = (1 << 16)
+    has_type_callback        = (1 << 16),
+
+    /// The user requested dynamic attributes. (i.e. a dict in the instance)
+    /// Note that the actual presence of an instance dict will depend on more
+    /// than this flag. (see nb_type_new for details)
+    has_dynamic_attr        = (1 << 17)
 };
 
 struct type_data {
@@ -112,6 +117,10 @@ template <typename T>
 NB_INLINE void type_extra_apply(type_data &t, supplement<T>) {
     static_assert(sizeof(T) <= 0xFF, "Supplement is too big!");
     t.supplement += sizeof(T);
+}
+
+NB_INLINE void type_extra_apply(type_data &t, dynamic_attr) {
+    t.flags |= (uint32_t) type_flags::has_dynamic_attr;
 }
 
 template <typename... Args> struct init {
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -120,6 +120,11 @@ PyObject *inst_new(PyTypeObject *type, PyObject *, PyObject *) {
 
 static void inst_dealloc(PyObject *self) {
     nb_type *type = (nb_type *) Py_TYPE(self);
+
+    PyObject **dict_ptr = _PyObject_GetDictPtr(self);
+    if (dict_ptr)
+        Py_CLEAR(*dict_ptr);
+
     if (type->ht.ht_type.tp_flags & Py_TPFLAGS_HAVE_GC)
         PyObject_GC_UnTrack(self);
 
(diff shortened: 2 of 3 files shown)


wjakob:
What seems to be missing is a `Py_XDECREF` that clears the dictionary when the instance is instance is garbage collected.

inducer:
Thanks for the tip! I'd missed that.

src/nb_type.cpp
```diff
@@ -324,6 +386,26 @@ PyObject *nb_type_new(const type_data *t) noexcept {
...
+                || !base->tp_dictoffset
+                || (base->tp_dictoffset == type->tp_basicsize));
+
+        enable_dynamic_attributes(&nbt->ht);
+    }
+
+    // tp_basicsize is computed above, depending on whether the class has
+    // a dict pointer for dynamic attributes. tp_basicsize can only
+    // grow, not shrink.
+    assert(!has_base || (base->tp_basicsize <= type->tp_basicsize));
```
inducer, line 407: Please take a close look here. I'm using the presence of `base->tp_dictoffset` to decide whether the base class has dynamic attributes and act accordingly. As long as it's C++ types only, I think that's safe. I've tried to be reasonably paranoid, but I don't understand where `base_py` might come from and how it might influence what needs to happen.

Skylion007:
You should test this on 3.11a as I think there were some changes to how the object dicts were initialized and it could affect things there.

src/nb_type.cpp
```diff
@@ -222,6 +227,62 @@ int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
...
+    }
+    PyObject *&dict = *_PyObject_GetDictPtr(self);
+    Py_INCREF(new_dict);
+    Py_CLEAR(dict);
+    dict = new_dict;
+    return 0;
+}
+
+/// dynamic_attr: Allow the garbage collector to traverse the internal instance `__dict__`.
+extern "C" inline int nb_traverse(PyObject *self, visitproc visit, void *arg) {
```
Skylion007, line 254: "Changed in version 3.9: Heap-allocated types are expected to visit Py_TYPE(self) in tp_traverse. In earlier versions of Python, due to [bug 40217](https://bugs.python.org/issue40217), doing this may lead to crashes in subclasses." We should probably fix this in PyBind11 as well: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse

include/nanobind/nb_class.h
```diff
@@ -57,7 +57,12 @@ enum class type_flags : uint32_t {
     is_arithmetic            = (1 << 15),
 
     /// This type is an arithmetic enumeration
-    has_type_callback        = (1 << 16)
+    has_type_callback        = (1 << 16),
+
+    /// The user requested dynamic attributes. (i.e. a dict in the instance)
+    /// Note that the actual presence of an instance dict will depend on more
```
wjakob, line 63: Not sure what lines 2 and 3 of this comment add, the same could be said about all other flags. I'd delete them.

src/nb_type.cpp
```diff
@@ -222,6 +227,62 @@ int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
     return 0;
 }
 
+/// dynamic_attr: Support for `d = instance.__dict__`.
+extern "C" inline PyObject *nb_get_dict(PyObject *self, void *) {
```
wjakob, line 231: You can drop the extern "C" annotation and make the function `static`. Ditto for all the others below.

src/nb_type.cpp
```diff
@@ -222,6 +227,62 @@ int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
...
+    if (!dict)
+        dict = PyDict_New();
+    Py_XINCREF(dict);
+    return dict;
+}
+
+/// dynamic_attr: Support for `instance.__dict__ = dict()`.
+extern "C" inline int nb_set_dict(PyObject *self, PyObject *new_dict, void *) {
+    if (!PyDict_Check(new_dict)) {
+        PyErr_Format(PyExc_TypeError, "__dict__ must be set to a dictionary, not a '%.200s'",
```
wjakob, line 242: I don't get what the `%.200s` means, and why it's there. Thoughts? I realize that this is a copy-paste from pybind11, but that doesn't mean that it should be copied as-is.

src/nb_type.cpp
```diff
@@ -222,6 +227,62 @@ int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
...
+
+/// Give instances of this type a `__dict__` and opt into garbage collection.
+inline void enable_dynamic_attributes(PyHeapTypeObject *heap_type) {
+    // This must be harmless to call if the base class has dynamic attributes
+    // also enabled.
+
+    auto type = &heap_type->ht_type;
+    type->tp_flags |= Py_TPFLAGS_HAVE_GC;
+    type->tp_dictoffset = type->tp_basicsize; // place dict at the end
+    type->tp_basicsize += (Py_ssize_t)sizeof(PyObject *); // and allocate enough space for it
```
wjakob, line 275: This doesn't seem right to me if the base class already has dynamic attributes. Won't we add space for yet another dynamic attribute table?

wjakob: I see now that there is some logic to only call this function selectively based on properties of the base class.
I suggest inlining the code from this function into the caller, and rather only incrementing `tp_basicsize` conditionally. We will need some testcases of classes with dynamic attributes that are child classes of others with dynamic attributes.

wjakob (changes requested):
A bit of feedback on the implementation. What this is also currently missing are tests -- some of them can likely be lifted from pybind11.

wjakob:
Closing this in exchange for PR #38, where I've tried to add the same functionality in a more forward-compatible (aka `Py_LIMITED_API`).

PR #23, opened 2022-04-07
Let `std::string` and `char *` casters also accept `bytes`

inducer:
I don't quite know whether I like this. Ideally, we'd be able to tell from C++ whether something is intended to be UTF-8 or not. (Maybe an argument attribute? An entirely separate type?) Examples where this is relevant include Unix filenames (bytes, not really UTF-8). I encountered it for passing compiler options to OpenCL's `clBuildProgram`, where any implicit encoding change is a risk.

Still to do:
- [ ] Add some tests


wjakob:
Humm, I don't think I like this one -- probably more in the territory of a custom type caster that you could define in your project.

PR #24, opened 2022-04-07
Arithmetic type caster: fall back to casting via `PyNumber_*`

inducer:
This is, again, largely copy-paste from pybind11. The use case is that PyOpenCL often seems things like `numpy.int32` passed as integers, and I'd like it to keep accepting those.

Still to do:
- [ ] Add some tests

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -69,22 +69,45 @@ public:
         const bool convert = flags & (uint8_t) cast_flags::convert;
 
         if constexpr (std::is_floating_point_v<T>) {
-            if (!convert && !PyFloat_Check(src.ptr()))
-                return false;
-            value_p = (Tp) PyFloat_AsDouble(src.ptr());
-        } else {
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @inducer,

I was just looking into this and first wanted to understand how/where the existing number casters are failing. From a quick test (https://github.com/wjakob/nanobind/commit/90d24455315f5c546a9ee80e4183e64ef68ad4e7), it seems to me that things just work and do not need this PR (?). Python itself has changed quite a bit and nanobind requires relatively recent versions that do some implicit conversions themselves in the `PyLong_*` and similar APIs. However, perhaps I am missing something.

Could you take a look?

inducer:
Thanks for investigating! And, weirdly, I distinctly recall needing this to get a test to pass earlier on, but dropping this patch now did not result in any failures. I am not entirely sure why. Let's close this for now. I'll pipe up if/when I rediscover what the issue was. Sorry for the noise!

wjakob:
np!

inducer:
(See #30 for part 2 of the story.)

PR #25, opened 2022-04-08
Add `MANIFEST.in`, vim files to `.gitignore`

inducer:
The point of having `MANIFEST.in` is to let me easily roll packages to use with CI. I don't know how you make your releases, I found it convenient to have.

The other potential benefi of this is that, once it's in, you wouldn't have to keep approving CI on my other PRs. :slightly_smiling_face:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ cmake_install.cmake
 __pycache__
 nanobind.egg-info
 test_*_ext*
+
+*~
+.*.sw[op]
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,13 @@
+recursive-include include *.h
+recursive-include src *.h *.cpp
+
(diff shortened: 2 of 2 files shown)


Skylion007:
@wjakob This would pair well with the corresponding pre-commit hook in #8

wjakob:
I generally am disinclined to anything that requires adding files to the top level of a project. Could you explain why a `MANIFEST.in` file is needed?

FWIW This project provides a `setup.py` file, which is the official way to create packages. And that already includes the information of which files to include. Usually you would not call this file directly, but through pip, e.g.

```
python -m pip wheel .
```
(this is what I upload to PyPI)

AFAIK the following also works
```
python setup.py sdist
```

wjakob:
Regarding the second part: I don't want to collect ignore directives for editor out there (I am saying that as a VIM user myself -- you can easily configure VIM so that it stores these files elsewhere).

PR #32, opened 2022-04-29
Partial PyPy support

wjakob:
This PR introduces partial support for PyPy. There are still crashing testcases that may point to larger issues.

- PyPy seems to have issues with PY_VECTORCALL_ARGUMENTS_OFFSET, it is not able to perform vector calls when given this flag (workaround implemented).

- PyPy does not intern keyword names passed to C API vector calls, so a more costly comparison is needed (workaround implemented).

- nanobind co-locates binding-specific data with the Python type object. When PyPy subclasses a nanobind class, it does not allocate enough memory to store this additional information, which causes crashes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,8 @@ endif()
 # ---------------------------------------------------------------------------
 
 set(Python_FIND_FRAMEWORK LAST) # Prefer Brew/Conda to Apple framework python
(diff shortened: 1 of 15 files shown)


wjakob:
(Giving up on PyPy for now until remaining `Py_LIMITED_API` subtleties are all worked out and PyPy has caught up with these APIs.)

PR #36, opened 2022-05-17
modify __all__ in init.py to suppress linter warning

kelvin34501:


wjakob:
Great catch, thank you!

PR #44, opened 2022-06-17
fix NB_STATIC library target name for no STABLE_ABI

kelvin34501:
The target name seems to be not correct when building nanobind as static library.


wjakob:
Good catch, thanks!

PR #55, opened 2022-07-26
Add an install target

lpapp:


wjakob:
Hi @lpapp ,

after thinking about it (for a long time, sorry!) I decided to close the PR. The official way to install nanobind is through Pip or Conda. Custom solutions like what you propose here are better handed in a specific project, or via a fork.

Wenzel

Issue #42, opened 2022-06-09
[BUG]: Cannot resolve numpy array as tensor input if flag `write=False` is set


chrisrichardson:
### Problem description

If we have a function which takes a const reference, it should be possible to do something like this:

`m.def("get_dim", [](const nb::tensor<const double>& x){ return x.ndim(); });`

and send in an unwriteable numpy array:

```
    w = np.array([[2.0,2.0]])
    w.setflags(write=False)
    assert (module.get_dim(w) == 2)
```

But this doesn't seem to work at the moment:
```
E       TypeError: get_dim(): incompatible function arguments. The following argument types are supported:
E           1. get_dim(arg: tensor[dtype=float64], /) -> int
E       
E       Invoked with types: ndarray
```

Removing the `write=False`, it works fine, but sometimes we have an unwriteable array from somewhere else...

brettc:
I think I have the same problem. I'm converting code from pybind11, and I pass in a numpy array which is constructed from a buffer:

```python
buf = fstream.read(RAW_CHUNK_SIZE)
arr = np.frombuffer(buf, dtype=np.int8)
```
Passing `arr` into a function with just `nb::tensor<>` as the parameter gives me the same error as above, I'm guessing because `frombuffer` makes it not writeable.

brettc:
I spent a bit of time futzing around trying to debug this. 

https://github.com/wjakob/nanobind/blob/8555cf8d22336fc0008c4a21fe708723ad0a4721/src/tensor.cpp#L152

  if (PyObject_GetBuffer(o, view.get(), PyBUF_RECORDS)) {

brettc:
It appears the readonly problem is not just with nanobind.

```python
import numpy as np
w = np.array([[2.0,2.0]])
w.setflags(write=False)
w.__dlpack__()
```
gives
```
TypeError: NumPy currently only supports dlpack for writeable arrays
```

Some discussion here: https://github.com/numpy/numpy/issues/20742

So is the issue that dlpack doesn't support readonly buffers?

whateverforever:
Relevant when working with [mikedh/trimesh](https://github.com/mikedh/trimesh), as the vertex and normal arrays have write=False by default

wjakob:
I will close this issue, the current `master` version of nanobind addresses it to my knowledge. If not, feel free to reopen.

PR #56, opened 2022-07-28
Draft: force buffer protocol in tensor creation for `numpy` to avoid read/write constraints with `dlpack`.

brettc:
This PR overcomes two current limitations with tensors in nanobind when using numpy (it does not affect torch or jax).
* returning tensors from C++ always results in read-only arrays. See #46.
* passing read-only arrays from python results in an error. See #42.

- [x] Allow returning writeable arrays from c++ to python. 
- [x] Allow passing read-only arrays to from python to c++.
- [x] clean up tests.
- [ ] Add documentation.

### DIFF
--- a/include/nanobind/nb_lib.h
+++ b/include/nanobind/nb_lib.h
@@ -349,7 +349,7 @@ NB_CORE dlpack::tensor *tensor_inc_ref(tensor_handle *) noexcept;
 NB_CORE void tensor_dec_ref(tensor_handle *) noexcept;
 
 /// Wrap a tensor_handle* into a PyCapsule
-NB_CORE PyObject *tensor_wrap(tensor_handle *, int framework) noexcept;
+NB_CORE PyObject *tensor_wrap(tensor_handle *, int framework, bool writeable) noexcept;
 
 // ========================================================================
 
--- a/include/nanobind/tensor.h
+++ b/include/nanobind/tensor.h
@@ -76,6 +76,7 @@ struct tensorflow { };
 struct pytorch { };
 struct jax { };
 struct writeable { };
+struct readonly { };
 
 template <typename T> constexpr dlpack::dtype dtype() {
     static_assert(
@@ -109,6 +110,7 @@ struct tensor_req {
     size_t *shape = nullptr;
     bool req_shape = false;
     bool req_dtype = false;
+    bool readonly = false;
     char req_order = '\0';
     uint8_t req_device = 0;
 };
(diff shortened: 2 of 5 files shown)


brettc:
@wjakob I've added an `nb::writeable` flag in tensor creation so far. Let me know if this is on track.

qnzhou:
I can confirm that `nb::writeable` works in my simple test cases.

brettc:
There is now a ```readonly``` flag that allows read-only numpy arrays to be passed through. It forces the translation through the buffer protocol, and has a static-assert guard so that it should only be combined with numpy.

Both issues are solved, but I wonder if there is a way of combining the readonly/writeable flags. I could not see how to do it as they really serve in different places. For clarity, maybe they should be ```accepts_readonly``` and ```returns_writeable```.

qnzhou:
Based on the table below, we only need buffer protocal for 2 cases: python readonly to C++ readonly, and C++ readwrite to python readwrite.

| C++ | Python | C++ -> Python | Python -> C++ |
|----|---|--|--|
| RO | RO | dlpack | Buffer Protocal |
| RO | RW | invalid | dlpack |
| RW | RO | dlpack | invalid |
| RW | RW | Buffer Protocal | dlpack |
 
It will be great to use the same flag for both cases.  Having both `writable` and `readonly` may leads to inconsistency settings.  It may be helpful to consider `writable` a property of the `nb::tensor` obj rather than the C++ or python buffer it binds to.  This may help resolve your semantic concern.

brettc:
Sorry, I don't understand the table -- it looks misleading. Currently nanobind throws an exception if you send it a RO buffer from python. So neither RO -> RO or RO -> RW from python -> C++ work.

Any readonly flag in C++ (at least for numpy arrays) is simply a warning to the programmer, unless we force the array to be const. Perhaps that is an option, but my c++ templating skills are already stretched!

Maybe @wjakob has some suggestions?

qnzhou:
> Sorry, I don't understand the table -- it looks misleading. Currently nanobind throws an exception if you send it a RO buffer from python. So neither RO -> RO or RO -> RW from python -> C++ work.

Sorry, I should clarify the table is a summary of the data transfer technologies that (I think) can be used for transferring data under different read/write permissions.  (and by python, I meant numpy array.)  It is not what nanobind does currently.  

In any case, I agree with you that we should combine `nb::writable` and `nb::readonly`.  However, I don't fully understand your comment that "they really serve in different places".  In my view, they are simply describing the read/write permission of the buffer associated with `nb::tensor` object regardless of whether it is from C++ to python or the other way around.

wjakob:
Sorry for taking a long time to respond -- generally this seems like a reasonable set of changes, though it does seem somewhat specific to NumPy. A similar readonly flag is on the horizon for DLPack (https://github.com/data-apis/array-api/issues/191), so whatever this turns into will have to generalize into passing all kinds of read-only tensors made using different frameworks. 

A code style comment: in nanobind, opening braces are generally merged with the previous line -- the PR deviates from this convention.

brettc:
Hi @wjakob. Thanks for the comments. I'll have a look at the DLPack changes, to try and see how they would fit.

wjakob:
I will close this PR, I don't think it is relevant anymore with recent changes to nanobind. I am happy to reopen it if you disagree.

PR #62, opened 2022-08-19
Support Python byte strings

brentyi:
Added support for using the `bytes` type as an input or output to bound functions. This would be super helpful for me!


wjakob:
Hi Brent,

this set of changes looks good to me, but it will need test coverage.

Thanks,
Wenzel

brentyi:
Added some tests! Open to suggestions for changes/additions.

wjakob:
LGTM, thanks!

PR #64, opened 2022-09-07
Fix tensor creation for zero-dimensional arrays

yosh-matsuda:
I found the following tensor creation for a 0-dim array ([array scalars in numpy](https://numpy.org/doc/stable/reference/arrays.scalars.html)) causes SIGSEGV.
This PR adds checking of array dimension to fix it and accept the creation. Could you please merge this?

```c++
nb::class_<Dummy>(module, "NanobindTest")
    .def(nb::init<>())
    .def(
        "ret_array_scalar", [](const Dummy&) {
            float* f = new float[1]{1};
            size_t shape[0] = {};

            nb::capsule deleter(f,
                                [](void* data) noexcept
                                {
                                    delete[](float*) data;
                                });

            return nb::tensor<nb::numpy, float>(f, 0, shape, deleter);
        });
```

### DIFF
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -505,17 +505,19 @@ tensor_handle *tensor_create(void *value, size_t ndim, const size_t *shape_in,
     for (size_t i = 0; i < ndim; ++i)
         shape[i] = (int64_t) shape_in[i];
 
-    int64_t prod = 1;
-    for (size_t i = ndim - 1; ;) {
-        if (strides_in) {
-            strides[i] = strides_in[i];
-        } else {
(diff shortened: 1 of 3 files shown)


wjakob:
This looks great, but it is missing a testcase for the issue you fixed. Could you please add one?

yosh-matsuda:
@wjakob  OK. Test added.

wjakob:
Seems like you'll have to make some tweaks to the test for it to build on Windows/MSVC.

yosh-matsuda:
Oops. I forgot that zero-size arrays are not C standard.

PR #67, opened 2022-09-21
std::optional and std::variant casters

yosh-matsuda:
The `std::optional` and `std::variant` casters are implemented. This makes it easy to handle overloading of some types that contain `None` as an argument or return value, and also simplifies function signatures. I would be grateful if this would be accepted.

### std::optional

```cpp
m.def("optional_copyable", [](std::optional<Copyable> &) {}, nb::arg("x").none());
m.def("optional_unbound_type", [](std::optional<int> &x) { return x; }, nb::arg("x").none() = nb::none());
```

```python
# optional_copyable(x: Optional[test_stl_ext.Copyable]) -> None
t.optional_copyable(t.Copyable())
t.optional_copyable(None)

# optional_unbound_type(x: Optional[int] = None) -> Optional[int]
t.optional_unbound_type(3)
t.optional_unbound_type(None)
t.optional_unbound_type()
```

In particular, if the argument of a non-bound type such as `nb::list` or `int` can take `None`, the `std::optional` caster can simplify overloading and function signatures.

Note that, receiving as `std::optional` occurs a copy from an input. Using `std::optional` of a pointer type can avoid a copy but it may not very useful since a normal pointer can receive `None` as `nullptr`.

### std::variant

```cpp
m.def("variant_copyable", [](std::variant<Copyable, int> &) {});
m.def("variant_copyable_none", [](std::variant<std::monostate, Copyable, int> &) {}, nb::arg("x").none());
m.def("variant_unbound_type", [](std::variant<std::monostate, nb::list, nb::tuple, int> &x) { return x; }, nb::arg("x").none() = nb::none());
```

```python
# variant_copyable(arg: Union[test_stl_ext.Copyable, int], /) -> None
t.variant_copyable(t.Copyable())
t.variant_copyable(5)

# variant_copyable_none(x: Optional[Union[test_stl_ext.Copyable, int]]) -> None
t.variant_copyable_none(t.Copyable())
t.variant_copyable_none(5)
t.variant_copyable_none(None)

# variant_unbound_type(x: Optional[Union[list, tuple, int]] = None) -> Union[None, list, tuple, int]
t.variant_unbound_type([1])
t.variant_unbound_type((1,2,3))
t.variant_unbound_type(5)
t.variant_unbound_type(None)
t.variant_unbound_type()
```

`std::monostate` is cast as `None`. This allows accepting multiple types, including `None`, by adding `std::monostate` to the argument types of `std::variant`. Multiple types can also be returned by wrapping with `std::variant`.

### DIFF
--- a/include/nanobind/nanobind.h
+++ b/include/nanobind/nanobind.h
@@ -32,6 +32,7 @@
 #include <typeinfo>
 #include <utility>
 #include <new>
+#include <optional>
 
 // Implementation. The nb_*.h files should only be included through nanobind.h
 #include "nb_python.h"
--- a/include/nanobind/nanobind.h
+++ b/include/nanobind/nanobind.h
@@ -32,8 +32,6 @@
 #include <typeinfo>
 #include <utility>
 #include <new>
-#include <optional>
-#include <variant>
 
 // Implementation. The nb_*.h files should only be included through nanobind.h
 #include "nb_python.h"
--- a/include/nanobind/nanobind.h
+++ b/include/nanobind/nanobind.h
@@ -33,6 +33,7 @@
 #include <utility>
 #include <new>
 #include <optional>
+#include <variant>
 
 // Implementation. The nb_*.h files should only be included through nanobind.h
 #include "nb_python.h"
--- a/include/nanobind/nb_traits.h
+++ b/include/nanobind/nb_traits.h
@@ -114,8 +114,16 @@ template <template <typename> typename Op, typename Arg>
 struct detector<std::void_t<Op<Arg>>, Op, Arg>
     : std::true_type { };
 
-template<typename T> struct remove_optional { using type = T; };
-template<typename T> struct remove_optional<std::optional<T>> : remove_optional<T> {};
+template <typename T, typename...>
+struct concat_variant { using type = T; };
+template <typename... Ts1, typename... Ts2, typename... Ts3>
+struct concat_variant<std::variant<Ts1...>, std::variant<Ts2...>, Ts3...>
+    : concat_variant<std::variant<Ts1..., Ts2...>, Ts3...> {};
+
+template<typename T> struct remove_opt_mono { using type = T; };
+template<typename T> struct remove_opt_mono<std::optional<T>> : remove_opt_mono<T> {};
+template <typename... Ts> struct remove_opt_mono<std::variant<Ts...>>
+    : concat_variant<std::conditional_t<std::is_same_v<std::monostate, Ts>, std::variant<>, std::variant<Ts>>...> {};
 
 NAMESPACE_END(detail)
 
@@ -127,6 +135,6 @@ template <template<typename> class Op, typename Arg>
 constexpr bool is_detected_v = detail::detector<void, Op, Arg>::value;
 
 template<typename T>
-using remove_optional_t = typename detail::remove_optional<T>::type;
+using remove_opt_mono_t = typename detail::remove_opt_mono<T>::type;
 
 NAMESPACE_END(NB_NAMESPACE)
--- a/include/nanobind/nb_traits.h
+++ b/include/nanobind/nb_traits.h
@@ -114,6 +114,9 @@ template <template <typename> typename Op, typename Arg>
 struct detector<std::void_t<Op<Arg>>, Op, Arg>
     : std::true_type { };
 
+template<typename T> struct remove_optional { using type = T; };
+template<typename T> struct remove_optional<std::optional<T>> : remove_optional<T> {};
+
 NAMESPACE_END(detail)
 
 template <typename... Args>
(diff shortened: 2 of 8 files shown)


include/nanobind/nanobind.h
```diff
@@ -32,6 +32,8 @@
 #include <typeinfo>
 #include <utility>
 #include <new>
+#include <optional>
```
wjakob, line 35: There is no way that we can include these heavy C++ headers in `nanobind.h` itself. See this excerpt from README.md:

> `#include <pybind11/pybind11.h>` pulls in a large portion of the STL (about 2.1 MiB of headers with Clang and libc++). nanobind minimizes STL usage to avoid this problem. Type casters even for for basic types like std::string require an explicit opt-in by including an extra header file (e.g. `#include <nanobind/stl/string.h>`).

include/nanobind/stl/optional.h
```diff
@@ -0,0 +1,65 @@
...
+    Value value = std::nullopt;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list* cleanup) noexcept {
+        if (src.is_none()) return true;
+
+        Caster caster;
+        if (!caster.from_python(src, flags, cleanup)) return false;
+
+        if constexpr (is_pointer_v<T>)
+        {
```
wjakob, line 30: braces on previous line, ditto for other places in this PR.

include/nanobind/stl/optional.h
```diff
@@ -0,0 +1,65 @@
...
+    static constexpr auto Name = const_name("Optional[") + concat(Caster::Name) + const_name("]");;
+    static constexpr bool IsClass = false;
+
+    template <typename T_>
+    using Cast = movable_cast_t<T_>;
+
+    Value value = std::nullopt;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list* cleanup) noexcept {
+        if (src.is_none()) return true;
```
wjakob, line 24: newline + indentation after condition, ditto for other places in this PR.

include/nanobind/stl/optional.h
```diff
@@ -0,0 +1,65 @@
...
+        if constexpr (is_pointer_v<T>)
+        {
+            static_assert(Caster::IsClass,
+                            "Binding 'optional<T*>' requires that 'T' can also be bound by nanobind.");
+            value = caster.operator cast_t<T>();
+        }
+        else if constexpr (Caster::IsClass)
+        {
+            value = caster.operator cast_t<T&>();
+        }
```
wjakob, line 38: `} else {`

include/nanobind/nb_traits.h
```diff
@@ -114,6 +114,17 @@ template <template <typename> typename Op, typename Arg>
 struct detector<std::void_t<Op<Arg>>, Op, Arg>
     : std::true_type { };
 
+template <typename T, typename...>
```
wjakob, line 117: Having a dependency of a core part of nanobind on `std::variant` is unfortunately an absolute no-go for me, so something bout these traits would have to change. 

Perhaps we can start with some discussion on the PR on what specific issue they address, and how else it could be achieved?

wjakob (changes requested):
HI @yosh-matsuda,

this is great -- I've been thinking that it would be nice to have `std::optional` and `std::variant` casters in the library. Here are a few first (superficial) comments before a thorough pass through every line. Most are just about formatting, but one is a major architectural issue that would need to be addressed. Nanobind is designed so that it _barely_ depends on the STL. For example, it provides its own tuple class internally instead of relying on `std::tuple`, due to the costs that including the associated header file adds. So adding 

```cpp
#include <optional>
#include <variant>
```
to `nanobind.h` is in general not an acceptable change for me. But perhaps we can figure out how to make this unnecessary.

include/nanobind/nb_traits.h
yosh-matsuda: I understood. It was rough to include two headers globally just for signature string generation. Actually, I also had trouble with this part, and the reason why `type traits` depends on STL was to avoid changing the design of the type caster.

I pushed a new commit to discuss implementation policy. Static member variables `ArgName` are added to optional and variant casters, and meta-classes are added to `nb_func.h`. The dependency on STL header has been removed, but it might look a bit messy. What do you think?

yosh-matsuda: I noticed that I should have put the template specialization of `remove_opt_mono` I added into each header. How about this change? (0142b12)

include/nanobind/nb_traits.h
```diff
@@ -114,6 +114,9 @@ template <template <typename> typename Op, typename Arg>
 struct detector<std::void_t<Op<Arg>>, Op, Arg>
     : std::true_type { };
 
+// Specialized in stl/optional.h and stl/variant.h
```
wjakob, line 117: Improve comment for posterity:
```
/* This template is used for docstring generation and specialized in
   ``stl/{variant,optional.h}`` to strip away std::optional and
   ``std::variant<std::monostate>`` in top-level argument types and
   avoid redundancy when combined with nb::arg(...).none(). */
```

Also, note: space between `template` and `<typename` matching other lines.

wjakob (commented):
This looks good to me now. Just a small remaining change request.

yosh-matsuda:
Thanks for the review. Conflicts have been resolved.

wjakob:
Merged 🎉  - thanks for your hard work on this!

PR #71, opened 2022-10-02
Custom exceptions

inducer:
Generally seems to work based on my use of this in https://github.com/inducer/pyopencl/pull/546.

- [x] Needs tests


inducer:
Added some tests, ready for a look.

wjakob:
Thanks for looking into this. I ended up implementing something equivalent in a slightly more nanobind-style fashion (minimize templated part and move static bits into the library). We can also avoid the ugly static variables within functions or at the global level that were needed for pybind11.

The syntax is similar, but not exactly the same as pybind11/your proposal. Please let me know it it works for you.

https://github.com/wjakob/nanobind/commit/41b7da33f1bc5c583bb98df66bdac2a058ec5c15

inducer:
Thanks! There's a slight inelegance in that I have to register my own translators (overriding the implicitly created ones) because I would like to set the exception with an object, not a string, but I can live with that. (That also brings back the static objects, but oh well.)

wjakob:
If you have an idea for a better design, then I'm all ears. This is still an experimental library, so the API could be changed if warranted by an improvement.

PR #72, opened 2022-10-02
Add (failing) test for `__eq__`-in-baseclass

inducer:
This is a bug report masquerading as a PR. With the code changes described below, there is a test failure
```
  File "/home/andreas/pack/nanobind/build/tests/test_classes.py", line 141, in test08_inheritance
    assert not dog == cat
RecursionError: maximum recursion depth exceeded while calling a Python object
```
I *believe* this is a regression, I recall code like this working at a point with nanobind. If it's important, I can try to bisect.


inducer:
Thanks for looking into this, and for finding a workaround!

PR #74, opened 2022-10-03
Minor CMake Update

qnzhou:
Summary:
* Avoid re-finding python if its CMake target already exists.  This is useful when the client's cmake already have python setup and re-finding python may result in error.


lrineau (approved):

wjakob:
Looks great!

PR #84, opened 2022-10-13
Add std::unordered_set type casters and corresponding tests

nepluno:
This PR adds `std::unordered_set` type casters that converts from/to Python's `set`. Similar to the current implementation for `list` and `dict`, the common logic is in `detail/nb_set.h`.

Simple tests have been added to `test_functions_ext` and `test_stl_ext`.

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -348,4 +348,16 @@ template <typename T> void list::append(T &&value) {
         detail::raise_python_error();
 }
 
+template <typename T>
+bool anyset::contains(T&& value) const
+{
+    return PySet_Contains(m_ptr, nanobind::cast(value).ptr()) == 1;
+}
+
+template <typename T>
+bool set::add(T&& value)
+{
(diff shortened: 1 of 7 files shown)


wjakob:
Two quick comments just from a brief pass over the PR:

- Could you split this into two separate contributions? Part 1 would be the caster for `std::[unordered_]set`.  Part 2 would be the wrappers for Python objects.

I think that part 1 is relatively unobjectionable and can go in quickly. Part 2 might need a bit more discussion. For example, the current way that exception handling is done is not right (it's e.g. not enough to check that `PySet_Contains` returns `1`. In the negative case, Python has an error status that must be converted into a C++ exception). The file `nb_common.cpp` has many wrappers that perform such logic in the library part of nanobind.

Also, minor: nanobind doesn't place braces on the new line -- I noticed that in a few places.

nepluno:
> * Could you split this into two separate contributions? Part 1 would be the caster for `std::[unordered_]set`.  Part 2 would be the wrappers for Python objects.

Sure. I've pushed PR #87. Closing the current one.

PR #85, opened 2022-10-13
Fix dict type caster compilation with `const &`

mferencevic:
The dictionary caster that I wrote has a compilation issue when the `std::map` is passed into the C++ caster as a `const &`. This change fixes it. I would also note that the same issue should probably be fixed in #84 as the conversion logic was modified from the dictionary caster.

```
nanobind/include/nanobind/stl/detail/nb_dict.h:57:43: error: type 'const std::map<std::__cxx11::basic_string<char>, std::vector<unsigned long, std::allocator<unsigned long>>, std::less<std::__cxx11::basic_string<char>>, std::allocator<std::pair<const std::__cxx11::basic_string<char>, std::vector<unsigned long, std::allocator<unsigned long>>>>> &' cannot be used prior to '::' because it has no members
                    forward_like<typename T::key_type>(item.first), policy, cleanup));
                                          ^
```

### DIFF
--- a/include/nanobind/stl/detail/nb_dict.h
+++ b/include/nanobind/stl/detail/nb_dict.h
@@ -54,9 +54,9 @@ template <typename Value_, typename Key, typename Element> struct dict_caster {
         if (ret) {
             for (auto &item : src) {
                 object k = steal(KeyCaster::from_cpp(
-                    forward_like<typename T::key_type>(item.first), policy, cleanup));
+                    forward_like<T>(item.first), policy, cleanup));
                 object e = steal(ElementCaster::from_cpp(
-                    forward_like<typename T::mapped_type>(item.second), policy, cleanup));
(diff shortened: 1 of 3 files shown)


wjakob:
Ah wait, sorry -- I forgot that #84 is not merged yet. No need to do anything about that commit then.

mferencevic:
Sure thing, I'm working on a test and I'm also investigating a potential `segfault` so I'll keep you posted.

wjakob:
That looks good. Shall I merge this, or are you still investigating the segfault? What was that, out of curiosity?

mferencevic:
I've investigated the segfault. At first it seemed that it could have something to do with the map/dict conversion logic, but I think that the logic is correct. The culprit ended up being the default conversion of a very complex `union` type (that was incidentally stored within a `map`).

wjakob:
Thanks!

PR #87, opened 2022-10-13
add support to unordered_set

nepluno:
This PR adds type castor for `std::unordered_set`, and is part of #84, which is split into two parts. Here we don't do the same for `std::set` since Python doesn't have an ordered set by default.

### DIFF
--- a/include/nanobind/stl/detail/nb_set.h
+++ b/include/nanobind/stl/detail/nb_set.h
@@ -15,7 +15,7 @@ template <typename Value_, typename Key> struct set_caster {
 
         PyObject* iter = obj_iter(src.ptr());
         if (iter == nullptr) {
-            PyErr_Clear();
+            PyErr_Print();
             return false;
         }
 
@@ -23,13 +23,17 @@ template <typename Value_, typename Key> struct set_caster {
         bool success = (size >= 0);
 
         KeyCaster key_caster;
-        while (PyObject* key = obj_iter_next(iter)) {
+        for (PyObject* key = PyIter_Next(iter); key; key = PyIter_Next(iter)) {
             if (!key_caster.from_python(key, flags, cleanup)) {
                 success = false;
                 break;
             }
             value.emplace(((KeyCaster &&) key_caster).operator cast_t<Key&&>());
             Py_DECREF(key);
+        }
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+            success = false;
         }
 
         Py_DECREF(iter);
@@ -44,9 +48,16 @@ template <typename Value_, typename Key> struct set_caster {
             for (auto& key : src) {
                 object k = steal(
                     KeyCaster::from_cpp(forward_like<T>(key), policy, cleanup));
-                if (PySet_Add(ret.ptr(), k.ptr()) != 0) return handle();
+                if (PySet_Add(ret.ptr(), k.ptr()) != 0) {
+                    if (PyErr_Occurred()) {
+                        PyErr_Print();
+                    }
+                    return handle();
+                }
                 if (!k.is_valid()) return handle();
             }
+        } else if (PyErr_Occurred()) {
+            PyErr_Print();
         }
         return ret.release();
     }
--- a/include/nanobind/stl/detail/nb_set.h
+++ b/include/nanobind/stl/detail/nb_set.h
@@ -15,16 +15,16 @@ template <typename Value_, typename Key> struct set_caster {
 
         PyObject* iter = obj_iter(src.ptr());
         if (iter == nullptr) {
-            PyErr_Print();
+            PyErr_Clear();
             return false;
         }
 
-        Py_ssize_t size = NB_SET_GET_SIZE(src.ptr());
-        bool success = (size >= 0);
+        bool success = true;
 
         KeyCaster key_caster;
         for (PyObject* key = PyIter_Next(iter); key; key = PyIter_Next(iter)) {
             if (!key_caster.from_python(key, flags, cleanup)) {
+                Py_DECREF(key);
                 success = false;
                 break;
             }
@@ -32,7 +32,7 @@ template <typename Value_, typename Key> struct set_caster {
             Py_DECREF(key);
         }
         if (PyErr_Occurred()) {
-            PyErr_Print();
+            PyErr_Clear();
             success = false;
         }
 
@@ -50,14 +50,14 @@ template <typename Value_, typename Key> struct set_caster {
                     KeyCaster::from_cpp(forward_like<T>(key), policy, cleanup));
                 if (PySet_Add(ret.ptr(), k.ptr()) != 0) {
                     if (PyErr_Occurred()) {
-                        PyErr_Print();
+                        PyErr_Clear();
                     }
                     return handle();
                 }
                 if (!k.is_valid()) return handle();
             }
         } else if (PyErr_Occurred()) {
-            PyErr_Print();
+            PyErr_Clear();
         }
         return ret.release();
     }
--- a/include/nanobind/stl/detail/nb_set.h
+++ b/include/nanobind/stl/detail/nb_set.h
@@ -1,0 +1,56 @@
+#pragma once
+
+#include <nanobind/nanobind.h>
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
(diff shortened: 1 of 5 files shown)


include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,56 @@
...
+        if (iter == nullptr) {
+            PyErr_Clear();
+            return false;
+        }
+
+        Py_ssize_t size = NB_SET_GET_SIZE(src.ptr());
+        bool success = (size >= 0);
+
+        KeyCaster key_caster;
+        while (PyObject* key = obj_iter_next(iter)) {
```
wjakob, line 26: potential issue: `obj_iter_next` can raise, but `from_python` cannot. We need a non-throwing version of that function which indicates a failure (and clears the underlying Python error status) so that the caster can return `false` in this case.

wjakob: minor: is in't it common to put an extra pair of parentheses in a `while` statement when a variable is defined there? Perhaps it's cleaner to use a `for` statement (without upper bound) that has that idea builtin.

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    template <typename T>
+    static handle from_cpp(T &&src, rv_policy policy, cleanup_list *cleanup) {
+        object ret = steal(PySet_New(nullptr));
+        if (ret) {
+            for (auto& key : src) {
+                object k = steal(
+                    KeyCaster::from_cpp(forward_like<T>(key), policy, cleanup));
+                if (PySet_Add(ret.ptr(), k.ptr()) != 0) return handle();
```
wjakob, line 47: `PySet_Add` can fail with an error status set. In that case it's not valid to just return.

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        Py_DECREF(iter);
+
+        return success;
+    }
+
+    template <typename T>
+    static handle from_cpp(T &&src, rv_policy policy, cleanup_list *cleanup) {
+        object ret = steal(PySet_New(nullptr));
```
wjakob, line 42: `PySet_New` can fail with an error status set. In that case it's not valid to just return.

include/nanobind/stl/detail/nb_set.h
nepluno: Thanks! I've changed this to a for-loop. And the error has been checked if `key` is a `nullptr` (exit the for-loop), where we will print & clear the error.

include/nanobind/stl/detail/nb_set.h
nepluno: Thanks! I've added an error printing (and clearing by the way) there.

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,67 @@
...
+            return false;
+        }
+
+        Py_ssize_t size = NB_SET_GET_SIZE(src.ptr());
+        bool success = (size >= 0);
+
+        KeyCaster key_caster;
+        for (PyObject* key = PyIter_Next(iter); key; key = PyIter_Next(iter)) {
+            if (!key_caster.from_python(key, flags, cleanup)) {
+                success = false;
```
wjakob, line 28: missing DECREF in the case of a failure?

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,67 @@
...
+    NB_TYPE_CASTER(Value_, const_name("set[") + make_caster<Key>::Name + const_name("]"));
+
+    using KeyCaster = make_caster<Key>;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
+        value.clear();
+
+        PyObject* iter = obj_iter(src.ptr());
+        if (iter == nullptr) {
+            PyErr_Print();
```
wjakob, line 18: `PyErr_Clear()`, not `PyErr_Print()`. The type casters should not print messages in the case of failure.

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,67 @@
...
+            return false;
+        }
+
+        Py_ssize_t size = NB_SET_GET_SIZE(src.ptr());
```
wjakob, line 22: Looks to me like `size` isn't actually used ..

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,67 @@
...
+    template <typename T>
+    static handle from_cpp(T &&src, rv_policy policy, cleanup_list *cleanup) {
+        object ret = steal(PySet_New(nullptr));
+        if (ret) {
+            for (auto& key : src) {
+                object k = steal(
+                    KeyCaster::from_cpp(forward_like<T>(key), policy, cleanup));
+                if (PySet_Add(ret.ptr(), k.ptr()) != 0) {
+                    if (PyErr_Occurred()) {
+                        PyErr_Print();
```
wjakob, line 53: ditto

include/nanobind/stl/detail/nb_set.h
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                    return handle();
+                }
+                if (!k.is_valid()) return handle();
+            }
+        } else if (PyErr_Occurred()) {
+            PyErr_Print();
```
wjakob, line 60: ditto

include/nanobind/stl/unordered_set.h
```diff
@@ -0,0 +1,16 @@
...
+#include "detail/nb_set.h"
+#include <unordered_set>
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+template <typename Key, typename Hash, typename Compare, typename Alloc>
+struct type_caster<std::unordered_set<Key, Hash, Compare, Alloc>>
+    : set_caster<std::unordered_set<Key, Hash, Compare, Alloc>, Key>
+{
```
wjakob, line 12: brace on previous line

wjakob:
I left a few more comments

include/nanobind/stl/detail/nb_set.h
nepluno: Added. Thanks!

include/nanobind/stl/detail/nb_set.h
nepluno: Replaced. Thanks.

include/nanobind/stl/detail/nb_set.h
nepluno: Removed. Thanks.

include/nanobind/stl/detail/nb_set.h
nepluno: > I left a few more comments

Thanks! I've fixed correspondingly.

nepluno:
> I left a few more comments

Thanks! I've fixed correspondingly.

wjakob:
Hi @nepluno ,

I just took another look. is it possible that you forgot to stage/commit the python portion of the tests? I see C++ test functions `set_in_value`, `set_in_lvalue_ref`, and `set_in_rvalue_ref`, but those are never called from the Python test suite.

Finally, would you be able to add a test that corresponds to a type caster failure? For example, trying to pass a set containing a string to a function taking a `std::unordered_map<int>`. (Doesn't have to be exactly that, the important thing is just that we check that nanobind then refuses to run the associated function overload.)

Thanks,
Wenzel

nepluno:
> is it possible that you forgot to stage/commit the python portion of the tests?

Hi @wjakob! Sorry for the missing tests and thanks for the reminder! I've added them.

> would you be able to add a test that corresponds to a type caster failure?

Done. I've added a test using `unordered_set<int>` as the input to a function expecting `unordered_set<string>`.

wjakob:
Thanks!

wjakob:
I committed a few minor tweaks to the code in commit b26338fed6fc69d4a11795a610b460d82e711a26 and also added bindings for `std::set`. I don't think the different ordering convention is a big deal, and the same also applies to `std::map`. Python dict orders entries via the original insertion order, whereas the C++ one sorts by key. 🤷

PR #91, opened 2022-10-17
Change return type of tensor::data() to Scalar, use consistent spelling

froody:
When mapping the data (e.g. via Eigen::Map) a cast is needed to convert the output of data() to the correct type. Since this is always deterministic I think the return type of data should be changed.


wjakob:
LGTM, thanks!

PR #98, opened 2022-11-04
PEP 585 generics in type names

yosh-matsuda:
This is a very minor change: this PR unifies the type names of collection types appearing in STL casters, depending on the Python version.  
According to the [PEP 585](https://peps.python.org/pep-0585/), Python 3.9 or later allows built-in collection types to be used for type annotations.

### DIFF
--- a/include/nanobind/nb_defs.h
+++ b/include/nanobind/nb_defs.h
@@ -86,8 +86,16 @@
 
 #if PY_VERSION_HEX < 0x03090000
 #  define NB_INTERPRETER_STATE_GET _PyInterpreterState_Get
+#  define NB_TYPING_DICT "Dict"
+#  define NB_TYPING_SET "Set"
+#  define NB_TYPING_TUPLE "Tuple"
+#  define NB_TYPING_TYPE "Type"
 #else
 #  define NB_INTERPRETER_STATE_GET PyInterpreterState_Get
+#  define NB_TYPING_DICT "dict"
+#  define NB_TYPING_SET "set"
+#  define NB_TYPING_TUPLE "tuple"
+#  define NB_TYPING_TYPE "type"
 #endif
 
 #if defined(Py_LIMITED_API)
--- a/include/nanobind/nb_types.h
+++ b/include/nanobind/nb_types.h
@@ -513,7 +513,7 @@ public:
 
 template <typename T> class type_object_t : public type_object {
 public:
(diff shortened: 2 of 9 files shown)


wjakob:
This is great. One nit from me is that this duplicates quite a lot of lines of code that goes beyond the name of the type annotation.

What do you think about centralizing these in `nb_defs.h` based on the Python version:

```cpp
#if PY_VERSION_HEX < 0x03090000
#  define NB_TYPING_DICT "Dict"
...
#else
#  define NB_TYPING_DICT "dict"
...
#endif
```

with the following code e.g. for ``dict``:
```cpp
    NB_TYPE_CASTER(Value_, const_name(NB_TYPING_DICT) + const_name("[") +
                               make_caster<Key>::Name + const_name(", ") +
                               make_caster<Element>::Name + const_name("]"));
```

Let's not change `Sequence` to `List`: this type is intentionally chosen because the associated bindings accept any sequence type.

yosh-matsuda:
I am actually in the process of trying mypy-stubgen with nanobind. It does not work properly as it is, so some modifications seem to be necessary.

This is a current ad hoc fix: https://github.com/yosh-matsuda/mypy/commit/5fec4fc3dc3f2c55315242bde27ed7e6851dd790

wjakob:
I'm open to making changes to `nanobind` if they help with this. The `is_c_function` part seems tricky, but nanobind could, e.g., use some other way to refer to anonymous functions.

wjakob:
Thanks!

wjakob:
Hi @yosh-matsuda,

> Since the return value is only a list, I wonder if there is a good way to have a different "Name" for the argument and return type, but it seems difficult.

I tried this just now, and it is (IMO) introducing too much code.

Furthermore, after merging this commit, I changed my mind about the `Sequence[]` type annotation and committed a change that moves everything towards more "concrete" types.  The commit message explains the rationale going forward: https://github.com/wjakob/nanobind/commit/c855c90fc91d180f7c904c612766af6a84c017e3

I hope that makes sense.

yosh-matsuda:
@wjakob Thanks. This change looks good to me because it is consistent between array and map casters, and also from a readability standpoint.

> For those that need a signature to be ultra-correct, there is still nb::raw_doc() as an escape hatch.

Yes! We can also use it for other special cases (e.g. `args` annotations). Thanks for letting me know.

PR #99, opened 2022-12-01
Fix dangling tp_members pointer.

hawkinsp:
In a heap-allocated type, tp_members normally points to the interior of the type object, at offset tp_basicsize. By copying the tp_members field of a temporary heap-allocated type object to the final type object, we leave a dangling pointer. Instead, allocate space for the PyMemberDef objects at the end of the nanobind-allocated type object, copy them into place, and set tp_members to point to that location.

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -484,7 +484,7 @@ PyObject *nb_type_new(const type_data *t) noexcept {
 
     const char *tp_name = PyUnicode_AsUTF8AndSize(temp_ht->ht_name, nullptr);
 
-    PyObject *result = PyType_GenericAlloc(metaclass, 0);
+    PyObject *result = PyType_GenericAlloc(metaclass, Py_SIZE(temp_tp));
     if (!temp || !result)
         fail("nanobind::detail::nb_type_new(\"%s\"): type construction failed!",
              t->name);
@@ -533,7 +533,6 @@ PyObject *nb_type_new(const type_data *t) noexcept {
     COPY_FIELD(tp_iter);
     COPY_FIELD(tp_iternext);
     COPY_FIELD(tp_methods);
-    COPY_FIELD(tp_members);
     COPY_FIELD(tp_getset);
     COPY_FIELD(tp_base);
     COPY_FIELD(tp_descr_get);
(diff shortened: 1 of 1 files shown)


src/nb_type.cpp
```diff
@@ -550,6 +551,12 @@ PyObject *nb_type_new(const type_data *t) noexcept {
 
     #undef COPY_FIELD
 
+    tp->tp_members = nullptr;
```
wjakob, line 554: Very minor: the `nullptr` assignment isn't needed as the output of `PyType_GenericAlloc` is zero-initialized.

src/nb_type.cpp
```diff
@@ -484,7 +486,7 @@ PyObject *nb_type_new(const type_data *t) noexcept {
 
     const char *tp_name = PyUnicode_AsUTF8AndSize(temp_ht->ht_name, nullptr);
 
-    PyObject *result = PyType_GenericAlloc(metaclass, 0);
+    PyObject *result = PyType_GenericAlloc(metaclass, num_members ? num_members + 1 : 0);
```
wjakob, line 489: I feel like this is missing some kind of adaptation to the actual number of members in `temp_tp`, which was created by `PyType_FromSpec`. What if that process adds further members? Copying `num_members + 1` elements then would not produce a copy with a null-initialized last (sentinel) element.

wjakob:
Looks like `PyHeapType_GET_MEMBERS` was removed in Python 3.11. PyPy also lacks that definition.

src/nb_type.cpp
hawkinsp: I realized there's a better way to do this, which is to copy `Py_SIZE(x) * tp_itemsize` bytes.

src/nb_type.cpp
```diff
@@ -484,7 +486,7 @@ PyObject *nb_type_new(const type_data *t) noexcept {
 
     const char *tp_name = PyUnicode_AsUTF8AndSize(temp_ht->ht_name, nullptr);
 
-    PyObject *result = PyType_GenericAlloc(metaclass, 0);
+    PyObject *result = PyType_GenericAlloc(metaclass, num_members);
```
wjakob, line 489: Did you forget to replace `num_members` by `Py_SIZE(temp_tp)` here, or am I missing something?

hawkinsp: Yeah I guess that's better. Done.

wjakob:
Great, thank you!

PR #100, opened 2022-12-01
Add support for making nanobind classes weak-referenceable.

hawkinsp:
Adds an optional nb::weak_referenceable attribute. If present, we allow space for a weak reference list in the object and set tp_weaklistoffset.

This change incorporates the changes from https://github.com/wjakob/nanobind/pull/99

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -19,22 +19,33 @@ NAMESPACE_BEGIN(detail)
 static PyObject **nb_dict_ptr(PyObject *self) {
     PyTypeObject *tp = Py_TYPE(self);
 #if !defined(Py_LIMITED_API)
-    return (PyObject **) ((uint8_t *) self + tp->tp_dictoffset);
+    Py_ssize_t dictoffset = tp->tp_dictoffset;
 #else
-    return (PyObject **) ((uint8_t *) self + nb_type_data(tp)->dictoffset);
+    Py_ssize_t dictoffset = nb_type_data(tp)->dictoffset;
+#endif
+    return dictoffset ? (PyObject **) ((uint8_t *) self + dictoffset) : nullptr;
+}
+
+static PyObject **nb_weaklist_ptr(PyObject *self) {
+    PyTypeObject *tp = Py_TYPE(self);
+#if !defined(Py_LIMITED_API)
+    return (PyObject **) ((uint8_t *) self + tp->tp_weaklistoffset);
+#else
+    return (PyObject **) ((uint8_t *) self + nb_type_data(tp)->weaklistoffset);
 #endif
 }
 
 static int inst_clear(PyObject *self) {
-    PyObject *&dict = *nb_dict_ptr(self);
-    Py_CLEAR(dict);
+    PyObject **dict = nb_dict_ptr(self);
+    if (dict)
+        Py_CLEAR(*dict);
     return 0;
 }
 
 static int inst_traverse(PyObject *self, visitproc visit, void *arg) {
-    PyObject *&dict = *nb_dict_ptr(self);
+    PyObject **dict = nb_dict_ptr(self);
     if (dict)
-        Py_VISIT(dict);
+        Py_VISIT(*dict);
 #if PY_VERSION_HEX >= 0x03090000
     Py_VISIT(Py_TYPE(self));
 #endif
@@ -57,6 +68,7 @@ PyObject *inst_new_impl(PyTypeObject *tp, void *value) {
 
     if (!gc) {
         size_t size = sizeof(nb_inst);
+
         if (!value) {
             // Internal storage: space for the object and padding for alignment
             size += t->size;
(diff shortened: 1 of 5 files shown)


tests/test_classes.py
```diff
@@ -545,3 +546,11 @@ def test31_cycle():
     a = t.Wrapper()
     a.value = a
     del a
+
+
+def test32_weak_references():
+    o = t.StructWithWeakrefs(42)
+    w = weakref.ref(o)
+    assert w() is o
+    del o
```
wjakob, line 555: you will likely need a `collect()` call after `del o` for this to work on PyPy.

src/nb_type.cpp
```diff
@@ -437,6 +449,21 @@ PyObject *nb_type_new(const type_data *t) noexcept {
...
+
+        members[num_members] = PyMemberDef{ "__weaklistoffset__", T_PYSSIZET,
+                                            weaklistoffset, READONLY,
+                                            nullptr };
+        ++num_members;
+
+        spec.basicsize = (int) basicsize;
+    }
+    if (num_members > 0) {
+      *s++ = { Py_tp_members, (void *) members };
```
wjakob, line 464: Minor: fix indentation, remove braces.

wjakob:
Thanks for this nice set of patches @hawkinsp. Before merging, I have one more request for a testcase with a more complex configuration: A type that is weak-referencable *and* has a per-instance dictionary (to make sure the layout is correct). A variation of that test could subclass the type in Python, deposit a value in the dictionary, and then test weak-referenceability.

src/nb_type.cpp
```diff
@@ -418,17 +434,43 @@ PyObject *nb_type_new(const type_data *t) noexcept {
...
+        weaklistoffset = (Py_ssize_t) basicsize;
+        basicsize += ptr_size;
+
+        members[num_members] = PyMemberDef{ "__weaklistoffset__", T_PYSSIZET,
+                                            weaklistoffset, READONLY,
+                                            nullptr };
+        ++num_members;
 
         // Install GC traverse and clear routines if not inherited/overridden
+        // Technically having weak references does not imply we need GC. However
```
hawkinsp, line 471: I'm not entirely happy with this, but I'm trying to avoid the direct call to `PyObject_Malloc()` in inst_new_impl which calculates its own size and wouldn't get the right size. This seemed like it was the simplest fix.

src/nb_type.cpp
```diff
@@ -550,7 +594,7 @@ PyObject *nb_type_new(const type_data *t) noexcept {
     #undef COPY_FIELD
 
     if (temp_tp->tp_members) {
-        tp->tp_members = (PyMemberDef*)((char *)tp + tp->tp_basicsize);
+        tp->tp_members = (PyMemberDef*)((char *)tp + Py_TYPE(tp)->tp_basicsize);
```
hawkinsp, line 597: I realize my last PR had a tiny bug: we should be using the basicsize of the type itself.

hawkinsp (commented):
Added the requested test.

src/nb_type.cpp
wjakob: Fixed here: https://github.com/wjakob/nanobind/commit/6a25ef398b2fd91d82c2241ff42b482843d2b130

src/nb_type.cpp
wjakob: It's fine with me. The cost of enabling GC for a type is not that high. The requirement of a weakly-reference type is sufficiently narrow that this seems like the easiest compromise.

wjakob:
Could you add a variation of that test that works as before, but which subclasses your test classes from within Python?  Extending C++ types from Python involves some extra complexity (`nb_type_init`) that would be good to test here as well. It should be as simply as writing

```
class MyStructWithWeakrefs(t.StructWithWeakrefs):
    pass
class MyStructWithWeakrefsAndDynamicAttrs(t.StructWithWeakrefsAndDynamicAttrs):
    pass
```
and then adjusting the rest of your existing test with those types. 

Also: in the test, instead of assigning the `int` value `101` to the dynamic attributes dict, could you instead use `t.Struct(101)` and read the attribute using `assert p.a_dynamic_attr.value() = 101`? That way, nanobind's leak checker will be active on top of all the other checks.

wjakob:
It seems that there is something specific about this test that PyPy does not like. It's possible that there might be a bug in the PyPy CPython emulation (would not have been the first). On the other hand types are always weak-referenceable in pybind11, and the test suite passes there with PyPy. Not sure.

wjakob:
Hi @hawkinsp  -- any news on this change?

wjakob:
Hi @hawkinsp -- I have not heard back and will close the PR for now. If you change your mind, please @-mention me here and I will reopen it.

hawkinsp:
Yeah sorry I ran out of time to look at the failing tests. Closing the PR is the right action for now.

PR #103, opened 2022-12-02
Add failing tests with tensor default arguments

mnijhuis-tos:
While trying out replacing pybind11 by nanobind, I noticed that using default values for tensor arguments does not work. I also found no tests for this feature.

This MR adds some (disabled) tests. Please let me know if the tests are correct, and if not, please let me know how to use default tensor argument values.

### DIFF
--- a/tests/test_tensor.cpp
+++ b/tests/test_tensor.cpp
@@ -8,6 +8,9 @@ using namespace nb::literals;
 
 int destruct_count = 0;
 
+const size_t default_shape[1] = {3};
+int default_content[3] = {42, 43, 44};
+
 NB_MODULE(test_tensor_ext, m) {
     m.def("get_shape", [](const nb::tensor<> &t) {
         nb::list l;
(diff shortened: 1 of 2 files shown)


wjakob:
Hi @mnijhuis-tos,

thanks for the report. I added some fixes in `master` so that nanobind no longer segfaults. However, what you are trying to do here -- passing a None value to a `nb::tensor` field is not supported by Nanobind. It only accepts `None` arguments for classes bound using `nb::class_<>`, and when a `.none()` argument modifier is furthermore specified.

My suggestion would be that you rewrite your code to accept `nb::object` instead. You can then use `nb::cast` to turn it into a tensor if it is not None`.

mnijhuis-tos:
Thanks for the hint, however, I'm still having troubles when using default tensor arguments. I am still using this branch with the following global constants:
```
const size_t default_shape[1] = {3};
int default_content[3] = {42, 43, 44};
```

1. With the following binding, calling `default_valid()` in python yields a `<capsule object "dltensor" at ...>`. Calling it a second (or third) time, yields "RuntimeError: nanobind::cast(...): conversion failed!".
```
    m.def("default_valid", [](nb::object& o) {
        return nb::cast<nb::tensor<>>(o);
    }, "array"_a = nb::tensor<int, nb::shape<3>>(default_content, 1, default_shape));
```

2. When I change the `nb::object&` into an `nb::tensor<>&` argument, and remove the cast, the first invocation again yields a `<capsule object "dltensor" at ...>` but further calls yield:
```
TypeError: default_valid(): incompatible function arguments. The following argument types are supported:
    1. default_valid(array: tensor[] = <capsule object "used_dltensor" at 0x7f9a1837a4b0>) -> tensor[]
```

3. When I return a numpy tensor using `"array"_a = nb::tensor<nb::numpy, int, nb::shape<3>>(default_content, 1, default_shape)`, I get a `RuntimeError: nanobind::cast(...): conversion failed!` on the first invocation of `default_valid()`. I found that `tensor_import` does not recognise the numpy array: The object is not a capsule, the module name ("numpy") does not match "tensorflow", "torch" or "jaxlib", and `dlpack_from_buffer_protocol` returns an invalid capsule.

Some conclusions:
- Specifying a default tensor argument does seem to work: Using `nb::object&` and `nb::cast` is not needed.
- After the first usage, the default argument cannot be used anymore. The name "used_dltensor" confirms this conclusion.
- Using numpy tensors as default argument value does not work.

mnijhuis-tos:
When I take a different approach, and use an `nb::object&` together with `"array"_a = nb::none()`, I can create a default tensor argument. I will update my branch, as it is a nice example of how to do this.

PR #106, opened 2022-12-05
Support hashing enumeration values

mnijhuis-tos:
Add nb_enum_hash using Py_tp_hash, as suggested in https://github.com/wjakob/nanobind/discussions/105 .


wjakob:
Looks great, thank you!

PR #114, opened 2023-01-06
Add `nb::bind_map`

nicholasjng:
Adds a port of pybind11's `bind_map` to nanobind, including tests.

For now, all additions have been made inline in the new header file, nanobind/stl/bind_map.h. This file opts the user into the standard library headers `<string>` (due to its use of string concatenation for constructing dynamic keys/values view names) and `<memory>` (for dynamically allocating keys/values/items views.)

--------------------------------------------------------------------------

This is my best effort of porting over `py::bind_map`, so far. Since I am not writing C++ that often, please be relentless so that I can learn and do better :)

The use of `<string>` and `<memory>` are only necessary because I ported over https://github.com/pybind/pybind11/pull/4353, which pulls them in for dynamically constructing string reprentations for keys/value/items views. Please give guidance on whether this is ok, or whether that makes `bind_map` incompatible with nanobind's design goals.

I would like guidance on the following questions that came up over the course of developing (I also left some TODOs which are essentially questions that I asked myself while implementing):

- How should I structure the added `using` defs (I assume they are detail-namespace-wide)?
- Is the polymorphism in `KeysViewImpl` et al. compatible with nanobind? (this ties into the above)
- I was confused at having to use `NB_MAKE_OPAQUE` in the tests (they are essentially ported verbatim from `pybind`), to avoid a static assertion about a missing type caster, where pybind does apparently not. What am I doing wrong there?

Thanks in advance for guidance and comments.

### DIFF
--- a/include/nanobind/stl/bind_map.h
+++ b/include/nanobind/stl/bind_map.h
@@ -1,0 +1,301 @@
+/*
+    nanobind/stl/bind_map.h: nb::bind_map()
+
+    This implementation is a port from pybind11 with minimal adjustments.
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <nanobind/nanobind.h>
+#include <nanobind/make_iterator.h>
+#include <nanobind/stl/unique_ptr.h> // TODO: Opt out of unique_ptr (used for returning key/value views)?
+#include <nanobind/stl/string.h>     // TODO: Opt out of string (used for constructing "(Keys|Values|Items)View[type]"-style names to bind to the module scope)
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+// only available since C++20
+using std::remove_cv_t;
+using std::remove_reference_t;
+
+template <class T>
+struct remove_cvref
+{
+    using type = remove_cv_t<remove_reference_t<T>>;
+};
+template <class T>
+using remove_cvref_t = typename remove_cvref<T>::type;
+
+template <typename, typename, typename... Args>
+void map_if_insertion_operator(const Args &...) {}
+template <typename, typename, typename... Args>
+void map_assignment(const Args &...) {}
+
+// Map assignment when copy-assignable: just copy the value
+template <typename Map, typename Class_>
+void map_assignment(
+    std::enable_if_t<std::is_copy_assignable<typename Map::mapped_type>::value, Class_> &cl)
+{
+    using KeyType = typename Map::key_type;
+    using MappedType = typename Map::mapped_type;
+
+    cl.def("__setitem__", [](Map &m, const KeyType &k, const MappedType &v)
+           {
+    auto it = m.find(k);
+    if (it != m.end()) {
+      it->second = v;
+    } else {
+      m.emplace(k, v);
+    } });
+}
+
+// Not copy-assignable, but still copy-constructible: we can update the value by
+// erasing and reinserting
+template <typename Map, typename Class_>
+void map_assignment(
+    std::enable_if_t<!std::is_copy_assignable<typename Map::mapped_type>::value &&
+                         std::is_copy_constructible<typename Map::mapped_type>::value,
+                     Class_> &cl)
+{
+    using KeyType = typename Map::key_type;
+    using MappedType = typename Map::mapped_type;
+
+    cl.def("__setitem__", [](Map &m, const KeyType &k, const MappedType &v)
+           {
+    // We can't use m[k] = v; because value type might not be default
+    // constructable
+    auto r = m.emplace(k, v);
+    if (!r.second) {
+      // value type is not copy assignable so the only way to insert it is to
+      // erase it first...
+      m.erase(r.first);
+      m.emplace(k, v);
+    } });
+}
+
+// template KeysView, ValuesView and ItemsView on their key type, so as
+// to bind different keys views for different key types
+template <typename KeyType>
+struct keys_view
+{
+    virtual size_t len() = 0;
+    virtual iterator iter() = 0;
+    virtual bool contains(const KeyType &k) = 0;
+    virtual bool contains(const object &k) = 0;
+    virtual ~keys_view() = default;
+};
+
+template <typename MappedType>
+struct values_view
+{
+    virtual size_t len() = 0;
+    virtual iterator iter() = 0;
+    virtual ~values_view() = default;
+};
+
+template <typename KeyType, typename MappedType>
+struct items_view
+{
+    virtual size_t len() = 0;
+    virtual iterator iter() = 0;
+    virtual ~items_view() = default;
+};
+
+template <typename Map, typename KeysView>
+struct KeysViewImpl : public KeysView
+{
+    explicit KeysViewImpl(Map &map) : map(map) {}
+    size_t len() override { return map.size(); }
+    // TODO: Does type<Map> at any point return nullptr (since Map is not bound yet)?
+    iterator iter() override { return make_key_iterator(type<Map>(), "make_key_iterator", map.begin(), map.end()); }
+    bool contains(const typename Map::key_type &k) override
+    {
+        return map.find(k) != map.end();
+    }
+    bool contains(const object &) override { return false; }
+    Map &map;
+};
+
+template <typename Map, typename ValuesView>
+struct ValuesViewImpl : public ValuesView
+{
+    explicit ValuesViewImpl(Map &map) : map(map) {}
+    size_t len() override { return map.size(); }
+    // TODO: Does type<Map> at any point return nullptr in this implementation (since Map is not bound yet)?
+    iterator iter() override
+    {
+        return make_value_iterator(type<Map>(), "make_value_iterator", map.begin(), map.end());
+    }
+    Map &map;
+};
+
+template <typename Map, typename ItemsView>
+struct ItemsViewImpl : public ItemsView
+{
+    explicit ItemsViewImpl(Map &map) : map(map) {}
+    size_t len() override { return map.size(); }
+    // TODO: Does type<Map> at any point return nullptr (since Map is not bound yet)?
+    iterator iter() override { return make_iterator(type<Map>(), "make_item_iterator", map.begin(), map.end()); }
+    Map &map;
+};
+
+NAMESPACE_END(detail)
+
+template <typename Map, typename... Args>
+class_<Map> bind_map(handle scope, const char *name, Args &&...args)
+{
+    using KeyType = typename Map::key_type;
+    using MappedType = typename Map::mapped_type;
+    // we do not want a different key type / mapped type between
+    // e.g. double and const double, or double and double& -> strip cv qualifier
+    using StrippedKeyType = detail::remove_cvref_t<KeyType>;
+    using StrippedMappedType = detail::remove_cvref_t<MappedType>;
+    using KeysView = detail::keys_view<StrippedKeyType>;
+    using ValuesView = detail::values_view<StrippedMappedType>;
+    using ItemsView = detail::items_view<StrippedKeyType, StrippedMappedType>;
+    using Class_ = class_<Map>;
+
+    Class_ cl(scope, name, std::forward<Args>(args)...);
+    static constexpr auto key_type_descr = detail::make_caster<KeyType>::Name;
+    static constexpr auto mapped_type_descr = detail::make_caster<MappedType>::Name;
+    std::string key_type_str(key_type_descr.text), mapped_type_str(mapped_type_descr.text);
+
+    // If key type isn't properly wrapped, fall back to C++ names
+    if (key_type_str == "%")
+    {
+        key_type_str = std::string(typeid(KeyType).name());
+    }
+    // Similarly for value type:
+    if (mapped_type_str == "%")
+    {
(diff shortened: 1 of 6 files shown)


wjakob:
Hi Nicholas,

this looks like a great start. A few comments:

nanobind uses C++17, which brought one huge simplification: `if constexpr`. All the ugly declaring of various functions with SFINAE and `std::enable_if` can go. You can just have a single function with a corresponding `if constexpr (condition) { ...}` block containing logic that must be disabled in some settings. This feature contributed significantly to the simplicity of nanobind compared to pybind11, and it would be good to also use it here.

hat way, you can e.g. merge the various ``__setitem__`` bindings and have an `if constexpr` adaptation *inside*. It will also significantly cut down on redundant type alias declarations like `using KeyType = ..`.

Minor: Again, due to C++17, type trait-derived values like `std::is_copy_constructible<typename Map::mapped_type>::value` can be turned into the shorter `std::is_copy_constructible_v<typename Map::mapped_type>`. T

The use of polymorphism seems overkill to me (it might have been more relevant for pybind11, where creating bindings for things in general has a larger cost). Let's instead use map-specific variants (e.g. replacing `ItemsView` by `ItemsViewImpl`, etc) without virtual destructor, etc. This means that the calls to `nb_type_lookup` can disappear.

You can probably even declare the helper view types *inline* within the main binding function.

The `unique_ptr` isn't needed -- you can directly `return new ItemsView(..)` etc., and nanobind will properly take ownership of the pointer.

Likewise, the `string` bindings also doesn't seem to be needed. You can install the views within the scope of their parent container (since there is now a 1:1 mapping), and then their relation ship will be clarified.

What was the static assertion you got?

nicholasjng:
Thanks for the suggestions! I read up on `if constexpr` and used it to refactor the `__setitem` operator. I integrated your suggestions in  [38c254d](https://github.com/wjakob/nanobind/pull/114/commits/38c254dbd63b09c393bf48c5ad2529067ac3bdd8), let me know if I understood everything correctly.

If so, I can also redesign the tests later. Regarding the compile-time assertion, this is the CMake output I'm getting without `NB_MAKE_OPAQUE` (which I just pushed):

[output.txt](https://github.com/wjakob/nanobind/files/10361704/output.txt)

tests/test_stl_bind_map.cpp
```diff
@@ -0,0 +1,17 @@
+#include <nanobind/nanobind.h>
+#include <nanobind/stl/bind_map.h>
+#include <nanobind/stl/string.h>
+#include <nanobind/stl/map.h>
```
wjakob, line 4: Can you remove `stl/map.h` and `stl/unordered_map.h` from this file? This is likely the source of your error message regarding `NB_OPAQUE` (you imported the default STL map type caster *and* tried to declare bindings for a specific STL map).

include/nanobind/stl/bind_map.h
```diff
@@ -0,0 +1,196 @@
...
+NAMESPACE_BEGIN(detail)
+
+template <typename Map, typename Class_>
+void map_assignment(Class_ &cl)
+{
+    using KeyType = typename Map::key_type;
+    using MappedType = typename Map::mapped_type;
+    // Map assignment when copy-assignable: just copy the value
+    if constexpr(std::is_copy_assignable_v<MappedType>){
+        cl.def("__setitem__", [](Map &m, const KeyType &k, const MappedType &v)
```
wjakob, line 25: You can merge this `cl.def` with the one below and put the `if constexpr` *inside* the lanbda function.  The entire body of `map_assignment` can be merged into `bind_map`, there is no longer a reason to let this be a separate function given the `if constexpr` feature.

wjakob: More specifically: run `cl.def` *if* `std::is_copy_assignable_v<MappedType>` or `std::is_copy_constructible_v<MappedType>` holds. Then, choose the right way to do the assignment inside the lambda function with another `if constexpr`.

include/nanobind/stl/bind_map.h
```diff
@@ -0,0 +1,196 @@
...
+    using Class_ = class_<Map>;
+
+    Class_ cl(scope, name, std::forward<Args>(args)...);
+
+    // install KeysView directly into the map class scope
+    class_<KeysView> keys_view(cl, "KeysView");
+    keys_view.def("__len__", &KeysView::len);
+    keys_view.def("__iter__", 
+                  &KeysView::iter,
+                  keep_alive<0, 1>() /* Essential: keep view alive while iterator exists */
```
wjakob, line 103: I think you can remove those `Essential: keep view alive while iterator exists` comments, which repeat many times.

include/nanobind/stl/bind_map.h
```diff
@@ -0,0 +1,196 @@
...
+
+    cl.def("__contains__", [](Map &m, const KeyType &k) -> bool {
+        auto it = m.find(k);
+        if (it == m.end()) {
+            return false;
+        }
+        return true; 
+    });
+    // Fallback for when the object is not of the key type
+    cl.def("__contains__", [](Map &, const object &) -> bool { return false; });
```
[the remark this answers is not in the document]
wjakob: replace `const object &` by `handle`

wjakob (changes requested):
A few minor comments. I think the most important missing part are testcases -- we will want to cover all the functions and also account for special cases (for example non-copy-assignable types that are mentioned in one place.)

My suggestion would be that you simply reuse the existing tests by porting them from pybind11.

include/nanobind/stl/bind_map.h
```diff
@@ -0,0 +1,196 @@
+/*
+    nanobind/stl/bind_map.h: nb::bind_map()
+
+    This implementation is a port from pybind11 with minimal adjustments.
```
wjakob, line 4: I think this line can go away, the implementation is quite different at this point.

include/nanobind/stl/bind_map.h
```diff
@@ -0,0 +1,196 @@
+/*
+    nanobind/stl/bind_map.h: nb::bind_map()
```
wjakob, line 2: `nb::bind_map()` -> `Automatic creation of bindings for map-style containers`

tests/test_stl_bind_map.cpp
nicholasjng: No, removing them makes the compiler complain (`no template 'map' in namespace 'std'`).

include/nanobind/stl/bind_map.h
nicholasjng: Done in 312a763.

include/nanobind/stl/bind_map.h
nicholasjng: Removed in 312a763.

include/nanobind/stl/bind_map.h
```diff
@@ -0,0 +1,196 @@
...
+            auto it = m.find(k);
+            if (it == m.end()) {
+                throw key_error();
+            }
+            return it->second;
+        },
+        rv_policy::reference_internal // ref + keepalive
+    );
+
+    cl.def("__contains__", [](Map &m, const KeyType &k) -> bool {
```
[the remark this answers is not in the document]
nicholasjng: Changed to `const Map &m` in 312a763.

include/nanobind/stl/bind_map.h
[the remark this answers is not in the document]
nicholasjng: Changed `const object &` to `handle` in 312a763, made Map const.

include/nanobind/stl/bind_map.h
nicholasjng: Added in 312a763.

tests/test_stl_bind_map.cpp
wjakob: You can include `map` from the STL:
```
#include <map>
#include <unordered_map>
```
The nanobind header pulls in type casters that are unwanted in this context.

nicholasjng: Thanks, that works. Should I do the same with `<string>` instead of `<nanobind/stl/string.h>`?

wjakob: right. You only need `nanobind/stl/..` to have automated conversion of the associated type between Python and C++ (and those automated converters work by always performing a copy).

This means that you would only need `nanobind/stl/string.h` if there is a testcase which returns a string to Python (or takes one as an input).

nicholasjng:
Thanks for the review, I just ported over the relevant tests from pybind. I am still getting compile errors, since nanobind attempts to bind a copy constructor to the map here for the non-copyable value class `E_nc`:

https://github.com/wjakob/nanobind/blob/f4d5e120fb505927ce0cfd9b03dadec0fc344ca5/include/nanobind/nb_class.h#L277-L279

I think this might be an error in my `if constexpr` branch, and I need to check for `std::is_trivially_copy_constructible_v`, too.

tests/test_stl_bind_map.cpp
nicholasjng: Included `<string>` now instead of `nanobind/stl/string.h`, but the tests are still failing (the maps with non-copyable values).  I don't know what is going wrong there, the branch seems correct - I will look at it later with a fresh mind.

nicholasjng:
I went into pybind's codebase to check why it works there, and they are using SFINAE to overload their own version of `std::is_copy_constructible_v` which returns `false` for containers with non-copyable values (same with `is_copy_assignable`:

https://github.com/pybind/pybind11/blob/a34596bfe1947b4a6b0bcc4218e1f72d0c2e9b4c/include/pybind11/detail/type_caster_base.h#L825-L841

Is there an obvious way to implement this as a `constexpr bool` to use C++17 style syntax?

wjakob:
How about
```
if contexpr(std::is_copy_constructible_v<typename Container::value_type>) { .. }
```
?

nicholasjng:
> How about
> 
> ```
> if contexpr(std::is_copy_constructible_v<typename Container::value_type>) { .. }
> ```
> 
> ?

That works for non-copyable value types, but fails for nested maps with non-copyable values (since `std::is_copy_constructible_v<std::map<key, noncopyable_value>> = true`), for example in L69:

```
test_stl_bind_map.cpp: L69
nb::bind_map<std::map<int, std::map<int, E_nc>>>(m, "MapMapENC");
```

wjakob:
Ah, I understand now. The logic in pybind11 actually looks good to me, and implementing this using partial overloads is probably still the best way to go. You can basically copy those definitions of `is_copy_constructible` and then query them with `if constexpr` from your code (replacing `std::is_copy_constructible` with `detail::is_copy_constructible`).

wjakob:
Can you add these type traits to a new file `include/nanobind/stl/detail/type_traits.h`? This is all very specific to STL bindings and should not be added to nanobind by default.

nicholasjng:
Will do, thanks for the clarification!

wjakob:
Ah wait -- if I understand correctly, the `if constexpr (!std::is_trivially_copy_constructible_v<T>)` part in nanobind is broken, correct?

nicholasjng:
Yes, that's due to the same phenomenon - a map with a non-copyable value is copyable, and if I understand correctly, the default copy constructor of a map tries to call the copy constructor of its value (or rather, the `std::pair<key, value>`, whose copy constructor is deleted if one of the two types has a deleted CC), which is an error here.

wjakob:
I pushed a fix, see https://github.com/wjakob/nanobind/commit/94bdb05d84bb15d1a57d23097529f53a0c82c453

nicholasjng:
Thanks, looks good to me. Should I add aliases for the values (`detail::is_copy_constructible_v`), and do I need to do the same for `std::is_copy_assignable`? (It was there in https://github.com/pybind/pybind11/pull/1886, so I assume that needs to go in too)

wjakob:
> Thanks, looks good to me. Should I add aliases for the values (detail::is_copy_constructible_v), and do I need to do the same for std::is_copy_assignable? (It was there in https://github.com/pybind/pybind11/pull/1886, so I assume that needs to go in too)

Sure, let's do that. The goal of my commit was just to fix the parts affecting core parts of nanobind without adding the complicated templates there by default.

nicholasjng:
Thanks, just pushed the ported templates and values. I'm confident that that fixed the `__setitem__` assignment - tests are still failing though, because the created `ValuesView` iterator impl ends up calling the wrong overload, which becomes an`insert`, creating a copy of the map value. Is there a missing reference somewhere? Seems like we should not be copying on iteration here.

wjakob:
Hi @nicholasjng -- I made a simplification pass over your code, please check that it still okay to you. I had to make a change to nanobind on master for this to work (https://github.com/wjakob/nanobind/commit/8916f51ad1a25318b5c9fcb07c153f6b72a43bd2) and I then rebased your code & force-pushed to your branch. 😇

wjakob (approved):

nicholasjng:
Nice, thank you for the fixes! Could you briefly clarify why it needs to be `const` as well in 8916f51?

wjakob:
The `const` may not be needed, I added it to accommodate every possible iterator. nanobind will in any case cast away the `const` part internally since such a concept isn't easily mapped onto Python.

wjakob:
Shall I merge this then?

wjakob:
Thanks!

nicholasjng:
Thanks for the help and the review! See you again soon for `nb::bind_vector`, I guess :)

wjakob:
That one should be refreshingly easy in comparison.

PR #115, opened 2023-01-20
Switch from C++20 designated initializers to aggregate initializers to support older compilers

matthewdcong:
nanobind fails to compile on Visual Studio 2017 and GCC 7.3.1 (the compiler version in the Red Hat Developer Toolset for RHEL 7) because it uses C++20 designated initializers for the `PyType_Spec` structs in nb_type.cpp and nb_internals.cpp. Replacing these with aggregate initializers allows nanobind to compile on these older compilers.

Refer to discussion in https://github.com/wjakob/nanobind/discussions/108

### DIFF
--- a/src/nb_internals.cpp
+++ b/src/nb_internals.cpp
@@ -132,12 +132,12 @@ static PyType_Slot nb_func_slots[] = {
 };
 
 static PyType_Spec nb_func_spec = {
-    .name = "nanobind.nb_func",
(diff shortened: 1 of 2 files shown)


wjakob:
Thanks!

wjakob:
merged with minor syntax amendments, and an added fix for the cmake build system (no longer necessary to enable c++20 mode on MSVC)

PR #119, opened 2023-01-31
Add a convenience function for calling the builtin `hash`

tjstum:
pybind11 allows for calling `pybind11::hash(foo)` in C++ as an equivalent of `hash(foo)` in Python. This is handy when binding a type whose Python hash should match that of some other Python object (like a string, perhaps). Expose something similar for nanobind

Test included

### DIFF
--- a/include/nanobind/nb_lib.h
+++ b/include/nanobind/nb_lib.h
@@ -161,6 +161,10 @@ NB_CORE size_t obj_len_hint(PyObject *o) noexcept;
 /// Obtain a string representation of a Python object
 NB_CORE PyObject* obj_repr(PyObject *o);
 
+/// Obtain a hash value of a Python object
+NB_CORE Py_hash_t obj_hash(PyObject *o);
+
+
 /// Perform a comparison between Python objects and handle errors
 NB_CORE bool obj_comp(PyObject *a, PyObject *b, int value);
 
--- a/include/nanobind/nb_lib.h
+++ b/include/nanobind/nb_lib.h
@@ -164,7 +164,6 @@ NB_CORE PyObject* obj_repr(PyObject *o);
 /// Obtain a hash value of a Python object
 NB_CORE Py_hash_t obj_hash(PyObject *o);
 
-
 /// Perform a comparison between Python objects and handle errors
 NB_CORE bool obj_comp(PyObject *a, PyObject *b, int value);
 
--- a/tests/test_classes.cpp
+++ b/tests/test_classes.cpp
@@ -439,4 +439,13 @@ NB_MODULE(test_classes_ext, m) {
     m.def("is_int_1", [](nb::handle h) { return nb::isinstance<int>(h); });
     m.def("is_int_2", [](nb::handle h) { return nb::isinstance<nb::int_>(h); });
     m.def("is_struct", [](nb::handle h) { return nb::isinstance<Struct>(h); });
+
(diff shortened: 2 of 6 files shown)


include/nanobind/nb_lib.h
```diff
@@ -154,6 +154,10 @@ NB_CORE size_t obj_len_hint(PyObject *o) noexcept;
 /// Obtain a string representation of a Python object
 NB_CORE PyObject* obj_repr(PyObject *o);
 
+/// Obtain a hash value of a Python object
+NB_CORE Py_hash_t obj_hash(PyObject *o);
+
```
wjakob, line 159: remove newline

tests/test_classes.cpp
```diff
@@ -435,4 +435,13 @@ NB_MODULE(test_classes_ext, m) {
 
     using NonCopyableVec = std::vector<NonCopyable>;
     nb::class_<NonCopyableVec>(m, "NonCopyableVec");
+
+    // test32_deference
+    struct Deferential {nb::object target; };
```
wjakob, line 440: space after `{`.

Can you rename the class from `Dereferential` (which I think is more common to express "honor" to another person) to something like `Indirect` or `HashWrapper`? Same for the test case name.

tests/test_classes.cpp
```diff
@@ -435,4 +435,13 @@ NB_MODULE(test_classes_ext, m) {
...
+    nb::class_<Deferential>(m, "Deferential")
+        .def(nb::init<nb::object>())
+        .def("__eq__", [](const Deferential &self, nb::object other){
```
wjakob, line 443: space beteween `)` and `{`.

wjakob (commented):
This looks good. Just a few minor change requests.

tests/test_classes.cpp
tjstum: Sure! I was thinking it was more "defer to someone else." I'll change to `HashWrapper`

tjstum:
Hi @wjakob!

I saw that the test case here fails on PyPy. I am just curious if you know of some reason why it would fail like this (but pass on all CPythons and ... the Windows PyPy?). I can do more investigation if you don't have any leads.

Thanks.

wjakob:
This looks to me like a PyPy issue. The efficient workflow to get it fixed (like with CPython) would be to create a standalone C-API reproducer and submit it here: https://foss.heptapod.net/pypy/pypy/-/issues

wjakob:
It's interesting that the hex encoding of these values is quite similar:
```
>>> hex(114184800867133550)
'0x195aa78ae6f286e'
>>> hex(4725870819294521452)
'0x4195aa78ae6f286c'
```

Makes me wonder if it could be a integer/sign conversion related issue.

wjakob:
Note: in addition to the issue with PyPy, there is a potential overlap with a `hash` function defined in `nanobind/operators.h`. It would be good to test your changes in the context of a C++ file that includes this header. (Maybe it's fine since that other version targets the `nb::self` placeholder value).

tjstum:
Thanks. I'll be sure to test with that header included as well.

I'm traveling for a holiday/work right now, so it might be a few more days/week before I make progress with this effort.

wjakob:
I will close this PR for now. It will reopen it if there are any news (e.g. tracking down those differences in PyPy).

PR #120, opened 2023-02-01
Initial support for the Eigen library

wjakob:
This commit adds custom casters for the Eigen linear algebra library. It uses the existing `nb_tensor` infrastructure to implement a conversion between Eigen types and NumPy arrays that currently supports:

- `Eigen::Matrix/Array<..>` and derived types (row/column vectors)
- `Eigen::Map<..>`
- `Eigen::Ref<..>`
- Expression templates (`CwiseBinaryOp`, etc.)

There are a few limitations:

- Eigen tensors are not supported yet.
- Sparse matrix types are not supported yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,10 +21,10 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'windows-2022', 'macos-latest']
-        python: ['3.8', '3.9', '3.10', '3.11', '3.12.0-alpha.4', 'pypy-3.9-nightly']
+        python: ['3.8', '3.9', '3.10', '3.11', '3.12.0-alpha.4', 'pypy3.9']
         exclude:
           - os: 'macos-latest'
-            python: 'pypy-3.9-nightly'
+            python: 'pypy3.9'
           - os: 'windows-2022'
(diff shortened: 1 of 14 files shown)


wjakob:
Could be interesting for you: @ianhbell, @brettc

ianhbell:
I am very happy about this addition; but doesn't this cut against the explicit design goals of nanobind to be small and targeted?

ianhbell:
Agreed; opt-in is always the way to go. I'm looking forward to playing with the Eigen support in nanobind.

brettc:
Thanks @wjakob,  this does look interesting. I have settled on using [xtensor](https://github.com/xtensor-stack/xtensor/) for my current projects, but this gives some hints on how to integrate it.

My apologies for letting the PR stagnate. I'm still keen to push it through, but it needs more thought and my TODO keeps growing.

wjakob:
> My apologies for letting the PR stagnate. I'm still keen to push it through, but it needs more thought and my TODO keeps growing.

Can you check with the latest version to see if it is needed? I moved the NumPy tensor exchange back onto the buffer protocol.

brettc:
FYI: https://github.com/xtensor-stack/xtensor-python/issues/284

PR #121, opened 2023-02-02
Discussion - All nanobind types should have a tp_traverse?

tjstum:
Hello again! I wasn't _totally_ sure what the best area of the repo was the best to bring this up in. Please forgive me if I picked incorrectly

While cleaning up some leak warnings in our extension module, I was looking into how `nanobind::enum_` handles the cycles that are created between the type -> enumeration members that are instances, stored as attributes on the type ->  the type. Our extension module has a few cases of this pattern: we have our own version of enums (with all the instances being stored as attributes of the type). We also sometimes expose "special" instances of a non-enumeration class as an attribute on the type object itself. That could be done in Python (as my example code shows here) or in C++.

I think the original issue is in [bpo-40217](https://bugs.python.org/issue40217). In 3.8, it looks like CPython tried to handle this automatically (by adding a `tp_traverse` slot; https://github.com/python/cpython/pull/19414). However, it was reverted (I think? https://github.com/python/cpython/pull/20264/), and it seems that it's now incumbent on heap types to define a `tp_traverse` that calls `Py_VISIT` on the type object. This is exactly what `nanobind::enum_` [does](https://github.com/wjakob/nanobind/blob/50e761af3f7cdc2a3986f01bcbd75bf5de2123b6/src/nb_enum.cpp#L242), and `nanobind::class_` [will do it](https://github.com/wjakob/nanobind/blob/50e761af3f7cdc2a3986f01bcbd75bf5de2123b6/src/nb_type.cpp#L432) if the objects have an instance dict. It seems, though, that nanobind should do this for _all_ created types. Otherwise they leak at shutdown.

There's almost certainly some nuance from this issue that I'm missing, so I figured I'd bring this up for discussion (showing the issue) before proposing any specific fix. Thanks for your consideration.

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -191,15 +191,6 @@ NB_ENUM_UNOP(neg, PyNumber_Negative)
 NB_ENUM_UNOP(inv, PyNumber_Invert)
 NB_ENUM_UNOP(abs, PyNumber_Absolute)
 
-int nb_enum_clear(PyObject *) {
-    return 0;
-}
-
-int nb_enum_traverse(PyObject *o, visitproc visit, void *arg) {
-    Py_VISIT(Py_TYPE(o));
-    return 0;
-}
-
 Py_hash_t nb_enum_hash(PyObject *o) {
     Py_hash_t value = 0;
     type_data *t = nb_type_data(Py_TYPE(o));
@@ -239,8 +230,6 @@ void nb_enum_prepare(PyType_Slot **s, bool is_arithmetic) {
     *t++ = { Py_nb_int, (void *) nb_enum_int };
     *t++ = { Py_nb_index, (void *) nb_enum_int };
     *t++ = { Py_tp_getset, (void *) nb_enum_getset };
-    *t++ = { Py_tp_traverse, (void *) nb_enum_traverse };
-    *t++ = { Py_tp_clear, (void *) nb_enum_clear };
     *t++ = { Py_tp_hash, (void *) nb_enum_hash };
 
     if (is_arithmetic) {
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -25,16 +25,27 @@ static PyObject **nb_dict_ptr(PyObject *self) {
 #endif
 }
 
-static int inst_clear(PyObject *self) {
+static int inst_dict_clear(PyObject *self) {
     PyObject *&dict = *nb_dict_ptr(self);
     Py_CLEAR(dict);
     return 0;
 }
 
-static int inst_traverse(PyObject *self, visitproc visit, void *arg) {
+static int inst_dict_traverse(PyObject *self, visitproc visit, void *arg) {
+#if PY_VERSION_HEX >= 0x03090000
+    Py_VISIT(Py_TYPE(self));
+#endif
     PyObject *&dict = *nb_dict_ptr(self);
     if (dict)
         Py_VISIT(dict);
+    return 0;
+}
+
+static int inst_nodict_clear(PyObject *self) {
+    return 0;
+}
+
+static int inst_nodict_traverse(PyObject *self, visitproc visit, void *arg) {
 #if PY_VERSION_HEX >= 0x03090000
     Py_VISIT(Py_TYPE(self));
 #endif
@@ -386,7 +397,7 @@ PyObject *nb_type_new(const type_data *t) noexcept {
         /* .name = */ name_copy,
         /* .basicsize = */ (int) basicsize,
         /* .itemsize = */ 0,
-        /* .flags = */ Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+        /* .flags = */ Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE | Py_TPFLAGS_HAVE_GC,
         /* .slots = */ slots
     };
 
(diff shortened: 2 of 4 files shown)


wjakob:
Hi @tjstum,

You are right that Python types that could be part of reference cycles should have a `tp_traverse` slot. It is possible to accomplish this in nanobind using the `nb::type_slots` annotation (see `tests/test_classes.cpp lines 88 and 426).

However, I don't see how nanobind could completely automate this aspect (besides special cases like instance dictionaries that you already noticed). How should it know what fields/data structure members to traverse in complex user-defined types?

Best,
Wenzel

tjstum:
Hi Wenzel,

Thanks for writing back. I agree that there's probably not much that nanobind can do in the general case to completely automate this. I was thinking that it might be a good default, though, for nanobind to create type objects that use the [`inst_traverse`](https://github.com/wjakob/nanobind/blob/50e761af3f7cdc2a3986f01bcbd75bf5de2123b6/src/nb_type.cpp#L34) for the `tp_traverse`, as long as the user hasn't specified any other traverse slot (basically, move [this conditional](https://github.com/wjakob/nanobind/blob/50e761af3f7cdc2a3986f01bcbd75bf5de2123b6/src/nb_type.cpp#L433) out from the `has_dynamic_attr` if block).

Like I said initially, there might be something that I'm missing, so I'm interested in your thoughts. This seems to me to be animprovement for the majority of nanobind classes, though!

tjstum:
I should also mention that I'm happy to update this pull request to show my proposed change, in case it's not clear what I am suggesting

wjakob:
Aha, got it! Yes, I think that makes sense. I think that we would need two separate versions of `inst_traverse` (one for classes with instance dictionaries, and one for the base case where only `Py_VISIT(Py_TYPE(self))` needs to be visited). Please go ahead and create such a PR.

tjstum:
Thanks! I just updated my branch with the proposed code change. I am happy to create a new PR or edit commits or whatever you'd like.

wjakob:
Hi @tjstum,

I thought a bit more about this, and I am not sure that this PR is the right way to go for nanobind.

First of all, it seems that this is really a Python issue and not a nanobind issue. Unless I misunderstood, you should be able to create the same reference cycle with any other non-GCable Python object. It's just that nanobind is really noisy about leaks and will complain in this case, while Python leaks silently. So if anything, I feel like the onus is on Python to implement some special handling of cycles involving instances and their class object.

Second, Integrating a type with the GC comes with certain costs: the storage per object goes up by at least 16 bytes to store indirections in a bidirectional linked list. Which is fine a big and heavy class, but it's costly for simple types that themselves might only consume a few bytes.

I hope the reasoning makes sense -- let me know if I misunderstood something about your issue.

tjstum:
Thanks for the thoughtful consideration, @wjakob.

It has always occurred to me that this situation is quite odd... The instances themselves are not containers and therefore it seems strange to require every nanobind type include `Py_TPFLAGS_HAVE_GC` just because the class itself is a container.
I will try to compose a note to CPython. Would you think the best venue would be the original bpo issue? One of the pull requests? python-dev@? I haven't ever started a discussion in any of those places, so any advice you have would be welcome.

As to this change... I am curious as to your thinking about the use of `nb_traverse` in `nb::enum_` classes. These are all in all the same situation. Does it seem that the GC overhead is justified in that situation?

I think the reason that this doesn't come up in other non-GCable Python types is that they don't support setting arbitrary attributes on the type object:
```python
>>> str.foo = "foo"
TypeError: cannot set 'foo' attribute of immutable type 'str'
>>> int.five = 5
TypeError: cannot set 'five' attribute of immutable type 'int'
```
Perhaps nanobind could do something similar; perhaps nanobind classes should either have the `Py_TPFLAGS_IMMUTABLETYPE` flag (if there is never an `.attr` call to that class from nanobind) **or** have GC methods (and the `Py_TPFLAGS_HAVE_GC` flag). This would allow for simple types not to incur GC overhead while handling this automatically in the majority of cases.
I found this [note](https://docs.python.org/3/howto/isolating-extensions.html#changing-static-types-to-heap-types) in the CPython docs, which does also say that heap types all need the GC methods to resolve these cycles, which is not particularly encouraging.

I can understand if you want to leave this up to the authors of binding code and not have nanobind automatically choose. Please let me know what you think about my suggested compromise above. If you are still uninterested, I will totally understand ... and in that case, I will send a PR to update the docs with the result of this discussion.

wjakob:
I've done this a few times last year, and here is how you make an actionable issue:

1. Don't use nanobind. Create a tiny extension using the C API (example: https://github.com/wjakob/inheritance_issue) and publish it as a repository on GitHub. You can probably reuse my example with just a few examples, the key is to use `PyType_FromSpec` to create a new type from the C API that doesn't have the GC bit set.

2. Open an issue on the GitHub tracker: https://github.com/python/cpython. Explain the issue, add a link to your GitHub repo, and  how to install it easily using Pip
```
pip install git+https://github.com/wjakob/inheritance_issue/
```

You could tag ``encukou``, ``eerlend-asland``, and/or ``markshannon`` who are the experts on those low-level C API bits.

The point about immutable type objects is interesting! That may indeed be a possible direction. I would still go the route of reporting to CPython to see what they say.

wjakob:
By the way, setting an immutable flag on the type would not work: nanobind uses `.attr(name) = ...` to append method and property definitions. Those assignments would fail if the type is created with an immutable flag. The immutable flag could be "patched in" after the all fields/methods are registered, but this would not work for Stable ABI compilation, in which the `PyTypeObject` fields (including `tp_flags`) are hidden.

Basically I think the conclusion (before having heard from the CPython folks is):

- Having a non-GC heap type saves storage, but it is a potential footgun. The user can create a reference cycle by monkey-patching instances of a type into the type object.

- The benefit of fixing this potential footgun seems disproportionate when compared to the cost. Having a type leak is unfortunate, but it is not the end of the world (plus, the warning can be disabled). Enlarging all instances by a nontrivial amount is an unreasonable cost.

- The onus is on the nanobind extension developer. If this use case is expected (as is the case in the `nb::enum_` class), it is the developer's responsibility to add the `tp_traverse` slot.

Coming back to a question you asked earlier: isn't the cost also prohibitive for `nb:enum_<T>`? The overhead is small in this case because the number of enum instances matches the total number of enumeration entries.

wjakob:
I will close this PR for now since it is unlikely that it would be merged, but I am happy to continue the discussion here.

wjakob:
> would you be open to nanobind adding some convenience versions of class_ for users? Perhaps immutable_class, that has Py_TPFLAGS_IMMUTABLETYPE and doesn't support .attr;

This wouldn't be usable in practice. Even a trivial `nb::class_<..>().def(..)` would fail at the `.def()` call because it writes into the type object.

> [..] referential_class, that has Py_TPFLAGS_HAVE_GC (and the version of tp_traverse proposed here). 

This sounds more reasonable. We could have a ``nb::needs_gc`` class binding attribute that can be passed to ``nb::class_``.

tjstum:
> This wouldn't be usable in practice. Even a trivial `nb::class_<..>().def(..)` would fail at the `.def()` call because it writes into the type object.

Right! Of course. My mistake.

> This sounds more reasonable. We could have a `nb::needs_gc` class binding attribute that can be passed to `nb::class_`.

I'll look into that! The attribute sounds like a great compromise.

PR #126, opened 2023-02-14
eigen: support sparse matrix types

hmenke:
This is a first attempt at supporting sparse matrix types from Eigen. The code is mostly ported over from [pybind11](https://github.com/pybind/pybind11/blob/master/include/pybind11/eigen/matrix.h). It compiles but the test segfaults. This could have to with the fact that I have not really a good idea of what I'm doing. Help is very much appreciated.

### DIFF
--- a/include/nanobind/eigen/dense.h
+++ b/include/nanobind/eigen/dense.h
@@ -172,7 +172,7 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_xpr_v<T>>> {
 
 /// Caster for Eigen::Map<T>
 template <typename T, int Options, typename StrideType>
-struct type_caster<Eigen::Map<T, Options, StrideType>> {
+struct type_caster<Eigen::Map<T, Options, StrideType>, enable_if_t<is_eigen_plain_v<T>>> {
     using Map = Eigen::Map<T, Options, StrideType>;
     using NDArray = array_for_eigen_t<Map>;
     using NDArrayCaster = type_caster<NDArray>;
@@ -231,7 +231,7 @@ struct type_caster<Eigen::Map<T, Options, StrideType>> {
 
 /// Caster for Eigen::Ref<T>
 template <typename T, int Options, typename StrideType>
-struct type_caster<Eigen::Ref<T, Options, StrideType>> {
+struct type_caster<Eigen::Ref<T, Options, StrideType>, enable_if_t<is_eigen_plain_v<T>>> {
     using Ref = Eigen::Ref<T, Options, StrideType>;
     using Map = Eigen::Map<T, Options, StrideType>;
     using MapCaster = make_caster<Map>;
--- a/include/nanobind/eigen/dense.h
+++ b/include/nanobind/eigen/dense.h
@@ -13,8 +13,8 @@
 #include <nanobind/ndarray.h>
 #include <Eigen/Core>
 
-static_assert(EIGEN_VERSION_AT_LEAST(3, 2, 7),
-              "Eigen matrix support in pybind11 requires Eigen >= 3.2.7");
+static_assert(EIGEN_VERSION_AT_LEAST(3, 3, 0),
+              "Eigen matrix support in nanobind requires Eigen >= 3.3.0");
 
 NAMESPACE_BEGIN(NB_NAMESPACE)
 
@@ -53,9 +53,13 @@ is_base_of_template_v<T, Eigen::EigenBase>;
 template <typename T> constexpr bool is_eigen_plain_v =
 is_base_of_template_v<T, Eigen::PlainObjectBase>;
 
+/// Detect Eigen::SparseMatrix
+template <typename T> constexpr bool is_eigen_sparse_v =
+is_base_of_template_v<T, Eigen::SparseMatrixBase>;
+
 /// Detects expression templates
 template <typename T> constexpr bool is_eigen_xpr_v =
-    is_eigen_v<T> && !is_eigen_plain_v<T> &&
+    is_eigen_v<T> && !is_eigen_plain_v<T> && !is_eigen_sparse_v<T> &&
     !std::is_base_of_v<Eigen::MapBase<T, Eigen::ReadOnlyAccessors>, T>;
 
 template <typename T> struct type_caster<T, enable_if_t<is_eigen_plain_v<T>>> {
--- a/include/nanobind/eigen/dense.h
+++ b/include/nanobind/eigen/dense.h
@@ -13,8 +13,8 @@
 #include <nanobind/ndarray.h>
 #include <Eigen/Core>
 
-static_assert(EIGEN_VERSION_AT_LEAST(3, 3, 0),
-              "Eigen matrix support in nanobind requires Eigen >= 3.3.0");
+static_assert(EIGEN_VERSION_AT_LEAST(3, 3, 1),
+              "Eigen matrix support in nanobind requires Eigen >= 3.3.1");
 
 NAMESPACE_BEGIN(NB_NAMESPACE)
 
@@ -26,11 +26,14 @@ template <typename T> using DMap = Eigen::Map<T, 0, DStride>;
 NAMESPACE_BEGIN(detail)
 
 template <typename T>
+constexpr int NumDimensions = int(T::MaxSizeAtCompileTime) == 1 ? 0 : bool(T::IsVectorAtCompileTime) ? 1 : 2;
+
+template <typename T>
 using array_for_eigen_t = ndarray<
     typename T::Scalar,
     numpy,
     std::conditional_t<
-        T::NumDimensions == 1,
+        NumDimensions<T> == 1,
         shape<(size_t) T::SizeAtCompileTime>,
         shape<(size_t) T::RowsAtCompileTime,
               (size_t) T::ColsAtCompileTime>>,
@@ -38,7 +41,7 @@ using array_for_eigen_t = ndarray<
         T::InnerStrideAtCompileTime == Eigen::Dynamic,
         any_contig,
         std::conditional_t<
-            T::IsRowMajor || T::NumDimensions == 1,
+            T::IsRowMajor || NumDimensions<T> == 1,
             c_contig,
             f_contig
         >
@@ -75,7 +78,7 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_plain_v<T>>> {
             return false;
         const NDArray &array = caster.value;
 
-        if constexpr (T::NumDimensions == 1) {
+        if constexpr (NumDimensions<T> == 1) {
             value.resize(array.shape(0));
             memcpy(value.data(), array.data(),
                    array.shape(0) * sizeof(Scalar));
@@ -97,10 +100,10 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_plain_v<T>>> {
     }
 
     static handle from_cpp(const T &v, rv_policy policy, cleanup_list *cleanup) noexcept {
-        size_t shape[T::NumDimensions];
-        int64_t strides[T::NumDimensions];
+        size_t shape[NumDimensions<T>];
+        int64_t strides[NumDimensions<T>];
 
-        if constexpr (T::NumDimensions == 1) {
+        if constexpr (NumDimensions<T> == 1) {
             shape[0] = v.size();
             strides[0] = v.innerStride();
         } else {
@@ -143,7 +146,7 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_plain_v<T>>> {
             policy == rv_policy::move ? rv_policy::reference : policy;
 
         object o = steal(NDArrayCaster::from_cpp(
-            NDArray(ptr, T::NumDimensions, shape, owner, strides),
+            NDArray(ptr, NumDimensions<T>, shape, owner, strides),
             array_rv_policy, cleanup));
 
         return o.release();
@@ -184,10 +187,10 @@ struct type_caster<Eigen::Map<T, Options, StrideType>> {
     }
 
     static handle from_cpp(const Map &v, rv_policy, cleanup_list *cleanup) noexcept {
-        size_t shape[T::NumDimensions];
-        int64_t strides[T::NumDimensions];
+        size_t shape[NumDimensions<T>];
+        int64_t strides[NumDimensions<T>];
 
-        if constexpr (T::NumDimensions == 1) {
+        if constexpr (NumDimensions<T> == 1) {
             shape[0] = v.size();
             strides[0] = v.innerStride();
         } else {
@@ -198,7 +201,7 @@ struct type_caster<Eigen::Map<T, Options, StrideType>> {
         }
 
         return NDArrayCaster::from_cpp(
-            NDArray((void *) v.data(), T::NumDimensions, shape, handle(), strides),
+            NDArray((void *) v.data(), NumDimensions<T>, shape, handle(), strides),
             rv_policy::reference, cleanup);
     }
 
--- a/include/nanobind/eigen/sparse.h
+++ b/include/nanobind/eigen/sparse.h
@@ -13,6 +13,7 @@
 #include <nanobind/eigen/dense.h>
 #include <Eigen/SparseCore>
 
+#include <memory>
 #include <type_traits>
 #include <utility>
 
@@ -92,14 +93,13 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
         return from_cpp((const T &) v, policy, cleanup);
     }
 
-    static handle from_cpp(const T &v, rv_policy policy, cleanup_list *cleanup) noexcept {
+    static handle from_cpp(const T &v, rv_policy policy, cleanup_list *) noexcept {
         if (!v.isCompressed()) {
             PyErr_SetString(PyExc_ValueError,
                             "nanobind: unable to return an Eigen sparse matrix that is not in a compressed format. "
                             "Please call `.makeCompressed()` before returning the value on the C++ end.");
             return handle();
         }
-        T & src = const_cast<T &>(v);
 
         object matrix_type;
         try {
@@ -109,16 +109,26 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
             return handle();
         }
 
+        const Index rows = v.rows();
+        const Index cols = v.cols();
         const size_t data_shape[] = { (size_t)v.nonZeros() };
-        const size_t outer_indices_shape[] = { (size_t)((row_major ? v.rows() : v.cols()) + 1) };
-        ScalarNDArray data(src.valuePtr(), 1, data_shape);
-        StorageIndexNDArray outer_indices(src.outerIndexPtr(), 1, outer_indices_shape);
-        StorageIndexNDArray inner_indices(src.innerIndexPtr(), 1, data_shape);
+        const size_t outer_indices_shape[] = { (size_t)((row_major ? rows : cols) + 1) };
+
+        T *src = std::addressof(const_cast<T &>(v));
+        object owner;
+        if (policy == rv_policy::move) {
+            src = new T(std::move(v));
+            owner = capsule(src, [](void *p) noexcept { delete (T *) p; });
+        }
+
+        ScalarNDArray data(src->valuePtr(), 1, data_shape, owner);
+        StorageIndexNDArray outer_indices(src->outerIndexPtr(), 1, outer_indices_shape, owner);
+        StorageIndexNDArray inner_indices(src->innerIndexPtr(), 1, data_shape, owner);
 
         try {
             return matrix_type(make_tuple(
                                    std::move(data), std::move(inner_indices), std::move(outer_indices)),
-                               make_tuple(v.rows(), v.cols()))
+                               make_tuple(rows, cols))
                 .release();
         } catch (python_error &e) {
             e.restore();
--- a/include/nanobind/eigen/sparse.h
+++ b/include/nanobind/eigen/sparse.h
@@ -21,8 +21,14 @@ NAMESPACE_BEGIN(NB_NAMESPACE)
 
 NAMESPACE_BEGIN(detail)
 
+/// Detect Eigen::SparseMatrix
+template <typename T> constexpr bool is_eigen_sparse_matrix_v =
+    is_eigen_sparse_v<T> &&
+    !std::is_base_of_v<Eigen::SparseMapBase<T, Eigen::ReadOnlyAccessors>, T>;
+
+
 /// Caster for Eigen::SparseMatrix
-template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
+template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_matrix_v<T>>> {
     using Scalar = typename T::Scalar;
     using StorageIndex = typename T::StorageIndex;
     using Index = typename T::Index;
@@ -135,6 +141,35 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
             return handle();
         }
     }
+};
+
+
+/// Caster for Eigen::Map<Eigen::SparseMatrix>
+template <typename T>
+struct type_caster<Eigen::Map<T>, enable_if_t<is_eigen_sparse_matrix_v<T>>> {
+    using Map = Eigen::Map<T>;
+    using SparseMatrixCaster = type_caster<T>;
+    static constexpr auto Name = SparseMatrixCaster::Name;
+    template <typename T_> using Cast = Map;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept = delete;
+
+    static handle from_cpp(const Map &v, rv_policy policy, cleanup_list *cleanup) noexcept = delete;
+};
+
+
+/// Caster for Eigen::Ref<Eigen::SparseMatrix>
+template <typename T, int Options>
+struct type_caster<Eigen::Ref<T, Options>, enable_if_t<is_eigen_sparse_matrix_v<T>>> {
+    using Ref = Eigen::Ref<T, Options>;
+    using Map = Eigen::Map<T, Options>;
+    using MapCaster = make_caster<Map>;
+    static constexpr auto Name = MapCaster::Name;
+    template <typename T_> using Cast = Ref;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept = delete;
+
+    static handle from_cpp(const Ref &v, rv_policy policy, cleanup_list *cleanup) noexcept = delete;
 };
 
 NAMESPACE_END(detail)
--- a/include/nanobind/eigen/sparse.h
+++ b/include/nanobind/eigen/sparse.h
@@ -1,0 +1,132 @@
+/*
+    nanobind/eigen/sparse.h: type casters for sparse Eigen matrices
+
+    Copyright (c) 2023 Henri Menke and Wenzel Jakob
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <nanobind/ndarray.h>
+#include <nanobind/eigen/dense.h>
+#include <Eigen/SparseCore>
+
+#include <type_traits>
+#include <utility>
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+
+NAMESPACE_BEGIN(detail)
+
+/// Caster for Eigen::SparseMatrix
+template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
+    using Scalar = typename T::Scalar;
+    using StorageIndex = typename T::StorageIndex;
+    using Index = typename T::Index;
+    using SparseMap = Eigen::Map<T>;
+
+    static_assert(std::is_same_v<T, Eigen::SparseMatrix<Scalar, T::Options, StorageIndex>>,
+                  "nanobind: Eigen sparse caster only implemented for matrices");
+
+    static constexpr bool row_major = T::IsRowMajor;
+
+    using ScalarNDArray = ndarray<numpy, Scalar, shape<any>>;
+    using StorageIndexNDArray = ndarray<numpy, StorageIndex, shape<any>>;
+
+    using ScalarCaster = make_caster<ScalarNDArray>;
+    using StorageIndexCaster = make_caster<StorageIndexNDArray>;
+
+    NB_TYPE_CASTER(T, const_name<row_major>("scipy.sparse.csr_matrix[",
+                                            "scipy.sparse.csc_matrix[")
+                   + make_caster<Scalar>::Name + const_name("]"));
+
+    ScalarCaster data_caster;
+    StorageIndexCaster indices_caster, indptr_caster;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
+        object obj = borrow(src);
+        try {
+            object matrix_type = module_::import_("scipy.sparse").attr(row_major ? "csr_matrix" : "csc_matrix");
+            if (!obj.type().is(matrix_type))
+                obj = matrix_type(obj);
+        } catch (const python_error &) {
+            return false;
+        }
+
+        if (object data_o = obj.attr("data"); !data_caster.from_python(data_o, flags, cleanup))
+            return false;
+        ScalarNDArray& values = data_caster.value;
+
+        if (object indices_o = obj.attr("indices"); !indices_caster.from_python(indices_o, flags, cleanup))
+            return false;
+        StorageIndexNDArray& inner_indices = indices_caster.value;
+
+        if (object indptr_o = obj.attr("indptr"); !indptr_caster.from_python(indptr_o, flags, cleanup))
+            return false;
+        StorageIndexNDArray& outer_indices = indptr_caster.value;
+
+        object shape_o = obj.attr("shape"), nnz_o = obj.attr("nnz");
+        Index rows, cols, nnz;
+        try {
+            if (len(shape_o) != 2)
+                return false;
+            rows = cast<Index>(shape_o[0]);
+            cols = cast<Index>(shape_o[1]);
+            nnz = cast<Index>(nnz_o);
+        } catch (const python_error &) {
+            return false;
+        }
+
+        value = SparseMap(rows, cols, nnz, outer_indices.data(), inner_indices.data(), values.data());
+
+        return true;
+    }
+
+    static handle from_cpp(T &&v, rv_policy policy, cleanup_list *cleanup) noexcept {
+        if (policy == rv_policy::automatic ||
+            policy == rv_policy::automatic_reference)
+            policy = rv_policy::move;
+
+        return from_cpp((const T &) v, policy, cleanup);
+    }
+
+    static handle from_cpp(const T &v, rv_policy policy, cleanup_list *cleanup) noexcept {
+        if (!v.isCompressed()) {
+            PyErr_SetString(PyExc_ValueError,
+                            "nanobind: unable to return an Eigen sparse matrix that is not in a compressed format. "
+                            "Please call `.makeCompressed()` before returning the value on the C++ end.");
+            return handle();
+        }
+        T & src = const_cast<T &>(v);
+
+        object matrix_type;
+        try {
+            matrix_type = module_::import_("scipy.sparse").attr(row_major ? "csr_matrix" : "csc_matrix");
+        } catch (python_error &e) {
+            e.restore();
+            return handle();
+        }
+
+        const size_t data_shape[] = { (size_t)v.nonZeros() };
+        const size_t outer_indices_shape[] = { (size_t)((row_major ? v.rows() : v.cols()) + 1) };
+        ScalarNDArray data(src.valuePtr(), 1, data_shape);
+        StorageIndexNDArray outer_indices(src.outerIndexPtr(), 1, outer_indices_shape);
+        StorageIndexNDArray inner_indices(src.innerIndexPtr(), 1, data_shape);
+
+        try {
+            return matrix_type(make_tuple(
+                                   std::move(data), std::move(inner_indices), std::move(outer_indices)),
+                               make_tuple(v.rows(), v.cols()))
+                .release();
+        } catch (python_error &e) {
+            e.restore();
+            return handle();
+        }
+    }
+};
+
+NAMESPACE_END(detail)
+
+NAMESPACE_END(NB_NAMESPACE)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
     - name: Install NumPy
       if: matrix.python != 'pypy3.9' && matrix.python != '3.12.0-alpha.5'
       run: |
-        python -m pip install numpy
+        python -m pip install numpy scipy
 
     - name: Configure
       run: >
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -60,6 +60,9 @@ Version 0.2.0 (TBA)
 * Added casters for dense matrix/array types from the `Eigen library
   <https://eigen.tuxfamily.org/index.php?title=Main_Page>`__. (PR `#120
   <https://github.com/wjakob/nanobind/pull/120>`__).
+* Added casters for sparse matrix/array types from the `Eigen library
+  <https://eigen.tuxfamily.org/index.php?title=Main_Page>`__. (PR `#126
+  <https://github.com/wjakob/nanobind/pull/126>`_).
 * Implemented `nb::bind_vector\<T\>() <bind_vector>` analogous to similar
   functionality in pybind11. (commit `f2df8a
   <https://github.com/wjakob/nanobind/commit/f2df8a90fbfb06ee03a79b0dd85fa0e266efeaa9>`__).
--- a/docs/eigen.rst
+++ b/docs/eigen.rst
@@ -100,3 +100,16 @@ apply:
   .. code-block:: cpp
 
      void f4(nb::DRef<Eigen::MatrixXf> x) { x *= 2; }
+
+
+Sparse matrices
+---------------
+
+There is also support for Eigen's sparse matrices which are mapped to
+``scipy.sparse.csr_matrix``/``scipy.sparse.csc_matrix``. To use it the extra header
+:file:`nanobind/eigen/sparse.h` has to be included.
+
+.. note::
+
+    There is no support for Eigen sparse vectors because there exists no
+    equivalent type in ``scipy.sparse``.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -100,6 +100,7 @@ discourse:
    ownership_adv
    lowlevel
    typeslots
+   eigen
 
 .. toctree::
    :caption: API Reference
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,7 +21,7 @@ nanobind_add_module(test_intrusive_ext test_intrusive.cpp object.cpp object.h ${
 nanobind_add_module(test_exception_ext test_exception.cpp object.cpp object.h ${NB_EXTRA_ARGS})
 nanobind_add_module(test_make_iterator_ext test_make_iterator.cpp ${NB_EXTRA_ARGS})
 
-find_package (Eigen3 NO_MODULE)
+find_package (Eigen3 3.3.1 NO_MODULE)
 if (TARGET Eigen3::Eigen)
   nanobind_add_module(test_eigen_ext test_eigen.cpp ${NB_EXTRA_ARGS})
   target_link_libraries(test_eigen_ext PRIVATE Eigen3::Eigen)
--- a/tests/test_eigen.cpp
+++ b/tests/test_eigen.cpp
@@ -106,10 +106,16 @@ NB_MODULE(test_eigen_ext, m) {
     m.def("sparse_r", [mat]() -> SparseMatrixR {
         return Eigen::SparseView<Eigen::MatrixXf>(mat);
     });
-    m.def("sparse_c",
-          [mat]() -> SparseMatrixC { return Eigen::SparseView<Eigen::MatrixXf>(mat); });
+    m.def("sparse_c", [mat]() -> SparseMatrixC {
+        return Eigen::SparseView<Eigen::MatrixXf>(mat);
+    });
     m.def("sparse_copy_r", [](const SparseMatrixR &m) -> SparseMatrixR { return m; });
     m.def("sparse_copy_c", [](const SparseMatrixC &m) -> SparseMatrixC { return m; });
+    m.def("sparse_r_uncompressed", []() -> SparseMatrixR {
+        SparseMatrixR m(2,2);
+        m.coeffRef(0,0) = 1.0f;
+        return m;
+    });
 
     struct Buffer {
         uint32_t x[30] { };
--- a/tests/test_eigen.cpp
+++ b/tests/test_eigen.cpp
@@ -1,4 +1,5 @@
 #include <nanobind/eigen/dense.h>
+#include <nanobind/eigen/sparse.h>
 
 namespace nb = nanobind;
 
@@ -92,6 +93,23 @@ NB_MODULE(test_eigen_ext, m) {
 
     m.def("updateV3i", [](Eigen::Ref<Eigen::Vector3i> a) { a[2] = 123; });
     m.def("updateVXi", [](Eigen::Ref<Eigen::VectorXi> a) { a[2] = 123; });
+
+    using SparseMatrixR = Eigen::SparseMatrix<float, Eigen::RowMajor>;
+    using SparseMatrixC = Eigen::SparseMatrix<float>;
+    Eigen::MatrixXf mat(5, 6);
+    mat <<
+	 0, 3,  0, 0,  0, 11,
+	22, 0,  0, 0, 17, 11,
+	 7, 5,  0, 1,  0, 11,
+	 0, 0,  0, 0,  0, 11,
+	 0, 0, 14, 0,  8, 11;
+    m.def("sparse_r", [mat]() -> SparseMatrixR {
+        return Eigen::SparseView<Eigen::MatrixXf>(mat);
+    });
+    m.def("sparse_c",
+          [mat]() -> SparseMatrixC { return Eigen::SparseView<Eigen::MatrixXf>(mat); });
+    m.def("sparse_copy_r", [](const SparseMatrixR &m) -> SparseMatrixR { return m; });
+    m.def("sparse_copy_c", [](const SparseMatrixC &m) -> SparseMatrixC { return m; });
 
     struct Buffer {
         uint32_t x[30] { };
--- a/tests/test_eigen.py
+++ b/tests/test_eigen.py
@@ -172,3 +172,28 @@ def test07_mutate_arg():
     A2 = A.copy()
     t.mutate_MXu(A)
     assert np.all(A == 2*A2)
+
+
+@needs_numpy_and_eigen
+def test_sparse():
+    pytest.importorskip("scipy")
+    ref = np.array(
+        [
+            [0.0, 3, 0, 0, 0, 11],
+            [22, 0, 0, 0, 17, 11],
+            [7, 5, 0, 1, 0, 11],
+            [0, 0, 0, 0, 0, 11],
+            [0, 0, 14, 0, 8, 11],
+        ]
+    )
+    def assert_equal_ref(mat):
+        nonlocal ref
+        np.testing.assert_array_equal(mat, ref)
+    def assert_sparse_equal_ref(sparse_mat):
+        assert_equal_ref(sparse_mat.toarray())
+    assert_sparse_equal_ref(t.sparse_r())
+    assert_sparse_equal_ref(t.sparse_c())
+    assert_sparse_equal_ref(t.sparse_copy_r(t.sparse_r()))
+    assert_sparse_equal_ref(t.sparse_copy_c(t.sparse_c()))
+    assert_sparse_equal_ref(t.sparse_copy_r(t.sparse_c()))
+    assert_sparse_equal_ref(t.sparse_copy_c(t.sparse_r()))
--- a/tests/test_eigen.py
+++ b/tests/test_eigen.py
@@ -1,5 +1,7 @@
 import pytest
 import gc
+import re
+import sys
 
 try:
     import numpy as np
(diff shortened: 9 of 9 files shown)


include/nanobind/eigen/dense.h
```diff
@@ -12,9 +12,10 @@
 
 #include <nanobind/tensor.h>
 #include <Eigen/Core>
+#include <Eigen/SparseCore>
```
wjakob, line 15: Let's put the whole sparse-specific parts into a new headerfile `nanobind/eigen/sparse.h` that includes `nanobind/eigen/dense.h`

wjakob: This means that `is_eigen_sparse_v` and `is_eigen_xpr_v` have to be defined in a way that doesn't depend on elements in `Eigen/SparseCore` -- perhaps there is a constexpr field like `IsSparse` that can be used instead (or another field that is only defined for sparse types).

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
     operator Ref() { return Ref(caster.operator Map()); }
 };
 
+/// Caster for Eigen::SparseMatrix<T>
+template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
+    using Scalar = typename T::Scalar;
+    using StorageIndex = std::remove_reference_t<decltype(*std::declval<T>().outerIndexPtr())>;
```
wjakob, line 255: That definition of `StorageIndex` strikes me as surprisingly complicated. Can we not get the type more directly from a `using` alias defined within `T` (e.g. `T::StorageIndex`?)

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+    NB_TYPE_CASTER(T, const_name<rowMajor>("scipy.sparse.csr_matrix[",
+					   "scipy.sparse.csc_matrix[")
+		   + make_caster<Scalar>::Name + const_name("]"));
+
+    ScalarCaster scalar_caster;
+    StorageIndexCaster storageindex_caster;
+    IndexCaster index_caster;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
+	if (!src) {
```
wjakob, line 272: those three lines are not needed in nanobind.

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+	    return false;
+	}
+
+	auto obj = borrow<object>(src);
+	object sparse_module = module_::import_("scipy.sparse");
+	object matrix_type = sparse_module.attr(rowMajor ? "csr_matrix" : "csc_matrix");
+
+	if (!obj.type().is(matrix_type)) {
+            try {
```
wjakob, line 281: expand this `try` block so that it also includes the `module::import_`, etc. bits. These can raise exceptions, and the type caster is *not* allowed to raise.

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
     operator Ref() { return Ref(caster.operator Map()); }
 };
 
+/// Caster for Eigen::SparseMatrix<T>
+template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
+    using Scalar = typename T::Scalar;
+    using StorageIndex = std::remove_reference_t<decltype(*std::declval<T>().outerIndexPtr())>;
+    using Index = typename T::Index;
+    static constexpr bool rowMajor = T::IsRowMajor;
```
wjakob, line 257: use `underscore_case`

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+
+	if (!obj.type().is(matrix_type)) {
+            try {
+                obj = matrix_type(obj);
+            } catch (const python_error &) {
+                return false;
+            }
+        }
+
+	if (!scalar_caster.from_python(obj.attr("data"), flags, cleanup)) {
```
wjakob, line 288: I would save the output of `obj.attr("data")` into a `object` variable to ensure that the reference is kept alive while `from_python` runs (you wrote code that does an implicit cast to a `handle` that doesn't hold a reference, and that may cause undefined behavior)

Applies to other casts below as well.

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+    IndexCaster index_caster;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
+	if (!src) {
+	    return false;
+	}
+
+	auto obj = borrow<object>(src);
+	object sparse_module = module_::import_("scipy.sparse");
+	object matrix_type = sparse_module.attr(rowMajor ? "csr_matrix" : "csc_matrix");
```
wjakob, line 278: You aren't using the `sparse_module` variable elsewhere, so I'd merge those two lines (`module::import_(..).attr(..)`).

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
 };
 
+/// Caster for Eigen::SparseMatrix<T>
+template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
+    using Scalar = typename T::Scalar;
+    using StorageIndex = std::remove_reference_t<decltype(*std::declval<T>().outerIndexPtr())>;
+    using Index = typename T::Index;
+    static constexpr bool rowMajor = T::IsRowMajor;
+
+    using ScalarCaster = make_caster<tensor<Scalar>>;
```
wjakob, line 259: Your specification seems too general (we wouldn't want this to accept a matrix, non-contiguous representations, or GPU arrays). I would write
```cpp
using ScalarTensor = tensor<Scalar, shape<any>, c_contig, cpu>;
```

and then a few lines below:
```
make_caster<ScalarTensor> scalar_caster;
```

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+
+	if (!scalar_caster.from_python(obj.attr("data"), flags, cleanup)) {
+	    return false;
+	}
+        auto values = scalar_caster.value;
+
+	if (!storageindex_caster.from_python(obj.attr("indices"), flags, cleanup)) {
+	    return false;
+	}
+        auto innerIndices = storageindex_caster.value;
```
wjakob, line 296: use `undescore_case` for variable names.

I really don't like `auto` when it is avoidable -- can you use a type definition? (`ScalarTensor&`, `IndexTensor&`, etc.)

Both comments apply to a few other places as well.

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+
+	if (!scalar_caster.from_python(obj.attr("indptr"), flags, cleanup)) {
```
wjakob, line 298: Minor style nit pick: throughout the codebase, short 1-line if statements are written

```cpp
if (x)
    return false;
```

as opposed to

```cpp
if (x) {
    return false;
}
```

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+	    return false;
+	}
+        auto outerIndices = storageindex_caster.value;
+
+	// FIXME: Write with caster.from_python
+	auto shape = cast<std::tuple<Index, Index>>(obj.attr("shape"));
```
wjakob, line 304: I am surprised that this works, because the `std::tuple` type caster hasn't been imported (and that seems like a rather heavy dependency). We really don't want to pull in any STL type casters since that may lead to unexpected consequences in user code.

Can you do it manually, e.g.:
```cpp
object shape_o = obj.attr("shape"), nnz_o = obj.attr("nnz");
int rows, cols, nnz;
try {
    if (len(shape_o) != 2)
       return false;
    rows = cast<int>(shape_o[0]);
    cols = cast<int>(shape_o[1]);
    nnz = cast<Index>(nnz_o);
} catch (const python_error &) {
    return false;
}
```

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+
+	if (!index_caster.from_python(obj.attr("nnz"), flags, cleanup)) {
+	    return false;
+	}
+        auto nnz = index_caster.value;
+
+        value = EigenMapSparseMatrix<Scalar,
```
wjakob, line 311: I don't remember all the details of the pybind11 version of this code, but a general question: who will keep this `Eigen::Map` alive? Is the idea that this will immediately be cast into a concrete Eigen sparse type that makes a copy?

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+    static handle from_cpp(T &&v, rv_policy policy, cleanup_list *cleanup) noexcept {
+        if (policy == rv_policy::automatic ||
+            policy == rv_policy::automatic_reference)
+            policy = rv_policy::move;
+
+        return from_cpp((const T &) v, policy, cleanup);
+    }
+
+    static handle from_cpp(const T &v, rv_policy policy, cleanup_list *cleanup) noexcept {
+	// FIXME: Looks like UB. Is there anything that can be done about this?
```
wjakob, line 332: did pybind11 use the same code?

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+	T & src = const_cast<T &>(v);
+	src.makeCompressed();
+
+	object sparse_module = module_::import_("scipy.sparse");
```
wjakob, line 336: merge with line below

include/nanobind/eigen/dense.h
```diff
@@ -241,6 +249,107 @@ struct type_caster<Eigen::Ref<T, Options, StrideType>> {
...
+	object matrix_type = sparse_module.attr(rowMajor ? "csr_matrix" : "csc_matrix");
+
+	// FIXME: These casts are ugly! Is there another way to write this?
+	const size_t data_shape[] = { static_cast<size_t>(v.nonZeros()) };
```
wjakob, line 340: using `static_cast` for an integer conversion seems quite overkill.

wjakob: Is there something else that bothers you here?

wjakob (commented):
Thanks for the PR! Here is a first bunch of feedback.

include/nanobind/eigen/dense.h
hmenke: I've moved everything to `nanobind/eigen/sparse.h`, but `is_eigen_sparse_v` cannot be moved (yet) unfortunately, because `is_eigen_v` matches `Eigen::MatrixBase` which `Eigen::SparseMatrix` inherits from. There is no member like `IsSparse` by we could maybe try to detect the `valuePtr()`, `outerIndexPtr()`, and `innerIndexPtr()` methods to decide. This would however entail SFINAE which is slow to compile

include/nanobind/eigen/dense.h
hmenke: Either assignment to an `Eigen::SparseMatrix` for copy or possibly assignment to another `Eigen::Map<Eigen::SparseMatrix>` for zero-copy (if that works?). If the latter does not work, then using an owning `Eigen::SparseMatrix` here will be better. I also don't know the pybind11 code, I just took “inspiration” from it :)

include/nanobind/eigen/dense.h
hmenke: Yes. https://github.com/pybind/pybind11/blob/master/include/pybind11/eigen/matrix.h#L677

include/nanobind/eigen/dense.h
hmenke: The cast is needed to narrow the results of the calls to `size_t` which is required for the `nb::tensor` constructor. Leaving out the cast generates a compiler warning in GCC and an error in Clang:
```
nanobind/include/nanobind/eigen/sparse.h:115:32: error: non-constant-expression cannot be narrowed from type 'Eigen::Index' (aka 'long') to 'size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]
        const size_t data_shape[] = { v.nonZeros() };
                                      ^~~~~~~~~~~~
```
Why do you think `static_cast` is overkill? What's the alternative?

hmenke:
Thank you very much for the thorough review. I've pushed a fixup commit which I will later rebase, but for the sake of review I will leave it split for now to better follow the progress.

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,125 @@
...
+	const size_t outerIndices_shape[] = { static_cast<size_t>((row_major ? v.rows() : v.cols()) + 1) };
+        ScalarTensor data(src.valuePtr(), 1, data_shape);
+        StorageIndexTensor outerIndices(src.outerIndexPtr(), 1, outerIndices_shape);
+        StorageIndexTensor innerIndices(src.innerIndexPtr(), 1, data_shape);
+
+        return matrix_type(make_tuple(
+                               std::move(data), std::move(innerIndices), std::move(outerIndices)),
+                           make_tuple(v.rows(), v.cols()))
+            .release();
+    }
```
hmenke, line 120: Currently the test segfaults in the return statement, but I have no idea where this comes from.

<details>

<summary>Backtrace</summary>

```cpp
Thread 1 (Thread 0x7ffff7891740 (LWP 143127) "python"):
#0  0x00007ffff791ebc7 in __pthread_kill_implementation () from /nix/store/9xfad3b5z4y00mzmk2wnn4900q0qmxns-glibc-2.35-224/lib/libc.so.6
No symbol table info available.
#1  0x00007ffff78d1b46 in raise () from /nix/store/9xfad3b5z4y00mzmk2wnn4900q0qmxns-glibc-2.35-224/lib/libc.so.6
No symbol table info available.
#2  0x00007ffff78bc4b5 in abort () from /nix/store/9xfad3b5z4y00mzmk2wnn4900q0qmxns-glibc-2.35-224/lib/libc.so.6
No symbol table info available.
#3  0x00007fffae4899fd in __gnu_cxx::__verbose_terminate_handler() [clone .cold] () from /home/henri/Code/nanobind/build/tests/test_eigen_ext.cpython-310-x86_64-linux-gnu.so
No symbol table info available.
#4  0x00007fffae4ffdca in __cxxabiv1::__terminate(void (*)()) () from /home/henri/Code/nanobind/build/tests/test_eigen_ext.cpython-310-x86_64-linux-gnu.so
No symbol table info available.
#5  0x00007fffae4ffe35 in std::terminate() () from /home/henri/Code/nanobind/build/tests/test_eigen_ext.cpython-310-x86_64-linux-gnu.so
No symbol table info available.
#6  0x00007fffae49633f in __clang_call_terminate () from /home/henri/Code/nanobind/build/tests/test_eigen_ext.cpython-310-x86_64-linux-gnu.so
No symbol table info available.
#7  0x00007fffae4b3878 in nanobind::detail::type_caster<Eigen::SparseMatrix<float, 1, int>, int>::from_cpp (v=..., policy=nanobind::rv_policy::move, cleanup=0x7fffffff2bb0) at /home/henri/Code/nanobind/include/nanobind/eigen/sparse.h:107
        src = @0x7fffffff29c8: {
          <Eigen::SparseCompressedBase<Eigen::SparseMatrix<float, 1, int> >> = {
            <Eigen::SparseMatrixBase<Eigen::SparseMatrix<float, 1, int> >> = {
              <Eigen::EigenBase<Eigen::SparseMatrix<float, 1, int> >> = {<No data fields>}, 
              members of Eigen::SparseMatrixBase<Eigen::SparseMatrix<float, 1, int> >:
              m_isRValue = false
            }, <No data fields>}, 
          members of Eigen::SparseMatrix<float, 1, int>:
          m_outerSize = 5,
          m_innerSize = 6,
          m_outerIndex = 0xabc890,
          m_innerNonZeros = 0x0,
          m_data = {
            m_values = 0xac7f30,
            m_indices = 0xacd210,
            m_size = 13,
            m_allocatedSize = 13
          }
        }
        matrix_type = {
          <nanobind::handle> = {
            <nanobind::detail::api<nanobind::handle>> = {
              <nanobind::detail::api_tag> = {<No data fields>}, <No data fields>}, 
            members of nanobind::handle:
            static Name = <optimized out>,
            m_ptr = 0xae74c0
          }, 
          members of nanobind::object:
          static Name = <optimized out>
        }
        data_shape = {13}
        outerIndices_shape = {6}
        data = {
          m_handle = 0x7fffae42b970,
          m_tensor = {
            data = 0xac7f30,
            device = {
              device_type = 1,
              device_id = 0
            },
            ndim = 1,
            dtype = {
              code = 2 '\002',
              bits = 32 ' ',
              lanes = 1
            },
            shape = 0x7fff6ba88150,
            strides = 0x7fff6ba88120,
            byte_offset = 0
          }
        }
        outerIndices = {
          m_handle = 0x7fffae42b790,
          m_tensor = {
            data = 0xabc890,
            device = {
              device_type = 1,
              device_id = 0
            },
            ndim = 1,
            dtype = {
              code = 0 '\000',
              bits = 32 ' ',
              lanes = 1
            },
            shape = 0x7fff6ba88140,
            strides = 0x7fff6ba88160,
            byte_offset = 0
          }
        }
        innerIndices = {
          m_handle = 0x7fffae42b870,
          m_tensor = {
            data = 0xacd210,
            device = {
              device_type = 1,
              device_id = 0
            },
            ndim = 1,
            dtype = {
              code = 0 '\000',
              bits = 32 ' ',
              lanes = 1
            },
            shape = 0x7fff6ba88170,
            strides = 0x7fff6ba88180,
            byte_offset = 0
          }
        }
#8  0x00007fffae4b31be in nanobind::detail::type_caster<Eigen::SparseMatrix<float, 1, int>, int>::from_cpp (v=..., policy=nanobind::rv_policy::move, cleanup=0x7fffffff2bb0) at /home/henri/Code/nanobind/include/nanobind/eigen/sparse.h:99
No locals.
#9  0x00007fffae494d9f in nanobind::detail::func_create<false, true, nanobind_init_test_eigen_ext(nanobind::module_&)::$_19, Eigen::SparseMatrix<float, 1, int>, , , nanobind::scope, nanobind::name>(nanobind_init_test_eigen_ext(nanobind::module_&)::$_19&&, Eigen::SparseMatrix<float, 1, int> (*)(), std::integer_sequence<unsigned long>, nanobind::scope const&, nanobind::name const&)::{lambda(void*, _object**, unsigned char*, nanobind::rv_policy, nanobind::detail::cleanup_list*)#1}::operator()(void*, _object**, unsigned char*, nanobind::rv_policy, nanobind::detail::cleanup_list*) const (this=0x0, p=0x7fffaee26468, args=0x7fffaee72538, args_flags=0x7fffffff2bf0 "\001\001\001\001\001\001\001\001", policy=nanobind::rv_policy::automatic, cleanup=0x7fffffff2bb0) at /home/henri/Code/nanobind/include/nanobind/nb_func.h:128
        cap = 0x7fffaee26468
        result = 0x7ffff660d3b0
#10 nanobind::detail::func_create<false, true, nanobind_init_test_eigen_ext(nanobind::module_&)::$_19, Eigen::SparseMatrix<float, 1, int>, , , nanobind::scope, nanobind::name>(nanobind_init_test_eigen_ext(nanobind::module_&)::$_19&&, Eigen::SparseMatrix<float, 1, int> (*)(), std::integer_sequence<unsigned long>, nanobind::scope const&, nanobind::name const&)::{lambda(void*, _object**, unsigned char*, nanobind::rv_policy, nanobind::detail::cleanup_list*)#1}::__invoke(void*, _object**, unsigned char*, nanobind::rv_policy, nanobind::detail::cleanup_list*) (p=0x7fffaee26468, args=0x7fffaee72538, args_flags=0x7fffffff2bf0 "\001\001\001\001\001\001\001\001", policy=nanobind::rv_policy::automatic, cleanup=0x7fffffff2bb0) at /home/henri/Code/nanobind/include/nanobind/nb_func.h:105
No locals.
#11 0x00007fffae4c4f3f in nanobind::detail::nb_func_vectorcall_simple (self=0x7fffaee26440, args_in=0x7fffaee72538, nargsf=9223372036854775808, kwargs_in=0x0) at /home/henri/Code/nanobind/src/nb_func.cpp:732
        f = 0x7fffaee26468
        e = @0x40d0f0: {
          <std::exception> = {<No data fields>}, 
          members of nanobind::python_error:
          m_type = 0x0,
          m_value = 0x439fb0,
          m_trace = 0x7fffaee723b0,
          m_what = 0x28 <error: Cannot access memory at address 0x28>
        }
        k = 0
        pass = 1
        args_flags = "\001\001\001\001\001\001\001\001"
        fr = 0x7fffaee26468
        count = 1
        nargs_in = 0
        is_method = false
        is_constructor = false
        self_flags = 0
        result = 0x0
        self_arg = 0x0
        cleanup = {
          static Small = 6,
          m_size = 1,
          m_capacity = 6,
          m_data = 0x7fffffff2bc0,
          m_local = {0x0, 0xac2aa7be639ae800, 0x40d0f0, 0x7fffaee723b0, 0x40d0f0, 0x7ffff66483b0}
        }
        error_handler = 0x0
        fail = false
        none_ptr = 0x7ffff7fafe30 <_Py_NoneStruct>
#12 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#13 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#14 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#15 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#16 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#17 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#18 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#19 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#20 0x00007ffff7cf7cd8 in method_vectorcall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#21 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#22 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#23 0x00007ffff7d02736 in _PyObject_Call_Prepend () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#24 0x00007ffff7d43b49 in slot_tp_call () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#25 0x00007ffff7c95c84 in _PyObject_MakeTpCall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#26 0x00007ffff7c47f69 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#27 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#28 0x00007ffff7c475b9 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#29 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#30 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#31 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#32 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#33 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#34 0x00007ffff7cf7cd8 in method_vectorcall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#35 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#36 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#37 0x00007ffff7d02736 in _PyObject_Call_Prepend () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#38 0x00007ffff7d43b49 in slot_tp_call () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#39 0x00007ffff7ca7b39 in PyObject_Call () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#40 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#41 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#42 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#43 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#44 0x00007ffff7cf7cd8 in method_vectorcall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#45 0x00007ffff7c46344 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#46 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#47 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#48 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#49 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#50 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#51 0x00007ffff7c46344 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#52 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#53 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#54 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#55 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#56 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#57 0x00007ffff7cf7cd8 in method_vectorcall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#58 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#59 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#60 0x00007ffff7d02736 in _PyObject_Call_Prepend () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#61 0x00007ffff7d43b49 in slot_tp_call () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#62 0x00007ffff7c95c84 in _PyObject_MakeTpCall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#63 0x00007ffff7c47f69 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#64 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#65 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#66 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#67 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#68 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#69 0x00007ffff7cf7cd8 in method_vectorcall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#70 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#71 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#72 0x00007ffff7d02736 in _PyObject_Call_Prepend () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#73 0x00007ffff7d43b49 in slot_tp_call () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#74 0x00007ffff7c95c84 in _PyObject_MakeTpCall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#75 0x00007ffff7c47f69 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#76 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#77 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#78 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#79 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#80 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#81 0x00007ffff7c41169 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#82 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#83 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#84 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#85 0x00007ffff7cf7cd8 in method_vectorcall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#86 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#87 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#88 0x00007ffff7d02736 in _PyObject_Call_Prepend () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#89 0x00007ffff7d43b49 in slot_tp_call () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#90 0x00007ffff7c95c84 in _PyObject_MakeTpCall () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#91 0x00007ffff7c47f69 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#92 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#93 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#94 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#95 0x00007ffff7c49cf2 in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#96 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#97 0x00007ffff7dc18e8 in PyEval_EvalCode () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#98 0x00007ffff7e590ed in builtin_exec () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#99 0x00007ffff7db69bf in cfunction_vectorcall_FASTCALL () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#100 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#101 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#102 0x00007ffff7c4669f in _PyEval_EvalFrameDefault () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#103 0x00007ffff7dc127f in _PyEval_Vector () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#104 0x00007ffff7e4e5c7 in pymain_run_module () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#105 0x00007ffff7e59f0b in Py_RunMain () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#106 0x00007ffff7e5a535 in Py_BytesMain () from /nix/store/5axq6aw8j3vcs2m7gi440cwpcckl7ql9-python3-3.10.9/lib/libpython3.10.so.1.0
No symbol table info available.
#107 0x00007ffff78bd24e in __libc_start_call_main () from /nix/store/9xfad3b5z4y00mzmk2wnn4900q0qmxns-glibc-2.35-224/lib/libc.so.6
No symbol table info available.
#108 0x00007ffff78bd309 in __libc_start_main_impl () from /nix/store/9xfad3b5z4y00mzmk2wnn4900q0qmxns-glibc-2.35-224/lib/libc.so.6
No symbol table info available.
#109 0x0000000000401075 in _start ()
No symbol table info available.
```

</details>

hmenke: Ah no, wait. It looks like `module_::import_("scipy.sparse")` throws and causes terminate.

include/nanobind/eigen/dense.h
wjakob: Just a c-style cast. I think static_cast is great when you have polymorphic classes and other complex aspects of C++ and want to be extra-clear about what kind of cast is acceptable. But I don't see the advantage for simple integer casts, and it just makes the code less readable IMO.

include/nanobind/eigen/dense.h
wjakob: `SparseMatrixBase` has a forward declaration in `Eigen/src/util/ForwardDeclarations.h`. Isn't it possible to do the test without pulling in the whole `SparseCore` header?

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,125 @@
...
+		   + make_caster<Scalar>::Name + const_name("]"));
+
+    ScalarCaster scalar_caster;
+    StorageIndexCaster storageindex_caster;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
+	auto obj = borrow<object>(src);
+	try {
+	    object matrix_type = module_::import_("scipy.sparse").attr(row_major ? "csr_matrix" : "csc_matrix");
+	    if (!obj.type().is(matrix_type)) {
```
wjakob, line 51: can skip the extra braces here.

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,125 @@
...
+
+    using ScalarTensor = tensor<Scalar, shape<any>, c_contig, device::cpu>;
+    using StorageIndexTensor = tensor<StorageIndex, shape<any>, c_contig, device::cpu>;
+
+    using ScalarCaster = make_caster<ScalarTensor>;
+    using StorageIndexCaster = make_caster<StorageIndexTensor>;
+
+    NB_TYPE_CASTER(T, const_name<row_major>("scipy.sparse.csr_matrix[",
+					    "scipy.sparse.csc_matrix[")
+		   + make_caster<Scalar>::Name + const_name("]"));
```
wjakob, line 42: `make_caster<Scalar>` -> `ScalarCaster`

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,125 @@
...
+	    rows = cast<Index>(shape_o[0]);
+	    cols = cast<Index>(shape_o[1]);
+	    nnz = cast<Index>(nnz_o);
+	} catch (const python_error &) {
+	    return false;
+	}
+
+        value = EigenMapSparseMatrix<Scalar,
+                                     T::Flags &(Eigen::RowMajor | Eigen::ColMajor),
+                                     StorageIndex>(std::move(rows),
```
wjakob, line 84: Let's remove the use of ``std::move`` to pass integer values here.

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,125 @@
...
+	T & src = const_cast<T &>(v);
+	src.makeCompressed();
+
+	object matrix_type = module_::import_("scipy.sparse").attr(row_major ? "csr_matrix" : "csc_matrix");
+
+	// FIXME: These casts are ugly! Is there another way to write this?
+	const size_t data_shape[] = { static_cast<size_t>(v.nonZeros()) };
+	const size_t outerIndices_shape[] = { static_cast<size_t>((row_major ? v.rows() : v.cols()) + 1) };
+        ScalarTensor data(src.valuePtr(), 1, data_shape);
+        StorageIndexTensor outerIndices(src.outerIndexPtr(), 1, outerIndices_shape);
```
wjakob, line 113: `outerIndices` -> `outer_indices`.

wjakob (commented):
This looks great. Just a few minor extra comments.

include/nanobind/eigen/sparse.h
hmenke: But `ScalarCaster = make_caster<ScalarTensor> ≠ make_caster<Scalar>`. I think we do want `make_caster<Scalar>::Name` here.

wjakob: my bad.

wjakob:
It would still be good to figure out a way to make `nanobind/eigen/dense.h` not depend on `Eigen/SparseCore` (see my earlier comment). Following this, would you be able to squash the commits and rebase on `master`?

wjakob:
By the way, you may also need to change `.github/workflows/ci.yml` to install the relevant SciPy packages so that your changes are actually tested.

include/nanobind/eigen/dense.h
hmenke: But according to https://eel.is/c++draft/expr.cast `(T)` will just try in order `const_cast<T>`, `static_cast<T>`, `reinterpret_cast<T>`. See also https://stackoverflow.com/questions/103512/why-use-static-castintx-instead-of-intx. But I'm happy to do C-style casts as well if you wish.

wjakob: Right. In those stackoverflow threads, you can see that there are footguns involving C++ classes, implicit conversions involving them, and polymorphism. None of which applies to a simple integer cast. For me, using static_cast even in such simple situations is an anti-pattern that makes code ugly to read. (This is somewhat subjective, I know that other people may disagree on what a good default policy is).

include/nanobind/eigen/dense.h
hmenke: Yes, this does work indeed. Good find! :ok_hand:

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
+/*
+    nanobind/eigen/dense.h: type casters for dense Eigen
```
wjakob, line 2: update

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+    vectors and matrices
+
+    Copyright (c) 2023 Wenzel Jakob
```
wjakob, line 5: Suggested: ``Copyright (c) 2023 Henri Menke and Wenzel Jakob``

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+              "Eigen matrix support in nanobind requires Eigen >= 3.3.0");
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+
+NAMESPACE_BEGIN(detail)
+
+template <typename Scalar, int Flags, typename StorageIndex>
+using EigenMapSparseMatrix = Eigen::Map<Eigen::SparseMatrix<Scalar, Flags, StorageIndex>>;
+
+/// Caster for Eigen::SparseMatrix<T>
```
wjakob, line 27: Is it really just ``Eigen::SparseMatrix<T>`` or also other types?

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+
+    NB_TYPE_CASTER(T, const_name<row_major>("scipy.sparse.csr_matrix[",
+                                            "scipy.sparse.csc_matrix[")
+                   + make_caster<Scalar>::Name + const_name("]"));
+
+    ScalarCaster scalar_caster;
+    StorageIndexCaster storageindex_caster;
+
+    bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
+        auto obj = borrow<object>(src);
```
wjakob, line 48: The `<object>` template argument is not needed. Replace `auto` with `object`.

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+            if (len(shape_o) != 2)
+                return false;
+            rows = cast<Index>(shape_o[0]);
+            cols = cast<Index>(shape_o[1]);
+            nnz = cast<Index>(nnz_o);
+        } catch (const python_error &) {
+            return false;
+        }
+
+        value = EigenMapSparseMatrix<Scalar,
```
wjakob, line 81: The formatting is a bit unwieldy here. Perhaps you can define this type with a `using SparseMap = ..` declaration at the top of the caster and then reference it here?

hmenke:
Rebased and squashed. Checks are enabled, but as already noted something throws in `from_cpp`, apparently during `make_tuple` in the return statement.

[Backtrace](https://github.com/wjakob/nanobind/files/10743367/gdb.txt)

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+    static handle from_cpp(T &&v, rv_policy policy, cleanup_list *cleanup) noexcept {
+        if (policy == rv_policy::automatic ||
+            policy == rv_policy::automatic_reference)
+            policy = rv_policy::move;
+
+        return from_cpp((const T &) v, policy, cleanup);
+    }
+
+    static handle from_cpp(const T &v, rv_policy policy, cleanup_list *cleanup) noexcept {
+        // FIXME: Looks like UB. Is there anything that can be done about this?
```
wjakob, line 102: This indeed doesn't seem like a good design. I suggest the following instead:

Option 1: give up (untested)

```cpp
if (!v.isCompressed()) {
    PyErr_SetString(PyExc_ValueError, "nanobind: unable to return an Eigen sparse matrix that is not in a compressed format. Please call `.makeCompressed()` before returning the value on the C++ end.");
    return handle();
}
```

Option 2: copy to turn the matrix into a compressed format (untested):

```cpp
std::unique_ptr<T> temp;
T *src = &v;

if (!v.isCompressed()) {
    temp = make_unique<T>(v);
    temp->makeCompressed();
    src = temp.get();
}

```

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+        object matrix_type = module_::import_("scipy.sparse").attr(row_major ? "csr_matrix" : "csc_matrix");
+
+        // FIXME: These casts are ugly! Is there another way to write this?
+        const size_t data_shape[] = { static_cast<size_t>(v.nonZeros()) };
+        const size_t outerIndices_shape[] = { static_cast<size_t>((row_major ? v.rows() : v.cols()) + 1) };
+        ScalarTensor data(src.valuePtr(), 1, data_shape);
+        StorageIndexTensor outer_indices(src.outerIndexPtr(), 1, outerIndices_shape);
+        StorageIndexTensor inner_indices(src.innerIndexPtr(), 1, data_shape);
+
+        return matrix_type(make_tuple(
```
wjakob, line 115: This function call can fail with an exception, and that is not allowed by a caster. Once more, I think there is a whole bunch of code that must be wrapped:

```cpp
try {
    // code that may throw
} catch (const python_error &e) {
    e.restore();
    return handle();
}
```

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,124 @@
...
+using EigenMapSparseMatrix = Eigen::Map<Eigen::SparseMatrix<Scalar, Flags, StorageIndex>>;
+
+/// Caster for Eigen::SparseMatrix<T>
+template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
+    using Scalar = typename T::Scalar;
+    using StorageIndex = typename T::StorageIndex;
+    using Index = typename T::Index;
+    static constexpr bool row_major = T::IsRowMajor;
+
+    using ScalarTensor = tensor<Scalar, shape<any>, c_contig, device::cpu>;
```
wjakob, line 34: I think one crucial thing is still missing in these type annotations: `numpy`! Otherwise, nanobind will call the SciPy sparse matrix constructor with a bunch of DLPack capsules, and SciPy does not know what to do with hose.

hmenke: Thanks! That's it.

wjakob:
Please also add a changelog entry ;)

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,115 @@
...
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <nanobind/tensor.h>
+#include <nanobind/eigen/dense.h>
+#include <Eigen/SparseCore>
+
+static_assert(EIGEN_VERSION_AT_LEAST(3, 3, 0),
```
wjakob, line 17: I don't think you need to repeat the version check, if it is already in `eigen/dense.h`.

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,112 @@
+/*
+    nanobind/eigen/sparse.h: type casters for sparse Eigen
+    vectors and matrices
```
wjakob, line 3: We do not support sparse vectors though, right? I remember there being a pybind11 bug report about it.

include/nanobind/eigen/sparse.h
hmenke: I chose option 1. Time will tell whether that was a good decision.

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,118 @@
...
+			    "Please call `.makeCompressed()` before returning the value on the C++ end.");
+	    return handle();
+	}
+        T & src = const_cast<T &>(v);
+
+        object matrix_type = module_::import_("scipy.sparse").attr(row_major ? "csr_matrix" : "csc_matrix");
+
+        const size_t data_shape[] = { (size_t)v.nonZeros() };
+        const size_t outer_indices_shape[] = { (size_t)((row_major ? v.rows() : v.cols()) + 1) };
+        ScalarTensor data(src.valuePtr(), 1, data_shape);
```
hmenke, line 105: `const_cast` is still needed because `nb::tensor` has no constructor for `const void *`. Is there a deeper reason for that?

wjakob: we currently don't have a way of making read-only tensors, though that is a feature I would like to support in the future.

hmenke:
I have to double check the minimum Eigen version. The current 3.3.0 seems too low (it's also from way back in 2016, i.e. before C++17). In principle I would also like to improve the test coverage a bit. Right now it's limited to the pybind11 tests.

wjakob:
Agreed, the tests seem a bit, ahem -- sparse. ;)

wjakob:
Another aspect I realized today is that one feature in the dense caster is not shared by this sparse version. In the `from_cpp` part, when the return value policy is `move`, nanobind can perform a zero-copy transfer by move-constructing the Eigen object into a new heap-allocated instance, and then tying the lifetime of the newly created NumPy wrapper using a `capsule` owner object.

See the special case here: https://github.com/wjakob/nanobind/blob/master/include/nanobind/eigen/dense.h#L132

include/nanobind/eigen/sparse.h
```diff
@@ -109,16 +109,26 @@ template <typename T> struct type_caster<T, enable_if_t<is_eigen_sparse_v<T>>> {
...
+        T *src = std::addressof(const_cast<T &>(v));
+        object owner;
+        if (policy == rv_policy::move) {
+            src = new T(std::move(v));
+            owner = capsule(src, [](void *p) noexcept { delete (T *) p; });
+        }
+
+        ScalarTensor data(src->valuePtr(), 1, data_shape, owner);
+        StorageIndexTensor outer_indices(src->outerIndexPtr(), 1, outer_indices_shape, owner);
+        StorageIndexTensor inner_indices(src->innerIndexPtr(), 1, data_shape, owner);
```
hmenke, line 126: Do I need one owner per tensor or is this okay? I ran the existing tests with ASan on and there were no failure, but I'm also not sure whether I'm triggering this optimization right now.

wjakob: A single owner shared by all tensors is good. Each tensor will reference the Python capsule, and it can only be deleted when they all expire.

wjakob:
Any further changes planned?

hmenke:
I would like to test returning `Eigen::Ref<Eigen::SparseMatrix>` and  `Eigen::Map<Eigen::SparseMatrix>` from C++. This could also go in a later PR if you prefer to get the feature in now. I'll quickly rebase.

wjakob:
I think that it would be great to have `Eigen::Ref<Eigen::SparseMatrix>` and `Eigen::Map<Eigen::SparseMatrix>`. You don't need to write documentation, I already started myself on writing some for Eigen/Dense and that can easily be adapted/extended.

Note that there was a change on `master` that impacts this PR: the `nb::tensor<..>` type was renamed to `nb::ndarray<..>`

hmenke:
I've tried to get `Map` and `Ref` support up and running but failed so far. For now I have added stubs for these implementation so that if anybody attempts to use this, the compiler errors are a little bit less useless.

wjakob:
Okay, fair enough. FYI, I added preliminary documentation for the dense part in the meantime: https://nanobind.readthedocs.io/en/latest/eigen.html — can you address the remaining conflicts so that the PR can be merged? (should just be a matter of moving the sparse-specific text into a subsection).

hmenke:
Done. I think it'd be good to run the testsuite again before merging just to be sure.

wjakob:
Awesome, all passes! I'll squash these into a single commit. Thank you for all of this hard work!

include/nanobind/eigen/sparse.h
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        ScalarNDArray data(src->valuePtr(), 1, data_shape, owner);
+        StorageIndexNDArray outer_indices(src->outerIndexPtr(), 1, outer_indices_shape, owner);
+        StorageIndexNDArray inner_indices(src->innerIndexPtr(), 1, data_shape, owner);
+
+        try {
+            return matrix_type(make_tuple(
+                                   std::move(data), std::move(inner_indices), std::move(outer_indices)),
+                               make_tuple(rows, cols))
```
hmenke, line 137: I've noticed that this sometimes create ADL problems, because `nanobind::make_tuple` competes with `std::make_tuple`. A simple fix would be to fully qualify `make_tuple`, but is there a better solution? @wjakob 
```suggestion
            return matrix_type(nanobind::make_tuple(
                                   std::move(data), std::move(inner_indices), std::move(outer_indices)),
                               make_tuple(rows, cols))
```

PR #127, opened 2023-02-15
nanobind-config.cmake: Avoid top-level warnings in nanobind headers

Lorac:
If you're using nanobind_add_module within a project that has many warnings enabled, the compiler may warn you that nanobind doesn't comply with the current warning settings.

To prevent the warnings set by the top-level project from affecting nanobind, we can add the nanobind include directory with the SYSTEM flag. This will mark the directory as a system directory, which typically means the compiler will not issue warnings for files included from this directory.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -139,7 +139,7 @@ function (nanobind_build_library TARGET_NAME)
   target_include_directories(${TARGET_NAME} PRIVATE
     ${NB_DIR}/ext/robin_map/include)
 
-  target_include_directories(${TARGET_NAME} PUBLIC
+  target_include_directories(${TARGET_NAME} SYSTEM PUBLIC
     ${Python_INCLUDE_DIRS}
     ${NB_DIR}/include)
 


wjakob:
Hi @Lorac,

I have never encountered such a problem before. It would be useful to have more details: what is the error message you get, on which compiler does this happen, etc.

Including nanobind as a `SYSTEM` include dependency is not an option for me, since its code should not issue warnings (if so, those parts of nanobind should be adapted, and unreasonable warnings could be silenced locally using suppressions).

It would be okay for the Python headers though. Can you adapt the PR, or does this then not address your problem anymore?

Wenzel

wjakob:
We already have a suppression for the size-0 array for MSVC, and it seems fine to add another one for GCC/Clang. The superfluous semicolon should likely be fixed.

wjakob:
> ..  but I'm thinking more of the uphill battle and supporting ..

Be that as it may, but adding nanobind as a `SYSTEM` header is simply not an option for me (whether conditionally or not).

Lorac:
Another way that I think could work would probably be to install headers inside a global include dir like: `/usr/include/nanobind/*.h`. That way, the headers are automatically systems headers. But that could be problematic with the files part of the src directory.

Anyhow, for the time being, I will support a fork on my side and try to find a better solution.

PR #138, opened 2023-03-04
Add filesystem support

qnzhou:
Summary:
* Add type caster for `std::filesystem::path` ~~or `std::experimental::filesystem::path`~~ to `pathlib.Path` based on [pybind11's counterpart](https://github.com/pybind/pybind11/blob/master/include/pybind11/stl/filesystem.h).
* Add unit test and update doc.

### DIFF
--- a/include/nanobind/stl/filesystem.h
+++ b/include/nanobind/stl/filesystem.h
@@ -20,32 +20,48 @@ template <>
 struct type_caster<std::filesystem::path> {
 private:
     static str unicode_from_fs_native(const std::string &w) {
-        PyObject* result = PyUnicode_DecodeFSDefaultAndSize(w.c_str(),
-                                                Py_ssize_t(w.size()));
+        PyObject *result =
+            PyUnicode_DecodeFSDefaultAndSize(w.c_str(), Py_ssize_t(w.size()));
         return steal<str>(result);
     }
 
     static str unicode_from_fs_native(const std::wstring &w) {
-        PyObject* result = PyUnicode_FromWideChar(w.c_str(), Py_ssize_t(w.size()));
+        PyObject *result =
+            PyUnicode_FromWideChar(w.c_str(), Py_ssize_t(w.size()));
         return steal<str>(result);
     }
 
 public:
     static handle from_cpp(const std::filesystem::path &path, rv_policy,
                            cleanup_list *) noexcept {
-        if (auto py_str = unicode_from_fs_native(path.native())) {
-            return module_::import_("pathlib")
-                .attr("Path")(py_str)
-                .release();
+        auto py_str = unicode_from_fs_native(path.native());
+        if (py_str.is_valid()) {
+            try {
+                return module_::import_("pathlib")
+                    .attr("Path")(py_str)
+                    .release();
+            } catch (python_error &e) {
+                e.restore();
+            }
+            return handle();
         }
-        return {};
+        return handle();
     }
 
     bool from_python(handle src, uint8_t, cleanup_list *) noexcept {
+        // PyUnicode_FSConverter and PyUnicode_FSDecoder normally take care of
+        // calling PyOS_FSPath themselves, but that's broken on PyPy (PyPy
+        // issue #3168) so we do it ourselves instead.
+        PyObject *buf = PyOS_FSPath(src.ptr());
+        if (!buf) {
+            PyErr_Clear();
+            return false;
+        }
+
         PyObject *native = nullptr;
         if constexpr (std::is_same_v<typename std::filesystem::path::value_type,
                                      char>) {
-            if (PyUnicode_FSConverter(src.ptr(), &native) != 0) {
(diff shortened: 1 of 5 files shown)


include/nanobind/stl/filesystem.h
```diff
@@ -0,0 +1,105 @@
...
+#endif
+#endif
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+template <typename T>
+struct path_caster {
+private:
+    static PyObject *unicode_from_fs_native(const std::string &w) {
```
wjakob, line 30: Can you update these functions so that they return a `str` object? You will have to wrap the result into `steal<str>(result)`. Then that detail can be removed from functions calling `unicode_from_fs_native` (it actually seems unsafe how these functions are being used right now with respect to exceptions)

tests/test_stl.py
```diff
@@ -728,3 +729,13 @@ def test65_class_with_movable_field(clean):
         move_constructed=2,
         destructed=4
     )
+
+def test66_replace_extension():
```
wjakob, line 733: If you grep through for `Path` in the pybind11 test suite, there seems to be more material that isn't included in this PR (`test_fs_path()` in `test_stl.py`)

include/nanobind/stl/filesystem.h
```diff
@@ -0,0 +1,105 @@
...
+        if (auto py_str = unicode_from_fs_native(path.native())) {
+            return module_::import_("pathlib")
+                .attr("Path")(steal(py_str))
+                .release();
+        }
+        return {};
+    }
+
+    bool from_python(handle src, uint8_t, cleanup_list *) noexcept {
+        // PyUnicode_FSConverter and PyUnicode_FSDecoder normally take care of
```
wjakob, line 55: The pypy issue mentioned here is ancient, it was created when the PyPy bugtracker was still on Bitbucket.  Can you remove the `PyOS_FSPath` line and actually check that the issue still exists? (We require very recent pypy version in nanobind)

include/nanobind/stl/filesystem.h
```diff
@@ -0,0 +1,105 @@
...
+        PyObject *native = nullptr;
+        if constexpr (std::is_same_v<typename T::value_type, char>) {
+            if (PyUnicode_FSConverter(buf, &native) != 0) {
+                if (auto *c_str = PyBytes_AsString(native)) {
+                    // AsString returns a pointer to the internal buffer, which
+                    // must not be free'd.
+                    value = c_str;
+                }
+            }
+        } else if constexpr (std::is_same_v<typename T::value_type, wchar_t>) {
```
wjakob, line 72: Just `} else {` will do here.

wjakob (changes requested):
Thanks for the PR, here are a few minor things that could still be improved.

include/nanobind/stl/filesystem.h
qnzhou: I have pushed a change.  Let me know if I understand your request correctly on this one.

include/nanobind/stl/filesystem.h
qnzhou: That issue seems to be still open: https://foss.heptapod.net/pypy/pypy/-/issues/3168

I am not sure if this is related to the CI failure with pypy. Any suggestions?

include/nanobind/stl/filesystem.h
```diff
@@ -0,0 +1,77 @@
...
+
+    static str unicode_from_fs_native(const std::wstring &w) {
+        PyObject* result = PyUnicode_FromWideChar(w.c_str(), Py_ssize_t(w.size()));
+        return steal<str>(result);
+    }
+
+public:
+    static handle from_cpp(const std::filesystem::path &path, rv_policy,
+                           cleanup_list *) noexcept {
+        if (auto py_str = unicode_from_fs_native(path.native())) {
```
wjakob, line 36: I would write: 

```str py_str = unicode_from_fs_native(path.native());
if (py_str.is_valid()) { ...}
```

include/nanobind/stl/filesystem.h
wjakob: Yes, I think you are right. That change caused the build error that now appears on PyPy, so it still does seem to be an issue. (That issue was so stale that I assumed it might have been fixed in the meantime without updating the issue tracker ticket)

include/nanobind/stl/filesystem.h
```diff
@@ -0,0 +1,77 @@
...
+            return module_::import_("pathlib")
```
wjakob, line 37: The code here is a little naïve -- the `Path()` call might, e.g., raise an exception that would crash the process. Let's write
```cpp
try {
    return module_::import_(..)...
} catch (const python_error &e) {
    e.restore();
}
return handle();
````

wjakob:
Merged, thanks!

PR #149, opened 2023-03-23
Add <exception> to headers included by nanobind.

hawkinsp:
nanobind uses types from <exception> without including the header, e.g., std::exception. This will break in future versions of libc++, where not all such types are available via transitive includes.

### DIFF
--- a/include/nanobind/nanobind.h
+++ b/include/nanobind/nanobind.h
@@ -27,6 +27,7 @@
 
 // Core C++ headers that nanobind depends on
 #include <cstdint>
+#include <exception>
 #include <stdexcept>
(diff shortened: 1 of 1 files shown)


wjakob:
My impression from is that this PR was prompted by  automated analysis tool, of which I am not a big fan. (churn and generally unnecessary work to silence warnings that don't represent real problems). Is this assumption correct?

In this particular case, I am also wondering: How could `stdexcept` not include `exception`? (It defines subclasses, which requires that the base class is defined).

wjakob:
That's good to know, I will merge it then. Thank you for the PR.

PR #161, opened 2023-03-30
NumPy: adding `size`, `shape_ptr`, `stride_ptr`

tdegeus:
Fixes https://github.com/wjakob/nanobind/discussions/158

I was not sure how you deal with tests. I could not find tests for e.g. `stride` to I tried to be non-intrusive. But let me know if you want otherwise.

### DIFF
--- a/include/nanobind/ndarray.h
+++ b/include/nanobind/ndarray.h
@@ -296,10 +296,20 @@ public:
     size_t ndim() const { return m_dltensor.ndim; }
     size_t shape(size_t i) const { return m_dltensor.shape[i]; }
     int64_t stride(size_t i) const { return m_dltensor.strides[i]; }
+    int64_t* shape_ptr() const { return &m_dltensor.shape[0]; }
+    int64_t* stride_ptr() const { return &m_dltensor.strides[0]; }
     bool is_valid() const { return m_handle != nullptr; }
     int32_t device_type() const { return m_dltensor.device.device_type; }
     int32_t device_id() const { return m_dltensor.device.device_id; }
     detail::ndarray_handle *handle() const { return m_handle; }
+
+    size_t size() const {
+        size_t ret = 1;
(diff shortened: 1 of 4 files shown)


include/nanobind/ndarray.h
```diff
@@ -293,9 +293,12 @@ template <typename... Args> class ndarray {
     }
 
     dlpack::dtype dtype() const { return m_dltensor.dtype; }
+    size_t size() const { return Info::size; }
```
wjakob, line 296: Isn't `size()` doing the same as `ndim()` here? I would have expected `size()` to return the total number of tensor coefficients.

include/nanobind/ndarray.h
```diff
@@ -293,9 +293,12 @@ template <typename... Args> class ndarray {
...
     size_t ndim() const { return m_dltensor.ndim; }
     size_t shape(size_t i) const { return m_dltensor.shape[i]; }
     int64_t stride(size_t i) const { return m_dltensor.strides[i]; }
+    auto* shape_ptr() const { return &m_dltensor.shape[0]; }
```
wjakob, line 300: Let's not use `auto` in a public interface especially when the type is easily knowable. You can return the array directly and don't need `&array[0]`.

wjakob (changes requested):
A quick review. Two further requests:

- Please at least reference those functions somewhere from a C++ test case.
- They will still need to be mentioned in the API documentation `docs/api_extra.rst`

include/nanobind/ndarray.h
tdegeus: Done. I'm just a bit confused about the type of `shape_ptr`. From the other return types I would have expected it to be `size_t*`. However, my compiler is convinced about it being `int64_t*`

wjakob:
The changed code looks good, but please note the request above regarding documentation + test coverage.

wjakob:
Thanks!

PR #162, opened 2023-03-31
Fix for ndarray_import when tensor imported is batch size 1.

galv:
Sorry for making a pullrequest instead of an issue, but it looks like pytorch does not always set the stride of its outermost dimension equal to 0 or the "expected stride" (i.e., the stride you would get if you increased the singleton dimension to a length greater than 1) when that dimension is singleton (i.e., that dimension's length is 1). Instead, it uses a stride of 1. `ndarray_import` assumes that singleton dimensions either have the "expected stride" or 0. If this isn't the case, you get an error about "incompatible function arguments".

I wrote a few reproducers for this issue, as well as a basic fix, though I admit it's a bit "ugly". However, since the only index that will ever be valid for a dimension of size 1 is 0, its stride will never be used anyway.

Edit: Upon further reflection, I don't believe this quick fix is entirely correct. What we want is to allow the slowest-changing dimension (So the first one in "C" ordering, or the last one in "F" ordering) to have any stride value when its length is equal to 1.

### DIFF
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -309,6 +309,10 @@ ndarray_handle *ndarray_import(PyObject *o, const ndarray_req *req,
         }
     }
 
+    int64_t size = 1;
+    for (uint32_t i = 0; i < req->ndim; ++i)
+        size *= t.shape[i];
+
     scoped_pymalloc<int64_t> strides(t.ndim);
     if ((req->req_order || !t.strides) && t.ndim > 0) {
         size_t accum = 1;
@@ -330,14 +334,14 @@ ndarray_handle *ndarray_import(PyObject *o, const ndarray_req *req,
             pass_order = false;
         }
 
-        if (req->req_order) {
+        if (req->req_order && size != 0) { // Tolerate any strides if empty
             if (!t.strides) {
-                // c-style strides assumed
+                /* The provided tensor does not have a valid strides
+                   field, which implies a C-style ordering. */
                 pass_order = req->req_order == 'C';
             } else {
                 for (uint32_t i = 0; i < (uint32_t) t.ndim; ++i) {
-                    if (!((strides[i] == t.strides[i]) ||
-                          (t.shape[i] == 1 && t.strides[i] == 0))) {
+                    if (t.shape[i] != 1 && strides[i] != t.strides[i]) {
                         pass_order = false;
                         break;
                     }
--- a/tests/test_ndarray.cpp
+++ b/tests/test_ndarray.cpp
@@ -141,4 +141,13 @@ NB_MODULE(test_ndarray_ext, m) {
 
             return nb::ndarray<nb::numpy, float>(f, 0, shape, deleter);
     });
+
+    m.def(
+        "noop_3d_c_contig",
+        [](nb::ndarray<float, nb::shape<nb::any, nb::any, nb::any>, nb::c_contig>) { return; });
+
+    m.def(
+        "noop_2d_f_contig",
+        [](nb::ndarray<float, nb::shape<nb::any, nb::any>, nb::f_contig>) { return; });
+
 }
--- a/tests/test_ndarray.py
+++ b/tests/test_ndarray.py
@@ -358,3 +358,52 @@ def test18_return_array_scalar():
     del x
     collect()
     assert t.destruct_count() - dc == 1
+
+# See PR #162
+@needs_torch
+def test19_single_and_empty_dimension_pytorch():
+    a = torch.ones((1,100,1025), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+    a = torch.ones((100,1,1025), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+    a = torch.ones((0,100,1025), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+    a = torch.ones((100,0,1025), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+    a = torch.ones((100,1025,0), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+    a = torch.ones((100,0,0), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+    a = torch.ones((0,0,0), dtype=torch.float32)
+    t.noop_3d_c_contig(a)
+
+# See PR #162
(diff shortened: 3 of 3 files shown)


wjakob:
The fix looks okay to me. Basically we don't care about strides for singleton or empty dimensions. While you are at it, could you change `(t.shape[i] == 1)` to  `t.shape[i] <= 1` (less than or equals, and removal of superfluous parenthesis.

Also, please rename the tests ("galvez") -- having your username in the will confuse people since it is not clear what it means. You could call them "pytorch_dim_1" and "pytorch_dim_2". Also add a comment (`see PR #162`).

galv:
> The fix looks okay to me. Basically we don't care about strides for singleton or empty dimensions. While you are at it, could you change (t.shape[i] == 1) to t.shape[i] <= 1 (less than or equals, and removal of superfluous parenthesis.

You are right. I also further tested to make sure that the t.shape[i] == 0 case works for both numpy and pytorch (Notably, numpy sets all strides equal to 0 if t.shape[i] == 0 for any i).

> Also, please rename the tests ("galvez") -- having your username in the will confuse people since it is not clear what it means. You could call them "pytorch_dim_1" and "pytorch_dim_2". Also add a comment (see PR #162).

Yes, absolutely. Apologies for the initial sloppiness. I submitted the PR yesterday right before I had to head out for a previous commitment.

galv:
I force pushed my updates. I removed an extra superfluous set of parentheses around `strides[i] == t.strides[i]` while doing this. I suppose I am not sure if you like those sorts of style changes being part of bug fix commits.

Let me know if you want anything else from me here. Thank you for your help.

galv:
Let's hold off on merging this for a bit. I noticed some strange behaviors around shape[i] = 0 where i != 0 for pytorch. I will make a new commit explaining the issue and proposing a fix.

src/nb_ndarray.cpp
```diff
@@ -335,12 +335,37 @@ ndarray_handle *ndarray_import(PyObject *o, const ndarray_req *req,
                 // c-style strides assumed
                 pass_order = req->req_order == 'C';
             } else {
-                for (uint32_t i = 0; i < (uint32_t) t.ndim; ++i) {
-                    if (!((strides[i] == t.strides[i]) ||
-                          (t.shape[i] == 1 && t.strides[i] == 0))) {
+                bool seen_previous_shape_eq_0 = false;
+                uint32_t i;
```
galv, line 339: This code is quite ugly. One nice thing would be to use a C++ iterator instead of lower-level indices. Just initialize the iterator to begin() if req_order = 'C', or rbegin() if req_order = 'F'. Also initialize end_it to either end() or rend().

Then we could drop the craziness going on at the bottom of the loop:

```
                    if ((req->req_order == 'C' && i == 0) ||
                        (req->req_order == 'F' && i == (uint32_t) (t.ndim - 1)))
                        break;
                    if (req->req_order == 'C') {
                        --i;
                    } else if (req->req_order == 'F') {
                        ++i;
                    }
```

and replace it with:

```
if (it == end_it) {
    break;
}
++it;
```

galv:
Please read the commit message here: https://github.com/wjakob/nanobind/pull/162/commits/814aa4c191bde6bcfd688919844d23aa2e0b7901

I realized that the original code change as you suggested did not handle cases when shape[i] = 0 when i was not the slowest-changing dimension. Here is one suggested way of handling that appropriately, handling column-major order correctly. Note again that this rigamarole is basically due to pytorch's behavior regarding strides when an element of shape is equal to 0.. Numpy is perfectly fine as-is.

galv:
Upon further reflection, I don't think https://github.com/wjakob/nanobind/commit/814aa4c191bde6bcfd688919844d23aa2e0b7901 is the "right" way to do things. I basically just need to encode the rule that if `t.shape[i] == 0` for any value of i, then we should accept any values for t.strides, since the tensor has 0 elements anyway. That would be much more concise than the current commit. I can probably get around to this by Monday (but wouldn't be upset if you went ahead and implemented that yourself).

wjakob:
>  I basically just need to encode the rule that if t.shape[i] == 0 for any value of i, then we should accept any values for t.strides, since the tensor has 0 elements anyway.

That sounds good to me. I will wait for your updated PR (have a few other things on my plate)

wjakob:
FYI, The failing test is a flake related to `capfd` in `pytest`, and not related to nanobind, I reported it separately.

galv:
Alright, I made my change. I think this is good to go (assuming CI passes).

src/nb_ndarray.cpp
```diff
@@ -335,9 +335,19 @@ ndarray_handle *ndarray_import(PyObject *o, const ndarray_req *req,
                 // c-style strides assumed
                 pass_order = req->req_order == 'C';
             } else {
+                bool size_is_zero = std::any_of(t.shape, t.shape + t.ndim,
```
wjakob, line 338: Even if the CI passes, this cannot be merged as is because `std::any_of` is part of the `algorithm` header that we (intentionally) do not import. You will need to write this test manually using plain C/C++.

src/nb_ndarray.cpp
```diff
@@ -335,9 +335,19 @@ ndarray_handle *ndarray_import(PyObject *o, const ndarray_req *req,
...
+                    if (// regular case: strides should match
+                        !(strides[i] == t.strides[i] ||
+                          // if the shape of this dimension is 1, any
+                          // value of t.strides[i] is acceptable,
+                          // since 0 is the only valid index
+                          t.shape[i] <= 1 ||
+                          // if the shape of this dimension or any
+                          // faster-changing dimension is 0, any value of
+                          // t.strides[i] is acceptable
+                          size_is_zero)) {
```
wjakob, line 350: Rather than including this as part of the loop condition, I would make the entire loop conditional on `!size_is_zero`. I am not a fan of comments within expressions, could you move them to the top? 

Also, note that your comment states ` if the shape of this dimension is 1,` but the actual test is `<= 1` which includes zero. Since that case is already covered by the `size_is_zero` condition, you could change it to `== 1` to remove potential confusion caused by the overlap.

wjakob (changes requested):
A few more comments

wjakob:
I realized the changes myself (want to push out 1.0.1 soon).

galv:
No problem. Thank you for the help reviewing.

PR #165, opened 2023-03-31
Pedantic fix : '... is is ...' -> '... is ...'

aidaco:
The word 'is' was repeated in "Why another binding library?".


wjakob:
Thanks!

PR #168, opened 2023-04-01
Wrap `PyInit_##name` in macro to allow macros as module names

tttapa:
This is similar to what pybind11 does:

https://github.com/pybind/pybind11/blob/4ce05175d51a4685232452bdc1e9cbb13a240a65/include/pybind11/detail/common.h#LL372-L374

Pasting the macro argument directly does not work, as demonstrated by: https://godbolt.org/z/Kv9EEfEsY


wjakob:
LGTM, thanks!

PR #170, opened 2023-04-02
Fix func_extra_apply for rv_policy

arsarabi:
Add missing 1 since `rv_policy` spans 3 bits. One effect of this bug is that supplying a return value policy to properties does not work properly.


wjakob:
Ugh, that's bad. Thanks for catching this!

PR #174, opened 2023-04-05
nb::typed<>: support use in return position

oremanj:
The uses are analogous to those in argument position; for example, you might want to return `nb::iterator` but indicate in the signature that it's an iterator over some specific type. It almost worked already, just had a typo in the `type_caster`.


wjakob:
Great catch, but what is the documentation change about?

oremanj:
The doc change is just "function argument" -> "function argument or return value".

wjakob:
Got it. (It looked identical after a superficial check). Thanks!

PR #175, opened 2023-04-05
Add <nanobind/stl/chrono.h> caster, ported from pybind11

oremanj:
The datetime object interactions are now encapsulated in inline functions outside the type caster, in order to be able to swap them out for Python-level operations in limited API builds.


oremanj:
Diff from pybind11:
```
--- /Users/oremanj/dev/atl/src/3ps/pybind11/include/pybind11/chrono.h	2022-03-17 15:03:30
+++ include/nanobind/stl/chrono.h	2023-04-05 13:58:13
@@ -1,5 +1,5 @@
 /*
-    pybind11/chrono.h: Transparent conversion between std::chrono and python's datetime
+    nanobind/stl/chrono.h: Transparent conversion between std::chrono and python's datetime

     Copyright (c) 2016 Trent Houliston <trent@houliston.me> and
                        Wenzel Jakob <wenzel.jakob@epfl.ch>
@@ -7,32 +7,23 @@
     All rights reserved. Use of this source code is governed by a
     BSD-style license that can be found in the LICENSE file.
 */
-
 #pragma once

-#include "pybind11.h"

+#define __STDC_WANT_LIB_EXT1__ 1 // for localtime_s
+#include <time.h>
+
 #include <chrono>
 #include <cmath>
 #include <ctime>
-#include <mutex>
+#include <type_traits>

 #include <datetime.h>

-// Backport the PyDateTime_DELTA functions from Python3.3 if required
-#ifndef PyDateTime_DELTA_GET_DAYS
-#define PyDateTime_DELTA_GET_DAYS(o)         (((PyDateTime_Delta*)o)->days)
-#endif
-#ifndef PyDateTime_DELTA_GET_SECONDS
-#define PyDateTime_DELTA_GET_SECONDS(o)      (((PyDateTime_Delta*)o)->seconds)
-#endif
-#ifndef PyDateTime_DELTA_GET_MICROSECONDS
-#define PyDateTime_DELTA_GET_MICROSECONDS(o) (((PyDateTime_Delta*)o)->microseconds)
-#endif
+NAMESPACE_BEGIN(detail)

-PYBIND11_NAMESPACE_BEGIN(PYBIND11_NAMESPACE)
-PYBIND11_NAMESPACE_BEGIN(detail)
-
 template <typename type> class duration_caster {
 public:
     using rep = typename type::rep;
@@ -40,7 +31,7 @@

     using days = std::chrono::duration<int_least32_t, std::ratio<86400>>; // signed 25 bits required by the standard.

-    bool load(handle src, bool) {
+    bool from_python(handle src, uint8_t /*flags*/, cleanup_list*) noexcept {
         using namespace std::chrono;

         // Lazy initialise the PyDateTime import
@@ -73,7 +64,7 @@
         return src.time_since_epoch();
     }

-    static handle cast(const type &src, return_value_policy /* policy */, handle /* parent */) {
+    static handle from_cpp(const type &src, rv_policy, cleanup_list*) noexcept {
         using namespace std::chrono;

         // Use overloaded function to get our duration from our source
@@ -95,30 +86,42 @@
         return PyDelta_FromDSU(dd.count(), ss.count(), us.count());
     }

-    PYBIND11_TYPE_CASTER(type, const_name("datetime.timedelta"));
+    NB_TYPE_CASTER(type, const_name("datetime.timedelta"));
 };

+template <class... Args>
+auto can_localtime_s(Args*... args) ->
+    decltype((localtime_s(args...), std::true_type{}));
+std::false_type can_localtime_s(...);
+
+template <class... Args>
+auto can_localtime_r(Args*... args) ->
+    decltype((localtime_r(args...), std::true_type{}));
+std::false_type can_localtime_r(...);
+
 inline std::tm *localtime_thread_safe(const std::time_t *time, std::tm *buf) {
-#if (defined(__STDC_LIB_EXT1__) && defined(__STDC_WANT_LIB_EXT1__)) || defined(_MSC_VER)
-    if (localtime_s(buf, time))
-        return nullptr;
-    return buf;
-#else
-    static std::mutex mtx;
-    std::lock_guard<std::mutex> lock(mtx);
-    std::tm *tm_ptr = std::localtime(time);
-    if (tm_ptr != nullptr) {
-        *buf = *tm_ptr;
+    if constexpr (decltype(can_localtime_s(time, buf))::value) {
+        // C11 localtime_s
+        std::tm* ret = localtime_s(time, buf);
+        return ret;
+    } else if constexpr (decltype(can_localtime_s(buf, time))::value) {
+        // Microsoft localtime_s (with parameters switched and errno_t return)
+        int ret = localtime_s(buf, time);
+        return ret == 0 ? buf : nullptr;
+    } else {
+        static_assert(decltype(can_localtime_r(time, buf))::value,
+                      "<nanobind/stl/chrono.h> type caster requires "
+                      "that your C library support localtime_r or localtime_s");
+        std::tm* ret = localtime_r(time, buf);
+        return ret;
     }
-    return tm_ptr;
-#endif
 }

 // This is for casting times on the system clock into datetime.datetime instances
 template <typename Duration> class type_caster<std::chrono::time_point<std::chrono::system_clock, Duration>> {
 public:
     using type = std::chrono::time_point<std::chrono::system_clock, Duration>;
-    bool load(handle src, bool) {
+    bool from_python(handle src, uint8_t /*flags*/, cleanup_list*) noexcept {
         using namespace std::chrono;

         // Lazy initialise the PyDateTime import
@@ -163,7 +166,7 @@
         return true;
     }

-    static handle cast(const std::chrono::time_point<std::chrono::system_clock, Duration> &src, return_value_policy /* policy */, handle /* parent */) {
+    static handle from_cpp(const std::chrono::time_point<std::chrono::system_clock, Duration> &src, rv_policy, cleanup_list*) noexcept {
         using namespace std::chrono;

         // Lazy initialise the PyDateTime import
@@ -182,9 +185,11 @@
         std::time_t tt = system_clock::to_time_t(time_point_cast<system_clock::duration>(src - us));

         std::tm localtime;
-        std::tm *localtime_ptr = localtime_thread_safe(&tt, &localtime);
-        if (!localtime_ptr)
-            throw cast_error("Unable to represent system_clock in local time");
+        if (!localtime_thread_safe(&tt, &localtime)) {
+            PyErr_SetString(PyExc_ValueError,
+                            "Unable to represent system_clock in local time");
+            return handle();
+        }
         return PyDateTime_FromDateAndTime(localtime.tm_year + 1900,
                                           localtime.tm_mon + 1,
                                           localtime.tm_mday,
@@ -193,7 +198,7 @@
                                           localtime.tm_sec,
                                           us.count());
     }
-    PYBIND11_TYPE_CASTER(type, const_name("datetime.datetime"));
+    NB_TYPE_CASTER(type, const_name("datetime.datetime"));
 };

 // Other clocks that are not the system clock are not measured as datetime.datetime objects
@@ -207,5 +212,5 @@
 : public duration_caster<std::chrono::duration<Rep, Period>> {
 };

-PYBIND11_NAMESPACE_END(detail)
-PYBIND11_NAMESPACE_END(PYBIND11_NAMESPACE)
+NAMESPACE_END(detail)
+NAMESPACE_END(NB_NAMESPACE)
```

wjakob:
This might involve some extra complications. Nanobind extensions should also compile in limited API/stable ABI mode, and this type caster uses functions that aren't part of that interface.

oremanj:
Yep, working on that now. I didn't have a 3.12 build on my machine to test with.

oremanj:
OK, I think this is now ready for review. Let me know if you'd prefer that the datetime types be held in internals when using the limited API; it didn't seem worth it to me.

wjakob:
Since datetime is such a specific feature (you are the first to request it), I would suggest to keep it out of the library (Otherwise, this will have a binary size impact on everyone).

You can simply turn those static library functions into inline functions in the corresponding type caster header. As long as they aren't templated, they will still just be compiled once and the overhead of this in your case should not be too bad.

oremanj:
The `size` command says `nb_datetime.cpp.o` is 927 bytes of text and 8 of data on my machine. The rest of `libnanobind-static.a` is 108263 bytes of text and 2840 bytes of data. I feel like a less than 1% increase in binary size is appropriate here, and is 10x smaller than `nb_ndarray.cpp.o` which also supports datatypes that are irrelevant to a large class of applications (though certainly quite relevant to many as well). When building nanobind as a static library, if datetimes are truly unused then they won't be included in the final extension module at all (since they're in their own compilation unit).

> As long as they aren't templated, they will still just be compiled once and the overhead of this in your case should not be too bad.

It's creating unnecessary work for the compiler and the linker on a per-compilation-unit basis, which I thought was something nanobind was trying to avoid. :-) I'm happy to make the change if you still feel strongly about it, but it didn't seem in keeping with nanobind's deliberate movement away from pybind's "throw it in the header and let the linker sort it out" model.

oremanj:
Another unfortunate outcome of making this header-only is that Python's `<datetime.h>` defines `PyDateTimeAPI` as a static global pointer, so every compilation unit gets its own copy. We can avoid this on the limited API by using an inline variable, but I don't see how to avoid data bloat on the regular API. 8 bytes per compilation unit isn't that bad all things considered, but it seems nice to avoid it if we can.

wjakob:
@oremanj -- I sympathize but am unfortunately not flexible on this point.

The library component should be as small as possible. Unfortunately, it's already much larger than I would like it to be :-(. N-D arrays are really special in this discussion because they are one of the main use cases of nanobind -- to process lots of data that is usually wrapped up in CPU/GPU/.. tensor types. The <1% argument is somewhat problematic because we've had a lot of these discussions in pybind11, and look how much bloat has accumulated over time..

If you care strongly about this point, there is nothing keeping you from shipping a custom typecaster in your own code that links against a static library component. I will merge the PR if you are willing to make the change to a header-only typecaster.

oremanj:
@wjakob fair enough, it's your library :-) I've updated the caster to be header-only.

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+    // Types defined by the datetime module
+    handle datetime;
+    handle time;
+    handle date;
+    handle timedelta;
+
+    explicit operator bool() const { return datetime.ptr() != nullptr; }
+
+    bool do_import() {
+        auto mod = module_::import_("datetime");
```
wjakob, line 45: `auto` -> `object`

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+// do it the hard way.
+
+struct datetime_types_t {
+    // Types defined by the datetime module
+    handle datetime;
+    handle time;
+    handle date;
+    handle timedelta;
+
+    explicit operator bool() const { return datetime.ptr() != nullptr; }
```
wjakob, line 42: `datetime.is_valid()`

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+// (since they might be in multiple translation units and the linker
+// should pick one) but NB_NOINLINE because we don't want the bloat
+// of actually inlining them. They are defined in this header instead
+// of in the built nanobind library in order to avoid increasing the
+// library size for users who don't care about datetimes.
+
+NB_NOINLINE inline bool set_from_int_attr(int *dest, PyObject *o,
+                                          const char *name) {
+    PyObject *value = PyObject_GetAttrString(o, name);
+    if (!value) return false;
```
wjakob, line 82: return statement on new line

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+    if (PyDelta_Check(o)) {
+        *days = PyDateTime_DELTA_GET_DAYS(o);
+        *secs = PyDateTime_DELTA_GET_SECONDS(o);
+        *usecs = PyDateTime_DELTA_GET_MICROSECONDS(o);
+        return true;
+    }
+#else
+    if (!datetime_types && !datetime_types.do_import()) {
```
wjakob, line 121: ditto (braces), won't mention later occurrences

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+    if (int is_td = PyObject_IsInstance(o, datetime_types.timedelta.ptr());
+        is_td < 0) {
+        return false;
+    } else if (is_td > 0) {
+        return (set_from_int_attr(days, o, "days") &&
+                set_from_int_attr(secs, o, "seconds") &&
+                set_from_int_attr(usecs, o, "microseconds"));
+    }
+#endif
+    return false;
```
wjakob, line 133: One general observation about this function is that some failure return values have a Python error state set, and others don't. Not sure if that's a problem.

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+#include <ctime>
+#include <limits>
+#include <type_traits>
+
+#ifndef Py_LIMITED_API
+#include <datetime.h>
+#endif
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
```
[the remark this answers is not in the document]
wjakob: Alternatively, other helper functionality like `unpack_timedelta` and `unpack_datetime` could move there as well.

include/nanobind/stl/chrono.h
wjakob: Probably this could just be merged into `ensure_ready()` function (see above)

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+#endif
+        if (is_float) {
+            value = type(duration_cast<duration<rep, period>>(duration<double>(PyFloat_AsDouble(src.ptr()))));
+            return true;
+        }
+        return false;
+    }
+
+    // If this is a duration just return it back
+    static const std::chrono::duration<rep, period>& get_duration(const std::chrono::duration<rep, period> &src) {
```
wjakob, line 308: Some `using` declarations for long type names might be helpful here and below.

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+
+template <class Time, class Buf>
+inline std::tm *localtime_thread_safe(const Time *time, Buf *buf) {
+    if constexpr (decltype(can_localtime_s(time, buf))::value) {
+        // C11 localtime_s
+        std::tm* ret = localtime_s(time, buf);
+        return ret;
+    } else if constexpr (decltype(can_localtime_s(buf, time))::value) {
+        // Microsoft localtime_s (with parameters switched and errno_t return)
+        int ret = localtime_s(buf, time);
```
wjakob, line 357: Ugh, what a mess. I did not realize that there were platform differences in such a core API.

wjakob (changes requested):
Thank you for the PR. Here is an initial set of comments from me. It will need at least one more iteration before changes can be merged.

What is still missing in this PR:
- Better test coverage of this surprisingly complicated type caster. (I daresay it's the most complicated one?). It would be good to retain more of the ones from pybind11: https://github.com/pybind/pybind11/blob/master/tests/test_chrono.cpp
- Documentation coverage, much can probably be lifted from pybind11 as well (https://pybind11.readthedocs.io/en/stable/advanced/cast/chrono.html)
- A changelog entry referencing this PR

include/nanobind/stl/chrono.h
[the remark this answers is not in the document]
oremanj: I moved all the non-std::chrono-dependent functions.

include/nanobind/stl/chrono.h
oremanj: This is intentional and is mentioned in the comments; I wanted to distinguish between "not the right type" errors (which are common during overload resolution and shouldn't produce a warning) and "something unexpected occurred" errors (which should be user-visible to reduce confusion if possible). However I noticed I was returning from `type_caster::from_python()` with an error set, which I think is unkosher; I've changed that to use `PyErr_WriteUnraisable`.

include/nanobind/stl/chrono.h
```diff
@@ -0,0 +1,439 @@
...
+    handle time;
+    handle date;
+    handle timedelta;
+
+    explicit operator bool() const { return datetime.ptr() != nullptr; }
+
+    bool do_import() {
+        auto mod = module_::import_("datetime");
+        object datetime_o, time_o, date_o, timedelta_o;
+        try {
```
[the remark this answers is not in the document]
oremanj: I changed the unpack functions to raise an exception, which makes the false-with-error vs false-without-error return path distinction less confusing IMO. Pack still uses the Python error indicator so that 'return pack_thing(...);' at the end of a type caster works nicely.

include/nanobind/stl/chrono.h
oremanj: Yep; the pybind caster tried one form of localtime_s and then fell back to localtime + std::mutex if that wasn't available.

oremanj:
It doesn't look like the compiler setup used in the Ubuntu CI runs likes the newly added `-Oz`.

wjakob:
Yeah, I removed that commit. Please don't feel the need to keep rebasing your PR to the latest master, I did some force-pushes that are probably causing pain on your end.

oremanj:
All right, I did one more rebase to hopefully get a clean CI, and will stop worrying about it after that. :-)

docs/conf.py
```diff
@@ -15,9 +15,14 @@
 extensions = [
     "sphinx_copybutton",
     "sphinxcontrib.rsvgconverter",
-    "sphinxcontrib.moderncmakedomain"
+    "sphinxcontrib.moderncmakedomain",
+    "sphinx.ext.intersphinx",
```
wjakob, line 19: neat!

wjakob:
I took a look over this PR, it looks great now. One last request: can you introduce something akin to `discard_as_unraisable` in the `python_error` class? I will merge the PR after this.

wjakob:
Hi @oremanj,

nanobind now supports compilation for 32 bit platforms as well. On WIN32, the tests for this type caster fail with

```
=============================================== short test summary info ===============================================
FAILED test_chrono.py::test_chrono_system_clock_roundtrip_time[None-time11] - ValueError: Unable to represent system_clock in local time; got time_t -1
FAILED test_chrono.py::test_chrono_system_clock_roundtrip_time[None-time12] - ValueError: Unable to represent system_clock in local time; got time_t -1
FAILED test_chrono.py::test_chrono_system_clock_roundtrip_time[None-time13] - ValueError: Unable to represent system_clock in local time; got time_t -1
FAILED test_chrono.py::test_chrono_system_clock_roundtrip_time[None-time14] - ValueError: Unable to represent system_clock in local time; got time_t -1
====================================== 4 failed, 224 passed, 59 skipped in 0.96s ======================================
```

Any ideas? The failing line is

```python
# Roundtrip the time
datetime = m.test_chrono2(time1)
```

oremanj:
I don't have a win32 machine handy and probably won't be able to do CI-based debugging of this in the next week. godbolt.org doesn't seem to have a win32 VM either; they'll let you compile stuff with MSVC for x86 but not run it. I think I might understand what's going on though.

The tests in question are the ones that try to roundtrip some times that are less than 1 hour after Jan 1 1970 in the local timezone (because combining with 1970-01-01 is how pybind treats a date-less time and I kept that behavior because I didn't have any better ideas):
```
        datetime.time(0, 0, 0),
        datetime.time(0, 0, 0, 1),
        datetime.time(0, 28, 45, 109827),
        datetime.time(0, 59, 59, 999999),
```
But it appears that Microsoft does not acknowledge the existence of times before the epoch?! From https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/mktime-mktime32-mktime64?view=msvc-170 :
> After an adjustment to UTC, _mktime32 handles dates from midnight, January 1, 1970, to 23:59:59 January 18, 2038, UTC. _mktime64 handles dates from midnight, January 1, 1970 to 23:59:59, December 31, 3000. This adjustment may cause these functions to return -1 (cast to time_t, __time32_t or __time64_t) even though the date you specify is within range. For example, if you are in Cairo, Egypt, which is two hours ahead of UTC, two hours will first be subtracted from the date you specify in timeptr; the subtraction may now put your date out of range.

Since the test you quoted involves roundtripping datetimes that are shortly after midnight _local time_ on Jan 1 1970, its success on Windows appears to be timezone-dependent; if you run it on a system where local time is set west of UTC, it will pass, while east of UTC it will fail. Did you perchance observe these failures on a system that is one hour east of UTC?

AFAIK all Unixes allow time_t to take on negative values to represent times before 1970. I'm honestly shocked that Windows doesn't.

I'm not sure what to do here except skip the test on Windows. We could disable conversion of a bare datetime.time to time_point entirely and remove the test; it's technically a breaking change but I doubt anyone is using it (I only included it because pybind did, and the logic is unchanged afaict). We could change the logic so bare times convert to that many hours/minutes/seconds after the epoch, i.e. interpret a bare time as being on Jan 1 1970 UTC even though full datetimes are interpreted in local time. Neither of these is especially appealing to me; what do you think?

wjakob:
Hi @oremanj — I tested this from a machine in Switzerland, which is at CET (currently GMT+2) — Does that invalidate the theory. By the way, this was actually on a 64 bit windows machine. It's easy to cross-compile legacy win32 binaries from such equipment with a flag (can be set in the interactive frontend `cmake-gui` when choosing the compiler to use for the build).

oremanj:
No, that supports the theory -- CET was GMT+1 on Jan 1 1970 (because summer time isn't active in the winter). So now we at least know what's going on; do you have any opinions on the best way to solve it?

By "win32" I meant the whole Windows platform; I don't have a Windows machine that I can easily interact with regardless of architecture.

PR #192, opened 2023-04-25
Refactor type_data to save space and reduce nb_type/nb_enum entanglement

oremanj:
* Split fields that are only needed during type object initialization (doc, base, base_py, type_slots) into a new `type_data_prelim` structure, so that the persistent `type_data` doesn't need to store them.
* Allow `nb::supplement<T*>` to store the `T*` as `type_data::supplement` directly, rather than requiring another layer of indirection to access the pointer. This is useful when `T` does not meet the requirements of a supplement type on its own (e.g., has a non-trivial destructor).
* Remove all enum-specific logic from `type_data` and `nb_type.cpp`; instead, `nb::enum_` now implements all of its additional functionality through the same `type_slots` interface available to user code. This serves as a demonstration of how nanobind users can implement their own enum-like semantics, and frees up three flag bits for other uses.

(I made these changes together in order to put all the `type_data` ABI effects in one place, but they're separable if you like some of them and not others.)

Enums no longer explicitly save their signedness; instead they fill in the `Py_nb_int` and `Py_nb_index` with a different function depending on the signedness known at type creation time. This makes `int()` very marginally faster (no need to branch on the signedness) and `repr()` / `__name__` very marginally slower (need to indirect through the type slot instead of calling `nb_enum_int` directly).

On my laptop, `python3.9 -m timeit -s "from test_enum_ext import Enum"` with the following microbenchmarks had these results:
* `"int(Enum.C)"` went from (52.7, 54) ns per loop (two separate runs) to (50.5, 50.6) ns per loop
* `"repr(Enum.C)"` went from (353, 372) ns per loop to (347, 353) ns per loop
* `"Enum.C.__name__"` went from (114, 112) ns per loop to (128, 125) ns per loop

An alternative option would be to extend the inline-supplement support beyond single pointers, and use it to store the enum signedness and any other useful information (such as a borrowed reference to the `__entries` dict to avoid an attribute lookup, and/or the `type_data::scope` pointer since enums are the only user of it after the type has been created). However, this would potentially be a breaking change if user code is using enum supplements for other purposes already, unless we introduce a second type of supplement.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1409,8 +1409,7 @@ Class binding annotations
 -------------------------
 
 The following annotations can be specified using the variable-length ``Extra``
-parameter of the constructor :cpp:func:`class_::class_` and
-:cpp:func:`enum_::enum_`.
+parameter of the constructor :cpp:func:`class_::class_`.
 
 .. cpp:struct:: is_final
 
@@ -1419,16 +1418,6 @@ parameter of the constructor :cpp:func:`class_::class_` and
 .. cpp:struct:: dynamic_attr
 
    Indicate that a type requires a Python dictionary to support the dynamic addition of attributes.
-
-.. cpp:struct:: is_enum
-
-   .. cpp:function:: is_enum(bool is_signed)
-
-      Mark the bound class as an enumeration backed by a signed or unsigned integer type.
-
-.. cpp:struct:: is_arithmetic
-
-   Indicate that the enumeration may be used as part of arithmetic operations.
 
 .. cpp:struct:: template <typename T> supplement
 
@@ -1455,6 +1444,27 @@ parameter of the constructor :cpp:func:`class_::class_` and
 
       Declares a callback that will be invoked when a C++ instance is first
       cast into a Python object.
+
+
+.. _enum_binding_annotations:
+
+Enum binding annotations
+------------------------
+
+The following annotations can be specified using the variable-length
+``Extra`` parameter of the constructor :cpp:func:`enum_::enum_`.
+Enums also support the :cpp:struct:`dynamic_attr`,
+:cpp:struct:`supplement`, and :cpp:struct:`type_slots` annotations
+documented for :ref:`classes <class_binding_annotations>`.
+
+.. cpp:struct:: is_arithmetic
+
+   Indicate that the enumeration may be used with arithmetic
+   operations.  This enables the binary operators ``+ - * // & | ^ <<
+   >>`` and unary ``- ~ abs()``, with operands of either enumeration
+   or integer type; the result will be a regular integer. It is
+   unspecified whether operations on mixed enum types (such as
+   ``Shape.Circle + Color.Red``) are permissible.
 
 Function binding
 ----------------
@@ -1831,8 +1841,8 @@ Class binding
 
       Bind the enumeration of type `T` to the identifier `name` within the
       scope `scope`. The variable length `extra` parameter can be used to pass
-      a docstring and other :ref:`class binding annotations
-      <class_binding_annotations>` such as :cpp:class:`is_arithmetic`.
+      a docstring and other :ref:`enum binding annotations
+      <enum_binding_annotations>` such as :cpp:class:`is_arithmetic`.
 
    .. cpp:function:: enum_ &value(const char * name, T value, const char * doc = nullptr)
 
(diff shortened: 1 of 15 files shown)


oremanj:
Size impact (all text) :
* `nb_type.cpp.o` 30913 -> 30948 bytes (+35)
* `nb_enum.cpp.o` 4921 -> 5096 bytes (+175)
* `test_holders.cpp.o` (contains a simple enum with two enumerators): 16919 -> 17007 bytes (+88)

This is balanced against the fact that `nb_type.cpp` no longer depends on `nb_enum.cpp`, so static-linking users that don't use enums no longer pay for them in binary size.

wjakob:
There is lots of great material in this PR, but it is mixing together some things that might be better to split apart to make merging manageable:

- `nb_init_prelim`, etc: Nice, I like the split into two data structures (with one being used for initialization). Minor preference: could you rename `_prelim` to `_init`? ("preliminary" seems a bit generic as a word, whereas this data is specifically for type initialization). This part seems ready to merge as-is.

- Inline supplement for types: I am less convinced by this one. I would prefer to make the supplement part of the type object regardless of its size (for example in an upcoming set of python bindings for my `drjit` library, the supplement is big -- >500 bytes). That means that nanobind must be able to allocate different versions of the `nb_type` metaclass.

- enumeration refactor: generally looks good to me, but I did not like this part:
  ```cpp
  static constexpr uint32_t max_slots = 40;
  PyType_Slot slots_storage[max_slots];
  ```
  Somehow it should be possible to create an ABI that does not involve the slot array in user code (unless they specifically want to add custom slots).

wjakob:
(Maintaining ABI compatibility in these development releases is not a priority by the way, so you can just bump the ABI number once in the first PR.)

oremanj:
Sure, I'll split this into separate PRs.

> Minor preference: could you rename _prelim to _init? ("preliminary" seems a bit generic as a word, whereas this data is specifically for type initialization). This part seems ready to merge as-is.

I used `_prelim` to match `func_data_prelim`, but happy to change it.

> Inline supplement for types: I am less convinced by this one. I would prefer to make the supplement part of the type object regardless of its size (for example in an upcoming set of python bindings for my drjit library, the supplement is big -- >500 bytes). That means that nanobind must be able to allocate different versions of the nb_type metaclass.

That seems great if we can get it, but I was daunted by the complexity of that solution, and I think it might have performance implications since you can no longer test `Py_TYPE(o) == internals.nb_type` as a quick check for nanobind instances. (I guess you could use a meta-metaclass and test `Py_TYPE(Py_TYPE(o))` for not much more cost though?) I put this in because it seemed like the minimum viable way to work around the "supplement must be trivially default constructible" requirement. If you're planning to support inline supplements more generally then I'll just drop it.

> enumeration refactor: generally looks good to me, but I did not like this part: [...] Somehow it should be possible to create an ABI that does not involve the slot array in user code (unless they specifically want to add custom slots).

Can you elaborate more on what aspects of exposing the slot array you're uncomfortable with? I did consider augmenting `type_data_prelim` to allow passing a "slot filler" callback in addition to the current list of user-provided slots, but the slot filler for enums needs access to the enum attributes, so now you're dealing with passing a closure too, and it's hard to validate that the callback doesn't overfill the array you give it, except after-the-fact.

wjakob:
> I used _prelim to match func_data_prelim, but happy to change it.

I was just looking into this. In the function case, it's actually the other way around (the function creation adds further data, so the initial record is indeed preliminary in a sense). With what you added, the initialization is trimmed down later on. So I think the naming makes sense.

> If you're planning to support inline supplements more generally then I'll just drop it.
Sounds good.

wjakob:
> Can you elaborate more on what aspects of exposing the slot array you're uncomfortable with?

It was the hardcoded limit of 40 and the fact that the caller has to stack-allocate space for what is arguably a nanobind implementation detail. But I don't have any better ideas after thinking about it for a bit more.

oremanj:
Are you comfortable assuming that no one is currently using `nb::supplement` with enums, such that it would be OK for enums to use the supplement for their own internal purposes? Or is that too much of a breaking change?

wjakob:
I don't expect people to use `nb::supplement` with enumerations. Do tell how that could be useful here, I don't quite follow..

oremanj:
I've split this PR into #194 and #195 and dropped the inline-supplement stuff.

PR #193, opened 2023-04-26
Add support for instance-level supplemental data

oremanj:
This is useful for caching hash values and similar things; there are a couple of motivating examples in the documentation. The data is limited to one pointer in size, trivially copyable, in order to minimize the performance impact.

Microbenchmark: ``python3.9 -m timeit -s "from test_classes_ext import Struct"`` ...
* ``"Struct(10)"`` (create instance from Python): goes from (61.5, 61.5) nsec/loop (two runs) to (63.1, 63.2) nsec/loop
* ``"Struct.create_reference()"`` (wrap a C++ pointer): goes from (69.2, 69.5) nsec/loop to (65, 64.3) nsec/loop

Size impact: ``size tests/CMakeFiles/nanobind-static.dir/__/src/nb_type.cpp.o``: text increases by 52 bytes on my system (30913 -> 30965)

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1434,6 +1434,13 @@ parameter of the constructor :cpp:func:`class_::class_` and
 
    Indicate that ``sizeof(T)`` bytes of memory should be set aside to store supplemental data in the type object.
 
+.. cpp:struct:: instance_supplement
+
+   Indicate that ``sizeof(uintptr_t)`` bytes of memory should be set
+   aside to store supplemental data in each *instance* of this
+   type. See the discussion of :ref:`supplemental instance data
+   <instance_supplement>`.
+
 .. cpp:struct:: type_slots
 
    .. cpp:function:: type_slots(PyType_Slot * value)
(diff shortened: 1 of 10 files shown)


oremanj:
The CI error appears unrelated to my change; have you seen that one before?

wjakob:
> The CI error appears unrelated to my change; have you seen that one before?

Yes, it's a flake that occasionally occurs on Windows, with pytest being the prime suspect. I reported it here: https://github.com/pytest-dev/pytest/issues/10843

wjakob:
For this PR, I am sorry to say that it seems a little beyond the scope of nanobind. Putting supplemental data into the type makes sense because there isn't really another place to put it. For instances, the far easier solution is to modify the type itself and store the data there. This requires modifying the code to be exposed in python, which is part of the mantra ("The codebase has to adapt to the binding tool and not the other way around")

oremanj:
Fair enough. The C++ type I wanted to use this with is small (8 bytes) and very widely deployed including in wire protocols, so it's not feasible to extend it with a cached hash field. In Python bindings it has a slow-to-compute hash in order to provide `hash(obj) == hash(str(obj))` for heterogeneous dict lookups. So I will probably wind up maintaining something like this out-of-tree, but I can see how it's a bit outside nanobind's typical domain.

PR #195, opened 2023-04-26
Reduce nb_type/nb_enum entanglement

oremanj:
Remove all enum-specific logic from `type_data` and `nb_type.cpp`; instead, `nb::enum_` now implements all of its additional functionality through the same interface available to user code (using `type_slots` and `supplement`). This serves as a demonstration of how nanobind users can implement their own enum-like semantics, and frees up four flag bits for other uses.

Cache the enum's scope, `is_signed` flag, and entries dictionary in supplemental type data. Caching the scope allows it to move out of `type_data` (enums were the only entity using it after type initialization was complete). Caching the entries dictionary makes enum lookups by value (used in `__new__`, `__name__`, and `__repr__`) substantially faster.

On my laptop, `python3.9 -m timeit -s "from test_enum_ext import Enum"` with the following microbenchmarks had these results:
* `"int(Enum.C)"` went from (51.2, 50.4) ns per loop (two separate runs) to (50.3, 51.3) ns per loop
* `"Enum(0xffffffff)"` went from (98.2, 97) ns per loop to (41.5, 41.2) ns per loop
* `"repr(Enum.C)"` went from (348, 341) ns per loop to (306, 291) ns per loop
* `"Enum.C.__name__"` went from (105, 109) ns per loop to (50, 50) ns per loop

Size impacts:
* `nb_type.cpp.o` text size increases by 169 bytes (31032 -> 31201)
* `nb_enum.cpp.o` text size increases by 194 bytes (4921 -> 5115)
* `nb_enum.cpp.o` rodata size increases by 32 bytes (832 -> 864)
* `test_holders.cpp.o` (which contains one enum definition and is source-level unchanged across this PR) text size increases by 52 bytes

Additionally, `nb_type.cpp` no longer depends on `nb_enum.cpp`, so static-linking users that don't use `nb::enum_` will no longer pay for it in binary size.

Split off from #192.

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -345,6 +345,21 @@ static int nb_type_setattro(PyObject* obj, PyObject* name, PyObject* value) {
             return rv;
         }
         Py_DECREF(cur);
+
+        const char *cname = PyUnicode_AsUTF8AndSize(name, nullptr);
+        if (!cname) {
+            PyErr_Clear(); // probably a non-string attribute name
+        } else if (cname[0] == '@') {
+            /* Prevent type attributes starting with an `@` sign from being
+               rebound or deleted. This is useful to safely stash owning
+               references. The ``nb::enum_<>`` class, e.g., uses this to ensure
+               indirect ownership of a borrowed reference in the supplemental
+               type data. */
+            PyErr_Format(PyExc_AttributeError,
+                         "internal nanobind attribute '%s' cannot be "
+                         "reassigned or deleted.", cname);
+            return -1;
+        }
     } else {
         PyErr_Clear();
     }
@@ -631,12 +646,7 @@ static PyTypeObject *nb_type_tp(nb_internals &internals,
 
 /// Called when a C++ type is bound via nb::class_<>
 PyObject *nb_type_new(const type_init_data *t) noexcept {
-    bool is_signed_enum    = t->flags & (uint32_t) type_flags::is_signed_enum,
-         is_unsigned_enum  = t->flags & (uint32_t) type_flags::is_unsigned_enum,
-         is_arithmetic     = t->flags & (uint32_t) type_flags::is_arithmetic,
-         is_enum           = is_signed_enum || is_unsigned_enum,
-         has_scope         = t->flags & (uint32_t) type_flags::has_scope,
-         has_doc           = t->flags & (uint32_t) type_init_flags::has_doc,
+    bool has_doc           = t->flags & (uint32_t) type_init_flags::has_doc,
          has_base          = t->flags & (uint32_t) type_init_flags::has_base,
          has_base_py       = t->flags & (uint32_t) type_init_flags::has_base_py,
          has_type_slots    = t->flags & (uint32_t) type_init_flags::has_type_slots,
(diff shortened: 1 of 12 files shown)


wjakob:
Nice -- I am on board with the callback bits.

One thing that is somehow concerning is that `Enum.__entries` still contains the enum value dictionary, while the type supplement stores a non-owning handle to that dictionary. If user code somehow manages to delete this dict or replace it, then enum operations that reference the non-owning handle turn into undefined behavior. It might be better to entirely hide `__entries` from the user and find a different way to deal with ownership.

Minor: there are a few `auto *supp ` where I think it would be nicer to write the actual type (I use `auto` mainly for unknowable or complicated "computed" types).

Lets' rename `fill_type_slots` to `type_slots_callback` and the `cb` member to `callback`.

I didn't understand the removal of `d.flags = type_flags::has_scope` in the normal `class_` constructor.

oremanj:
> One thing that is somehow concerning is that `Enum.__entries` still contains the enum value dictionary, while the type supplement stores a non-owning handle to that dictionary. If user code somehow manages to delete this dict or replace it, then enum operations that reference the non-owning handle turn into undefined behavior. It might be better to entirely hide `__entries` from the user and find a different way to deal with ownership.

Yep, I wondered about that too. I'm going to try an alternative where the dict reference is owned by a weakref callback on the type object, much like how `keep_alive` works.

> I didn't understand the removal of `d.flags = type_flags::has_scope` in the normal class_ constructor.

I removed the `has_scope` type flag entirely, since every type sets it so it's not particularly meaningful. `nb_type_new` now checks for `scope != nullptr` instead, which catches `nb::class_<T>(nullptr, "T")`.

oremanj:
> I'm going to try an alternative where the dict reference is owned by a weakref callback on the type object, much like how keep_alive works.

This didn't work because the GC can't find the weakref object, which makes sense once I think about it. I instead modified the existing `nb_tp_setattro` to forbid reassignment of attributes whose names begin with `@`, and stored the enum entries dict as `@entries`.

src/nb_type.cpp
```diff
@@ -662,9 +659,16 @@ PyObject *nb_type_new(const type_init_data *t) noexcept {
...
-    if (temp_tp->tp_members) {
+    // CPython copies member definitions into the heap type object,
+    // presumably to support something related to __slots__. PyPy doesn't,
+    // and without the Py_SIZE != 0 guard here we can wind up with
+    // tp->tp_members pointing to garbage. We only use tp_members to
+    // set __dictoffset__, and we already copied that above, so there's
+    // no harm in leaving the final tp_members set to nullptr on PyPy.
+    if (temp_tp->tp_members && Py_SIZE(temp_tp) != 0) {
         tp->tp_members = (PyMemberDef*)((char *)tp + Py_TYPE(tp)->tp_basicsize);
-        std::memcpy(tp->tp_members, temp_tp->tp_members, tp->tp_itemsize * Py_SIZE(temp_tp));
```
oremanj, line 667: On PyPy I was seeing:
* `tp->tp_itemsize == 0`
* `Py_TYPE(tp)->tp_itemsize == 0`
* `Py_SIZE(temp_tp) == 0`
so this memcpy wasn't copying anything. AFAICT it only worked in the past because `tp->tp_members` happened to point at a zero, so PyType_Ready didn't think there were any members. Something about this change tickled that uninitialized value to not be zero anymore, resulting in a segfault.

oremanj:
Lots of good stuff in that PR! Thanks for the heads up.

wjakob:
The issue is now solved, and I merged those changes into `master`.

wjakob:
FYI, I rebased to your changes here `https://github.com/wjakob/nanobind/tree/enum_untangle` but wasn't able to push them to your branch.

wjakob:
The PR looks good to me -- one last change request would be to move some of the documentation you've added to the headers into the Sphinx documentation (especially `type_slots_callback`). The documentation in the nanobind headers is intentionally really minimal, and the goal is to explain longer things in the Sphinx format that is easier to process and distribute.

oremanj:
I figured type_slots_callback was private — it’s in the detail:: namespace, mostly because the callback has a parameter whose type is private (detail::type_init_data*). It’s possible I’m not understanding the public/private distinction that nanobind uses for documentation purposes though; type_caster is also in detail::. I’m happy to add it to the docs if you endorse documenting a detail:: thing.

There’s already a test in test_enum.py that @entries isn’t modifiable or deletable; I take it you’re looking for a less enum-specific test?

wjakob:
> There’s already a test in test_enum.py that @entries isn’t modifiable or deletable; I take it you’re looking for a less enum-specific test?

I missed it!

> I figured type_slots_callback was private — it’s in the detail:: namespace, mostly because the callback has a parameter whose type is private (detail::type_init_data*). It’s possible I’m not understanding the public/private distinction that nanobind uses for documentation purposes though; type_caster is also in detail::. I’m happy to add it to the docs if you endorse documenting a detail:: thing.

Fair enough, but the long comment might better be placed elsewhere. There is a `type_slots_callback` type annotation which seems to not be used anywhere (`enum_` directly assigns to the raw field).

Btw, I may have accidentally deleted the ABI version change message from the changelog.

oremanj:
Thanks! I picked up your rebase, publicized and documented `type_slots_callback` with a warning that `detail::type_init_data` might change without warning.

wjakob:
Thank you!

PR #197, opened 2023-04-28
Dummy PR to run CI on several upcoming changes

wjakob:


include/nanobind/nb_error.h
```diff
@@ -17,7 +17,7 @@ struct error_scope {
 };
 
 /// Wraps a Python error state as a C++ exception
-class NB_EXPORT python_error : public std::exception {
+class NB_CORE python_error : public std::exception {
```
oremanj, line 20: I think this change is why you're seeing CI errors. Python 3.8-3.10 use static library builds in CI, while 3.11+ (which are working fine) use shared. If `!defined(NB_SHARED) && !defined(_WIN32)` then you're going from `__attribute__((visibility("default")))` to nothing (thus hidden, from the namespace-level attribute). Seems like that's causing the exception type you throw to not match the exception type you catch. Could be some kind of funky Mach-O thing?

PR #200, opened 2023-05-01
Propagate includes as SYSTEM

friendlyanon:
This prevents users from getting warnings in their code from 3rd party includes when the toolchain supports system headers.

This makes code like this unnecessary:

    get_property(
        nanobind_includes TARGET nanobind-static
        PROPERTY INTERFACE_INCLUDE_DIRECTORIES
    )
    set_target_properties(
        nanobind-static PROPERTIES
        FOLDER nanobind-static
        INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${nanobind_includes}"
    )

The above code marks usage requirements (i.e. headers in this case) as system, so they are included using the appropriate -isystem or /external:I flags.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -156,9 +156,11 @@ function (nanobind_build_library TARGET_NAME)
   endif()
 
   target_include_directories(${TARGET_NAME} PRIVATE
-    ${NB_DIR}/ext/robin_map/include)
+    ${NB_DIR}/ext/robin_map/include
+    ${Python_INCLUDE_DIRS}
+    ${NB_DIR}/include)
 
-  target_include_directories(${TARGET_NAME} PUBLIC
+  target_include_directories(${TARGET_NAME} SYSTEM INTERFACE
     ${Python_INCLUDE_DIRS}
     ${NB_DIR}/include)
 


friendlyanon:
I guess that code in the commit message could use some more context:

https://github.com/friendlyanon/mhef-pp/blob/8c7fafe09d0e0e0045c84d83fd536a948c2d04bc/python/CMakeLists.txt#L16-L26

Here the library is created then its includes are marked as system includes, so my code doesn't get warnings from those includes.

friendlyanon:
Also note that this doesn't silence warnings when building nanobind intself. One would still need to emulate [`cronic`](http://habilis.net/cronic/)/[`chronic`](http://joeyh.name/code/moreutils/) like so:

```yaml
    - name: Build
      shell: bash
      run: |
        (output="$(cmake --build build -t nanobind-static 2>&1)" \
        || (printf %s\\n "$output" && false))
        cmake --build build
```

wjakob:
As mentioned in the related discussion, this is not a behavior I would like to have as a default. It's important that people report warnings so that they can be addressed in both this project and in Python itself. (What is even the point of these warnings when people then disable them in the next step?)

friendlyanon:
> What is even the point of these warnings when people then disable them in the next step?

The problem is that from a consumer's PoV nanobind is 3rd party code. Any warnings they get are warnings they have to deal with while waiting for upstream to process the change request, make a release and waiting for software distributions to propagate the new version. There is a reason why system default implicit include locations are always handled as if `-isystem` by compilers and why CMake by default marks includes from `IMPORTED` targets as `SYSTEM`.  

How about following the [CMake model](https://cmake.org/cmake/help/latest/prop_tgt/NO_SYSTEM_FROM_IMPORTED.html) then and making the `SYSTEM`ness of the includes conditional in some way? Opt-in or opt-out would be fine by me either way.

Since you want these to be default, how about an opt-out such as this:

    nanobind_build_library(nanobind-static SYSTEM) # INTERFACE includes are SYSTEM

PR #201, opened 2023-05-01
Introduce the ALIAS target nanobind::nanobind

friendlyanon:
Quote from the documentation of target_link_libraries:

> Items containing ::, such as Foo::Bar, are assumed to be IMPORTED or
> ALIAS library target names and will cause an error if no such target
> exists. See policy CMP0028.

https://cmake.org/cmake/help/latest/command/target_link_libraries.html

Using such target names prevents users from accidentally making a typo or related error and silently getting a library link flag.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -120,6 +120,7 @@ function (nanobind_build_library TARGET_NAME)
     ${NB_DIR}/src/trampoline.cpp
     ${NB_DIR}/src/implicit.cpp
   )
+  add_library(nanobind::nanobind ALIAS ${TARGET_NAME})
 
   if (TARGET_TYPE STREQUAL "SHARED")
     nanobind_link_options(${TARGET_NAME})


friendlyanon:
Don't focus too much on the actual content of the PR. You can put anything you want after `nanobind::`, the point would be to have the option available and strongly recommend usage of it. The line could say `add_library(nanobind::${TARGET_NAME} ALIAS ${TARGET_NAME})` and that would also be completely fine by me.

wjakob:
>Using such target names prevents users from accidentally making a typo or related error and silently getting a library link flag.

I still don't understand. The official interface with nanobind is through `nanobind_add_module`, which deals with all of the targets internally. How you would use this alias target, and why does it have to be created by nanobind itself? Why would one make a typo that would create a library link flag?

cmake/nanobind-config.cmake
```diff
@@ -120,6 +120,7 @@ function (nanobind_build_library TARGET_NAME)
     ${NB_DIR}/src/trampoline.cpp
     ${NB_DIR}/src/implicit.cpp
   )
+  add_library(nanobind::nanobind ALIAS ${TARGET_NAME})
```
eirrgang, line 123: Do you mean the following?
```suggestion
  add_library(nanobind::${TARGET_NAME} ALIAS ${TARGET_NAME})
```

eirrgang (commented):
I don't think `nanobind_build_library()` or the `nanobind` target is used (directly) by consumers, is it?

This PR does not update the existing call to `target_link_libraries(${name} PRIVATE ${libname})`, so I don't see what utility it is providing.

eirrgang:
> The official interface with nanobind is through `nanobind_add_module`

In that case, this change only has potential utility to nanobind maintainer(s), I think.

> Why would one make a typo that would create a library link flag?

I guess it would help catch an error in line 259 or 261?

cmake/nanobind-config.cmake
eirrgang: I'm skeptical of the utility of this PR. I don't know see the advantage of this ALIAS target if nanobind is being used as documented. But it definitely seems inappropriate to erase the abstraction provided by this function by aliasing all TARGET_NAMEs to the same alias, even if only one target name is ever used.

If it is merely for the convenience of the consumer who wants to use the `nanobind*` targets directly, instead of through `nanobind_add_module`, then such a consumer could make the alias in their own project.

In the example linked by @friendlyanon , it looks like this is, in fact, intended to help with (undocumented) direct use of the `nanobind` target by the consumer build system, for which a statically named alias for the target would be useful. In that case, this alias target would be more appropriately declared at, say, line 274.

But, again, I don't think this is a use case that you, @wjakob , intend to support this way. If the consuming project can't find or set the details it needs through the CMake input variables and `nanobind_add_module()` It seems like what may actually be needed is some more output variables from the config module or PARENT variables from the CMake subdirectory.

In any case, after reviewing the linked example, I don't think my proposed change is what the author intended, so I'll withdraw the suggestion.

eirrgang: Oh, you're right. Sorry @wjakob . After the third time through, I finally saw the sentence in https://github.com/wjakob/nanobind/pull/201#issuecomment-1529898556 that makes my comment completely pointless. I'll delete.

wjakob:
In that link to `mhef-pp`, what is the difference from writing
```
add_library(nanobind::nanobind ALIAS nanobind-static)
```
in your own code? What is the "silly typo" you are referring to?

cmake/nanobind-config.cmake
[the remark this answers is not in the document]
friendlyanon: The interface is documented though https://nanobind.readthedocs.io/en/latest/api_cmake.html#low-level-interface

I just think it could be improved for those who want more control.

friendlyanon:
The user writing `add_library(nanobind::nanobind ALIAS nanobind-static)` could make a mistake in writing `nanobind-static`. But it's also more customary for projects to provide "namespaced" target names when vendored (via an alias) or when installed (the NAMESPACE argument).

I don't see a drawback in doing this personally.

wjakob:
After re-reading the discussion post and the PR discussion here, I still don't understand why this change has to be part of nanobind proper. I will close the PR.

PR #203, opened 2023-05-02
Forward the arguments inside call_guard wrapper lambda

arashbm:
Currently, binding a function that takes an rvalue reference (let's say `std::ranges::find`) only works if you don't pass a call guard. Otherwise it fails with `error: cannot bind rvalue reference of type ‘[$type]&&’ to lvalue of type ‘[$type]’`. Forwarding the arguments should fix this issue.

### DIFF
--- a/tests/test_functions.cpp
+++ b/tests/test_functions.cpp
@@ -80,6 +80,9 @@ NB_MODULE(test_functions_ext, m) {
     });
 
     /// Test call_guard feature
+    m.def("test_call_guard_wrapper_rvalue_ref", [](int&& i) { return i; },
+          nb::call_guard<my_call_guard>());
+
     m.def("test_call_guard", []() {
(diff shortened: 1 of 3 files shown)


tests/test_functions.cpp
```diff
@@ -80,6 +80,9 @@ NB_MODULE(test_functions_ext, m) {
     });
 
     /// Test call_guard feature
+    m.def("test_call_guard_wrapper_rvalue_ref", [](int&& i) { return i; },
```
wjakob, line 83: Does that test really fail if the `forward` is missing? It does not seem to me that way.

arashbm: I wrote the test first and checked that it failed, just to make sure the problem is not on my end. It only fails when call_guard is present, though.

wjakob:
Thanks!

PR #207, opened 2023-05-09
Fix surprising behavior in enum comparisons and arithmetic

oremanj:
* Enum equality comparisons (``==`` and ``!=``) now can only be true if both operands have the same enum type, or if one is an enum and the other is an `int`. This resolves some confusing results and ensures that enumerators of different types have a distinct identity, which is important if they're being put into the same set or used as keys in the same dictionary. All of the following were previously true but will now evaluate as false:
  * ``FooEnum(1) == BarEnum(1)``
  * ``FooEnum(1) == 1.2``
  * ``FooEnum(1) == "1"``

* Enum ordering comparisons (``<``, ``<=``, ``>=``, ``>``) and arithmetic operations (when using the :cpp:struct:`is_arithmetic` annotation) now require that any non-enum operand be a Python number (an object that defines ``__int__``, ``__float__``, and/or ``__index__``) and will avoid truncating non-integer operands to integers. Note that unlike with equality comparisons, ordering and arithmetic operations *do* still permit two operands that are enums of different types. Some examples of changed behavior:
  * ``FooEnum(1) < 1.2`` is now true (used to be false)
  * ``FooEnum(2) * 1.5`` is now 3.0 (used to be 2)
  * ``FooEnum(3) - "2"`` now raises an exception (used to be 1)

* Enum comparisons and arithmetic operations with unsupported types now return `NotImplemented` rather than raising an exception. This means equality comparisons such as ``some_enum == None`` will return unequal rather than failing; order comparisons such as ``some_enum < None`` will still fail, but now with a more informative error.

Size impact: `nb_enum.cpp.o` text size decreases by 877 bytes while rodata increases by 32 bytes.

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -145,19 +145,59 @@ static PyGetSetDef nb_enum_getset[] = {
 };
 
 PyObject *nb_enum_richcompare(PyObject *a, PyObject *b, int op) {
-    PyObject *ia = PyNumber_Long(a);
-    PyObject *ib = PyNumber_Long(b);
-    if (!ia || !ib)
-        return nullptr;
-    PyObject *result = PyObject_RichCompare(ia, ib, op);
-    Py_DECREF(ia);
-    Py_DECREF(ib);
+    // SomeType.tp_richcompare(a, b, op) is always invoked with 'a'
+    // having type SomeType. Note that this is different than binary
+    // arithmetic operations because comparisons can be reversed;
+    // Python will ask type(a) to check 'a > b' if type(b) doesn't
+    // know how to check 'b < a'.
+
+    if (op == Py_EQ || op == Py_NE) {
+        // For equality/inequality comparisons, only allow enums to be
+        // equal with their same enum type or with their underlying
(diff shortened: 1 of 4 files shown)


src/nb_enum.cpp
```diff
@@ -145,36 +145,109 @@ static PyGetSetDef nb_enum_getset[] = {
...
         if (!ia)                                                               \
             return nullptr;                                                    \
         PyObject *result = op(ia);                                             \
         Py_DECREF(ia);                                                         \
         return result;                                                         \
     }
 
+// Binary operands are trickier due to the potential for reversed operations.
+// We know either a or b is an enum object, but not which one.
+NB_NOINLINE PyObject *nb_enum_binop(PyObject *a, PyObject *b,
```
oremanj, line 210: This noinline saves about 3KB of code size. It probably does make enum binops a bit slower (there's an additional indirect call); happy to remove it if you'd prefer.

wjakob:
Beautiful, this looks great. I will merge it as-is, including the `NB_NOINLINE` part.

PR #209, opened 2023-05-09
Add an enum annotation that permits unenumerated values to be constructed

oremanj:
Enums in C++ are sometimes used for their type-safety, even when the values are not enumerated. See for example https://embeddedartistry.com/blog/2020/09/28/strongly-typed-integers-and-registers-in-c/. This PR adds support to nanobind for binding enums that are used in this way.

Text size cost: 619 bytes

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -138,12 +138,23 @@ static PyObject *nb_enum_new(PyTypeObject *subtype, PyObject *args, PyObject *kw
             PyObject *inst = inst_new_impl(subtype, nullptr);
             if (!inst)
                 goto error;
+            PyObject *args[2] = {inst, arg};
+#if PY_VERSION_HEX < 0x03090000
+            PyObject *result = nullptr;
+            PyObject *method = PyObject_GetAttrString(inst, "__setstate__");
+            if (method) {
+                result = _PyObject_Vectorcall(
+                    method, args + 1, 1 + PY_VECTORCALL_ARGUMENTS_OFFSET,
+                    nullptr);
+                Py_DECREF(method);
+            }
+#else
             PyObject *setstate_name = PyUnicode_FromString("__setstate__");
             check(setstate_name, "Could not construct string '__setstate__'!");
-            PyObject *args[2] = {inst, arg};
             PyObject *result = PyObject_VectorcallMethod(
                     setstate_name, args, 2 + PY_VECTORCALL_ARGUMENTS_OFFSET,
                     nullptr);
+#endif
             if (result) {
                 Py_DECREF(result);
                 return inst;
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -41,15 +41,25 @@ static PyObject *nb_enum_lookup(PyObject *self) {
 }
 
 static PyObject *nb_enum_repr(PyObject *self) {
-    PyObject *entry = nb_enum_lookup(self);
-    if (!entry)
-        return nullptr;
-
     PyObject *name = nb_inst_name(self);
-    PyObject *result =
-        PyUnicode_FromFormat("%U.%U", name, NB_TUPLE_GET_ITEM(entry, 0));
+    PyObject *entry = nb_enum_lookup(self);
+    PyObject *result;
+    if (!entry) {
+        PyErr_Clear();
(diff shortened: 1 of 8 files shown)


src/nb_enum.cpp
```diff
@@ -41,15 +41,25 @@ static PyObject *nb_enum_lookup(PyObject *self) {
...
+    if (!entry) {
+        PyErr_Clear();
+        enum_supplement &supp = nb_enum_supplement(Py_TYPE(self));
+        PyObject *int_val = supp.is_signed ? nb_enum_int_signed(self)
+                                           : nb_enum_int_unsigned(self);
+        if (!int_val) {
+            Py_DECREF(name);
+            return nullptr;
+        }
+        result = PyUnicode_FromFormat("%U(%S)", name, int_val);
```
oremanj, line 56: This is not conditional on `allow_unenumerated`, because you can still get unenumerated values without that annotation by returning them from C++, and it's a pretty bad debugging experience when `repr()` throws.

src/nb_enum.cpp
```diff
@@ -107,6 +117,48 @@ static PyObject *nb_enum_init(PyObject *, PyObject *, PyObject *) {
...
+    if (inst->ready) {
+        PyErr_SetString(PyExc_ValueError, "can't change enum value");
+        return false;
+    }
+    enum_supplement &supp = nb_enum_supplement(Py_TYPE(self));
+    size_t size = nb_type_data(Py_TYPE(self))->size;
+    void *inst_data = inst_ptr(inst);
+    const char *too = nullptr;
+    if (supp.is_signed) {
+        long long lval = PyLong_AsLongLong(value);
```
oremanj, line 131: I also tried switching on `size` and calling the appropriate `load_[iu](8|16|32|64)` function, but that was larger in code size.

src/nb_enum.cpp
```diff
@@ -107,6 +117,48 @@ static PyObject *nb_enum_init(PyObject *, PyObject *, PyObject *) {
     return 0;
 }
 
+static bool nb_enum_set_value(PyObject *self, PyObject *value) {
```
wjakob, line 120: It's a nice idea, but the amount of library code needed to realize the feature seems high to me (esp. the `nb_enum_set_value` function).

Could this feature also be realized on the user's end using a custom `nb::init<>` constructor? (With some library to perhaps automate the creation of that)

wjakob (commented):
The feature itself is nice and looks useful to me, but the amount of library code needed to realize it seems high to me (esp. the `nb_enum_set_value` function).

I am wondering: could this feature also be realized on the user's end using a custom `nb::init<>` constructor? (i.e. with less generic code and more specific code). In that case, nanobind could provide a simple construction to automate the creation of such a constructor.

oremanj:
That's an interesting idea. It's tricky because nanobind (reasonably) fails an `__init__` call if the instance is already initialized, but Python will always call `__init__` if `__new__` returns an instance of the expected type. We would only want `__init__` to be invoked if we're constructing an unenumerated value. It might be possible to work around this by having `__new__` explicitly invoke a differently-named method. I'll see how that works out size-wise.

oremanj:
I pushed an update following your suggestion, but I don't think it's a win. The library size decreases by 300 bytes but the size of `test_enum.o` (which contains two allow-unenumerated enums) increases by 1350 bytes.

oremanj:
The compiler was codegen'ing a `__setstate__` for all enums; when I fix that, the size impact on `test_enum.cpp.o` decreases to 439 bytes, which is a more reasonable tradeoff.

Note that a `__setstate__` (either the `.def()`'ed one or a thin wrapper around `nb_enum_set_value`) is required to implement pickle support for enums in a way that is compatible with pybind11 enum pickles. If you think that's an important or desirable feature, then IMO we should go with the `nb_enum_set_value` approach, because all enums will wind up using it.

wjakob:
Hi @oremanj -- I thought more about this PR. Even with the latest changes, it seems to me too big/complex of a change to address a narrow use case (this is obviously subjective).

When other projects need a non-enumerated enumeration, they can build it with existing nanobind API (just not using `nb_enum`).

oremanj:
OK, thanks for your consideration.

PR #212, opened 2023-05-11
Support enable_shared_from_this

oremanj:
Per discussion in https://github.com/wjakob/nanobind/discussions/183, this is not actually dangerous, although it does have some caveats (which I think I've thoroughly documented).

This turned out better than I thought it would: it is unexpectedly possible to support the feature where returning a raw pointer whose `enable_shared_from_this` subobject knows about a shared_ptr creates a Python object that shares ownership with that shared_ptr, without requiring `#include <memory>` anywhere in nanobind that doesn't already have it.

Size impact: `nb_type.cpp.o` increases by 148 bytes. Use of `std::shared_ptr<T>` in non-`enable_shared_from_this` cases actually gets smaller; we were previously instantiating a deleter for `PyObject*` by mistake, and now we're not.

### DIFF
--- a/docs/ownership_adv.rst
+++ b/docs/ownership_adv.rst
@@ -168,15 +168,17 @@ behaviors:
   identical to what would happen if the C++ code did ``return
   obj->shared_from_this();`` (returning an explicit
   ``std::shared_ptr<ST>`` to Python) rather than ``return obj;``.
-  The return value policy has no effect in this case; you will get
+  The return value policy has limited effect in this case; you will get
   shared ownership on the Python side regardless of whether you used
-  ``rv_policy::take_ownership`` or ``rv_policy::reference``.
+  `rv_policy::take_ownership` or `rv_policy::reference`.
+  (`rv_policy::copy` and `rv_policy::move` will still create a new
+  object that has no ongoing relationship to the returned pointer.)
 
   * Note that this behavior occurs only if such a ``std::shared_ptr<ST>``
     already exists! If not, then nanobind behaves as it would without
     ``enable_shared_from_this``: a raw pointer will transfer exclusive
     ownership to Python by default, or will create a non-owning reference
-    if you use ``rv_policy::reference``.
+    if you use `rv_policy::reference`.
 
 * If a Python object is passed to C++ as ``std::shared_ptr<ST> obj``,
   and there already exists an associated ``std::shared_ptr<ST>`` which
@@ -234,9 +236,10 @@ a problem for your application, you might get better results by using
 
 .. warning:: C++ code that receives a raw pointer ``T *obj`` *must not*
    assume that it has exclusive ownership of ``obj``, or even that
-   ``obj`` is allocated on the heap; ``obj`` might be a subobject of a
-   nanobind instance allocated from Python. This applies even if
-   ``T`` supports ``shared_from_this()`` and there is no associated
+   ``obj`` is allocated on the C++ heap (via ``operator new``);
+   ``obj`` might instead be a subobject of a nanobind instance
+   allocated from Python. This applies even if ``T`` supports
+   ``shared_from_this()`` and there is no associated
    ``std::shared_ptr``. Lack of a ``shared_ptr`` does *not* imply
    exclusive ownership; it just means there's no way to share ownership
    with whoever the current owner is.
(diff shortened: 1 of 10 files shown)


docs/ownership_adv.rst
```diff
@@ -139,19 +139,107 @@ true global reference count.
...
+* If a raw pointer ``ST *obj`` is returned from C++ to Python, and
+  there already exists an associated ``std::shared_ptr<ST>`` which
+  ``obj->shared_from_this()`` can locate, then nanobind will produce a
+  Python instance that shares ownership with it. The behavior is
+  identical to what would happen if the C++ code did ``return
+  obj->shared_from_this();`` (returning an explicit
+  ``std::shared_ptr<ST>`` to Python) rather than ``return obj;``.
+  The return value policy has no effect in this case; you will get
+  shared ownership on the Python side regardless of whether you used
+  ``rv_policy::take_ownership`` or ``rv_policy::reference``.
```
wjakob, line 173: What about `rv_policy::copy` or `rv_policy::move`? 🤔

docs/ownership_adv.rst
```diff
@@ -139,19 +139,107 @@ true global reference count.
...
+  managed by the ``shared_ptr`` deleter will be dropped,
+  and ``shared_from_this()`` will stop working. It can be reenabled by
+  passing the Python object back to C++ as ``std::shared_ptr<ST>`` once more,
+  which will create another control block.)
+
+Bindings for a class that supports ``enable_shared_from_this`` will be
+slightly larger than bindings for a class that doesn't, as nanobind
+must produce type-specific code to implement the above behaviors.
+
+.. warning:: The ``shared_from_this()`` method will only work when there
```
wjakob, line 210: Doesn't the body (`The [..]`) need to start on the line _following_ the `warning::` block? Maybe I just misremember.

docs/ownership_adv.rst
```diff
@@ -139,19 +139,107 @@ true global reference count.
...
-``std::enable_shared_from_this<T>``. This is a fundamental limitation:
-nanobind instances do not create a base shared pointer that declares
-ownership of an object. Other parts of a C++ codebase might then incorrectly
-assume ownership and eventually try to ``delete`` a nanobind instance
-allocated using ``pymalloc`` (which is undefined behavior). A compile-time
-assertion catches this and warns about the problem.
-
-Replacing shared pointers with :ref:`intrusive reference counting
-<intrusive>` fixes this limitations.
+enable_shared_from_this
```
wjakob, line 142: Awesome, this is really clear documentation of the nuances of this feature 👍

docs/ownership_adv.rst
```diff
@@ -139,19 +139,107 @@ true global reference count.
...
+
+There is no way to enable ``shared_from_this`` immediately upon
+regular Python-side object construction (i.e., ``SomeType(*args)``
+rather than ``SomeType.some_fn(*args)``). If this limitation creates
+a problem for your application, you might get better results by using
+:ref:`intrusive reference counting <intrusive>` instead.
+
+.. warning:: C++ code that receives a raw pointer ``T *obj`` *must not*
+   assume that it has exclusive ownership of ``obj``, or even that
+   ``obj`` is allocated on the heap; ``obj`` might be a subobject of a
```
wjakob, line 237: "on the heap" -> "on the C++ heap (via ``operator new``)"

include/nanobind/nb_class.h
```diff
@@ -386,6 +391,20 @@ class class_ : public object {
             }
         }
 
+        if constexpr (detail::has_shared_from_this_v<T>) {
+            d.flags |= (uint32_t) detail::type_flags::has_shared_from_this;
+            d.keep_shared_from_this_alive = [](PyObject *self) noexcept {
+                if (auto sp = inst_ptr<T>(self)->weak_from_this().lock()) {
```
wjakob, line 397: Could you add 1-line comment here explaining what this does? (I never used the `weak_from_this().lock()` construction).

include/nanobind/nb_class.h
```diff
@@ -386,6 +391,20 @@ class class_ : public object {
...
+                    detail::keep_alive(self, new auto(std::move(sp)),
```
wjakob, line 398: `new auto` 🤯

wjakob (commented):
This is beautiful work, thank you a lot! Just a few tiny comments/requests, then this can go in.

docs/ownership_adv.rst
oremanj: I've always written them this way, and the docs look correct when built. I think the only requirement is that continuation lines line up with the 'w' in `warning::`.

PR #216, opened 2023-05-23
nb_bound_method: add introspection attributes and fix surprising behavior around wrapper descriptors

oremanj:
The introspection attributes `__self__` and `__func__` provide better interface compatibility with regular Python bound methods, and can be cheaply supported by adding new entries in the existing member list.

The previous wholesale forwarding of attribute accesses to the underlying `nb_func` produced surprising behavior when explicitly invoking dunder methods: `obj.method()` would work but `obj.method.__call__()` would not (it would need `obj.method.__call__(obj)`). Same thing for `__repr__`, `__hash__`, etc. Fix this by checking the 'normal' attribute path before delegating to the contained method, except for `__doc__` and `__module__` where we want `nb_func`'s version even though the attribute does exist in `nb_bound_method`'s class dict as well. This is necessary to support `__self__` and `__func__` but I think it's a good idea regardless.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -113,6 +113,11 @@ Version 1.3.0 (TBD)
   on object ownership <enable_shared_from_this>` for more details.
   (PR `#212 <https://github.com/wjakob/nanobind/pull/212>`__).
 
+* Added introspection attributes ``__self__`` and ``__func__`` to nanobind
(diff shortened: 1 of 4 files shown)


oremanj:
CI failure is the usual windows stderr-capturing flake.

PR #219, opened 2023-05-26
Support `Eigen::Block` API

adriendelsalle:
Description
---

Support `Eigen::Block` API:
- disambiguate `Eigen::Ref`, `Eigen::Block` and `Eigen::Map`
- add tests

Closes https://github.com/wjakob/nanobind/discussions/218

Discussion
---

I'm not really happy with the composition pattern of casters.
This PR probably conflicts with https://github.com/wjakob/nanobind/pull/215
Happy to rebase or update on your recommendations!

### DIFF
--- a/include/nanobind/eigen/dense.h
+++ b/include/nanobind/eigen/dense.h
@@ -231,27 +231,18 @@ struct type_caster<T, enable_if_t<is_eigen_xpr_v<T> &&
     }
 };
 
-/** \brief Type caster for ``Eigen::Map<T>``
 
-  The ``Eigen::Map<..>`` type exists to efficiently access memory provided by a
-  caller. Given that, the nanobind type caster refuses to turn incompatible
-  inputs into a ``Eigen::Map<T>`` when this would require an implicit
-  conversion.
-*/
-
-template <typename T, int Options, typename StrideType>
-struct type_caster<Eigen::Map<T, Options, StrideType>,
-                   enable_if_t<is_eigen_plain_v<T> &&
-                               is_ndarray_scalar_v<typename T::Scalar>>> {
-    using Map = Eigen::Map<T, Options, StrideType>;
+template <typename T, int IS, int OS>
+struct map_base_caster_impl
+{
     using NDArray =
-        array_for_eigen_t<Map, std::conditional_t<std::is_const_v<Map>,
-                                                  const typename Map::Scalar,
-                                                  typename Map::Scalar>>;
+        array_for_eigen_t<T, std::conditional_t<std::is_const_v<T>,
+                                                  const typename T::Scalar,
+                                                  typename T::Scalar>>;
     using NDArrayCaster = type_caster<NDArray>;
     static constexpr bool IsClass = false;
     static constexpr auto Name = NDArrayCaster::Name;
-    template <typename T_> using Cast = Map;
+    template <typename T_> using Cast = T;
 
     NDArrayCaster caster;
 
@@ -265,13 +256,13 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
             return false;
 
         // Check for memory layout compatibility of non-contiguous 'Map' types
-        if constexpr (!is_contiguous_v<Map>)  {
+        if constexpr (!is_contiguous_v<T>)  {
             // Dynamic inner strides support any input, check the fixed case
-            if constexpr (StrideType::InnerStrideAtCompileTime != Eigen::Dynamic) {
+            if constexpr (IS != Eigen::Dynamic) {
                 // A compile-time stride of 0 implies "contiguous" ..
-                int64_t is_expected = StrideType::InnerStrideAtCompileTime == 0
+                int64_t is_expected = IS == 0
                                       ? 1 /*  .. and equals 1 for the inner stride */
-                                      : StrideType::InnerStrideAtCompileTime,
+                                      : IS,
                         is_actual = caster.value.stride(
                             (ndim_v<T> != 1 && T::IsRowMajor) ? 1 : 0);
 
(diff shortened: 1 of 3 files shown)


adriendelsalle:
Any clue about the failing tests related to pypy?
It looks like unrelated to the current changes.

wjakob:
The failing test is a flake, I worked around it in the meantime (if you rebase on master, it should no longer occur).

wjakob:
A naïve question (not having looked at the Eigen caster in a while).

We previously had `is_eigen_xpr_v` to take care of Eigen expressions that weren't a base array/matrix etc. Wouldn't a `Block` be such an expression template that could be handled with the previous type caster? 

You wrote:

> I'm not really happy with the composition pattern of casters.

What is it specifically that caused you to be unhappy, and what use case does this PR enable?

include/nanobind/eigen/dense.h
```diff
@@ -227,31 +252,12 @@ struct type_caster<Eigen::Map<T, Options, StrideType>, enable_if_t<is_eigen_plai
     }
 
     operator Map() {
-        NDArray &t = caster.value;
+        NDArray &t = caster.caster.value;
```
adriendelsalle, line 255: there is probably room for improvement here

adriendelsalle:
> The failing test is a flake, I worked around it in the meantime (if you rebase on master, it should no longer occur).

I just rebased on master, thanks!

> We previously had is_eigen_xpr_v to take care of Eigen expressions that weren't a base array/matrix etc. Wouldn't a Block be such an expression template that could be handled with the previous type caster?

AFAIK a `Block` is mostly similar to a `Map` with different specializations depending on contiguity, fixed/dynamic start/size/step.
A `Block` can be instantiated using the `block` bound-method of a `Matrix` or `Vector`, but also from calling the `operator()` as one of the 3 specializations available. The others are indexing (when passing integral types), and indexed views which are indeed expressions to handle the general case.

It's also possible to create a `Block` using available constructors.

This PR allows to return an `Eigen::Block` as mentioned in the discussion:
```cpp
    nb::class_<ClassWithEigenMember>(m, "ClassWithEigenMember")
        .def(nb::init<>())
        .def_rw("member", &ClassWithEigenMember::member)
        .def("get_block", [](ClassWithEigenMember& c){ return c.member(Eigen::seq(0, 1), Eigen::seq(0, 1)); }); 
```

Does it help?

wjakob:
PR #215 was merged in the meantime. Could I ask you to rebase this PR? 

PS: @WKarel — given your interest in the Eigen type casters, it would also be useful to have your feedback on it once rebased.

adriendelsalle:
Will do that this week! Thanks

adriendelsalle:
I just rebased, ping @WKarel @wjakob

tests/test_eigen.py
```diff
@@ -349,3 +349,35 @@ def test12_cast():
...
             t.castToRef03CnstVXi(v)
+
+@needs_numpy_and_eigen
+def test13_view():
+    c = t.ClassWithEigenMember()
+    view = c.get_view()
+    ref = np.ones((2, 2))
+    assert view.size == 4
+    assert view.shape == (2, 2)
+    assert np.all(view == ref)
```
WKarel, line 360: assert_array_equal(view, ref)

WKarel (commented):
I suggest to introduce many more, thorough tests.
So far, the tests return `Eigen::Block`s for the member variable of `ClassWithEigenMember` (which has a hard-coded type), with hard-coded offsets, sizes, and strides.
How about vectors, fixed-size matrices, different offsets, (negative) strides, const/readonly?
`type_caster` for `Eigen::Block` supports `from_python`, but this is not tested at all.
Instead of dedicated methods, I suggest to bind functions that accept as arguments instances of Eigen types (including `Eigen::Block`) and the offset, size, and stride of a block to be returned, and call them in parameterized tests, as with `test05_matrix_large_nonsymm`. The expected return values are then not hidden in C++, but defined in the Python test code. Possibly, these tests should be incorporated into `test05_matrix_large_nonsymm`.

adriendelsalle:
Happy to have your feedback @WKarel, I'm learning about `Eigen` in the same time I'm writing this PR :).

> I suggest to introduce many more, thorough tests.
> So far, the tests return Eigen::Blocks for the member variable of ClassWithEigenMember (which has a hard-coded type), with  hard-coded offsets, sizes, and strides.

I agree, I'm mostly testing about using the `operator()` of an `Array` or a `Matrix`, not using the `block` bound-method or `Block<...>` class directly

> How about vectors, fixed-size matrices, different offsets, (negative) strides, const/readonly?

Has to be tested as well

> type_caster for Eigen::Block supports from_python, but this is not tested at all.

TBH, I have not considered enough this case

> Instead of dedicated methods, I suggest to bind functions that accept as arguments instances of Eigen types (including Eigen::Block) and the offset, size, and stride of a block to be returned, and call them in parameterized tests, as with test05_matrix_large_nonsymm. The expected return values are then not hidden in C++, but defined in the Python test code. Possibly, these tests should be incorporated into test05_matrix_large_nonsymm.

If you do that you're also missing the fact that the return type of `operator()` is deduced at compile type between scalar/block/indexed view types. Using this approach would always fall into the most general case where we only know information at runtime. A block is also specializing differently if the template parameters are integral-like types or not (`all` placeholder)

See
https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/plugins/IndexedViewMethods.h#L46
https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/plugins/IndexedViewMethods.h#L199

So it's probably worth to test more than this, also very particular, case

wjakob:
Hi @adriendelsalle,

just checking in on this PR's status. Will you be able to incorporate the feedback by @WKarel?

Thanks,
Wenzel

adriendelsalle:
Hi @wjakob 👋
Yes sure, but I guess it will probably take some weeks before I manage to find a slot for that.
Is it acceptable?

wjakob:
The issue has been open since May -- let me close it for now and reopen when it's ready for feedback.

PR #221, opened 2023-05-31
fix: support ABI3 on Windows

henryiii:
Fixing ABI3 support.


cmake/nanobind-config.cmake
```diff
@@ -195,7 +195,11 @@ endfunction()
 
 function(nanobind_extension_abi3 name)
   get_filename_component(ext "${NB_SUFFIX}" LAST_EXT)
-  set_target_properties(${name} PROPERTIES PREFIX "" SUFFIX ".abi3${ext}")
+  if(WIN32)
+    set_target_properties(${name} PROPERTIES PREFIX "" SUFFIX ".abi3${ext}")
+  else()
+    set_target_properties(${name} PROPERTIES PREFIX "" SUFFIX "${ext}")
```
wjakob, line 201: I believe aDid you leave out a period here? (before `${ext}`)

wjakob:
My proposed version: #222

cmake/nanobind-config.cmake
```diff
@@ -195,7 +195,11 @@ endfunction()
 
 function(nanobind_extension_abi3 name)
   get_filename_component(ext "${NB_SUFFIX}" LAST_EXT)
-  set_target_properties(${name} PROPERTIES PREFIX "" SUFFIX ".abi3${ext}")
+  if(WIN32)
+    set_target_properties(${name} PROPERTIES PREFIX "" SUFFIX ".abi3${ext}")
```
henryiii, line 199: I got this exactly backwards. 🤦

henryiii: Going to go ahead and flip these so I can test scikit-build-core, will switch to #222 when it's ready.

PR #222, opened 2023-05-31
Fix naming of ABI3 extension libraries on Windows

wjakob:
This commit fixes the naming conventions of ABI3 builds on Windows. It also adopts a recent CMake/FindPython interface whenever available.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -4,18 +4,25 @@ if (NOT TARGET Python::Module)
   message(FATAL_ERROR "You must invoke 'find_package(Python COMPONENTS Interpreter Development REQUIRED)' prior to including nanobind.")
 endif()
 
-# Determine the right suffix for ordinary and stable ABI extensions
-if (Python_SOSABI)
-  set(NB_SUFFIX_EXT "${CMAKE_SHARED_MODULE_SUFFIX}")
+# Determine the right suffix for ordinary and stable ABI extensions.
+# Python_SOSABI is guaranteed to be available in CMake 3.26+, and it may
+# also be available as part of backported FindPython in scikit-build-core
+if (DEFINED Python_SOSABI)
   if (WIN32)
     set(NB_SUFFIX_EXT ".pyd")
+  else()
+    set(NB_SUFFIX_EXT "${CMAKE_SHARED_MODULE_SUFFIX}")
   endif()
 
-  # These are guaranteed to both be available in CMake 3.26+
   set(NB_SUFFIX   ".${Python_SOABI}${NB_SUFFIX_EXT}")
-  set(NB_SUFFIX_S ".${Python_SOSABI}${NB_SUFFIX_EXT}")
+
+  if (Python_SOSABI STREQUAL "")
+    set(NB_SUFFIX_S "${NB_SUFFIX_EXT}")
+  else()
+    set(NB_SUFFIX_S ".${Python_SOSABI}${NB_SUFFIX_EXT}")
+  endif()
 else()
-  # Determine the Python extension suffix and stash in the CMake cache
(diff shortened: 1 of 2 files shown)


henryiii:
Though I guess `Python_SOABI` was incorrect for PyPy until recently.

wjakob:
`Python_SOABI` doesn't include the final part of the extension (`.dll` / `.so`). Is there a portable way to get that? I am asking because of the inconsistency on macOS, where `.so` is used instead of `.dylib`.

cmake/nanobind-config.cmake
```diff
@@ -4,20 +4,34 @@ if (NOT TARGET Python::Module)
...
-  OUTPUT_VARIABLE NB_SUFFIX
-  OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-if (NB_SUFFIX_RET AND NOT NB_SUFFIX_RET EQUAL 0)
-  message(FATAL_ERROR "nanobind: Python sysconfig query to "
-    "find 'EXT_SUFFIX' property failed!")
+# Determine the the right suffix for ordinary and stable ABI extensions
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.26)
+  set(NB_SUFFIX   ".${Python_SOABI}")
+  set(NB_SUFFIX_S ".${Python_SOSABI}")
```
henryiii, line 10: The SOABI/SOSABI values do not include the suffix (.so/.pyd).

cmake/nanobind-config.cmake
```diff
@@ -4,20 +4,40 @@ if (NOT TARGET Python::Module)
...
-    "import sysconfig; print(sysconfig.get_config_var('EXT_SUFFIX'))"
-  RESULT_VARIABLE NB_SUFFIX_RET
-  OUTPUT_VARIABLE NB_SUFFIX
-  OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-if (NB_SUFFIX_RET AND NOT NB_SUFFIX_RET EQUAL 0)
-  message(FATAL_ERROR "nanobind: Python sysconfig query to "
-    "find 'EXT_SUFFIX' property failed!")
+# Determine the right suffix for ordinary and stable ABI extensions
+if (Python_SOSABI)
```
henryiii, line 8:
```suggestion
if (DEFINED Python_SOSABI)
```

It'll be empty on Windows, which is false.

cmake/nanobind-config.cmake
```diff
@@ -4,20 +4,40 @@ if (NOT TARGET Python::Module)
...
+  set(NB_SUFFIX_EXT "${CMAKE_SHARED_MODULE_SUFFIX}")
```
henryiii, line 9: Personally, I'd put this in an else, rather than overriding optionally the line below, I think.

cmake/nanobind-config.cmake
wjakob: Thanks!

cmake/nanobind-config.cmake
```diff
@@ -4,20 +4,40 @@ if (NOT TARGET Python::Module)
...
+  if (WIN32)
+    set(NB_SUFFIX_EXT ".pyd")
+  endif()
+
+  # These are guaranteed to both be available in CMake 3.26+
+  set(NB_SUFFIX   ".${Python_SOABI}${NB_SUFFIX_EXT}")
+  set(NB_SUFFIX_S ".${Python_SOSABI}${NB_SUFFIX_EXT}")
```
henryiii, line 16: Wouldn't this have two `..` if `Python_SOSABI` is empty?

wjakob: yes, it would.. 🤦

wjakob:
Merging in favor of #221.

PR #226, opened 2023-06-02
fix: use FindPython's suffix computation if possible

henryiii:
This should speed up the configuration by sometimes reducing the calls to Python of CMake is new enough. 

Testing in https://github.com/wjakob/nanobind_example/pull/27, but the cross compile situation requires further work.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -5,41 +5,57 @@ if (NOT TARGET Python::Module)
 endif()
 
 # Determine the right suffix for ordinary and stable ABI extensions.
+
+# We always need to know the extension
+if(WIN32)
+  set(NB_SUFFIX_EXT ".pyd")
+else()
+  set(NB_SUFFIX_EXT "${CMAKE_SHARED_MODULE_SUFFIX}")
+endif()
+
+# This was added in CMake 3.17+, also available earlier in scikit-build-core.
+# PyPy sets an invalid SOABI (platform missing), causing older FindPythons to
+# report an incorrect value. Only use it if it looks correct (X-X-X form).
+if(DEFINED Python_SOABI AND "${Python_SOABI}" MATCHES ".+-.+-.+")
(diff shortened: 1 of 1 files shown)


cmake/nanobind-config.cmake
```diff
@@ -5,41 +5,57 @@ if (NOT TARGET Python::Module)
...
+  set(NB_SUFFIX_EXT ".pyd")
+else()
+  set(NB_SUFFIX_EXT "${CMAKE_SHARED_MODULE_SUFFIX}")
+endif()
+
+# This was added in CMake 3.17+, also available earlier in scikit-build-core.
+# PyPy sets an invalid SOABI (platform missing), causing older FindPythons to
+# report an incorrect value. Only use it if it looks correct (X-X-X form).
+if(DEFINED Python_SOABI AND "${Python_SOABI}" MATCHES ".+-.+-.+")
+    set(NB_SUFFIX ".${Python_SOABI}${NB_SUFFIX_EXT}")
```
wjakob, line 20: super-minor: Is the indentation too deep here?

henryiii: Probably. I usually use `clang-format` in `pre-commit` so I don't have to think about things like that. :)

I'll run the downstream test first before fixes.

wjakob:
Looks all fine to me, let me know when it's ready to merge.

cmake/nanobind-config.cmake
henryiii:
```suggestion
  set(NB_SUFFIX ".${Python_SOABI}${NB_SUFFIX_EXT}")
```

henryiii:
I think it's a nice little cleanup, though it doesn't fix cross compiles - that's a problem I'll have to continue to work on in scikit-build-core.

wjakob:
Can I convert to 'ready for review' and merge? The failing test is a fluke caused by pytest.

henryiii:
Yes, sorry, thought I converted it.

PR #231, opened 2023-06-11
Fix bool type detection for read-only ndarray construction

yosh-matsuda:
This PR fixes the construction of a read-only bool ndarray that is incorrectly treated as a uint type.


wjakob:
Thanks!

PR #236, opened 2023-06-18
Accept `none` arg in RW properties at 1st positional annotation

yosh-matsuda:
I noticed that the RW properties can't accept `nb::arg().none()`, eg:

```c++
    auto cls = nb::class_<Struct>(m, "Struct")
        .def_prop_rw("none_prop", [](Struct *) {}, [](Struct *, nb::handle) {}, nb::arg().none())
        .def_prop_rw_static("none_prop_static", []() {}, [](nb::handle) {}, nb::arg().none());
```

The above code causes that the static assertion fails with `The number of nb::arg annotations must match the argument count!` because the extra arguments is passed to both getter and setter, like:

```
class_ &def_prop_rw(const char *name_, Getter &&getter, Setter &&setter, const Extra &...extra) {
        get_p = cpp_function((detail::forward_t<Getter>) getter, scope(*this), is_method(), rv_policy::reference_internal, extra...);
        set_p = cpp_function((detail::forward_t<Setter>) setter, scope(*this), is_method(), extra...);
```

Is this by design? If not, I would like to ask you about how to fix it. There are several options on how to separate the extra arguments for the setter and the getter.

1. **(Changes in this PR)** Add the overload that the `nb::arg` type can be accepted only at the first extra argument, eg:

```c++
class_ &def_prop_rw(const char *name_, Getter &&getter, Setter &&setter, const arg &arg_, const Extra &...extra) {
        get_p = cpp_function((detail::forward_t<Getter>) getter, scope(*this), is_method(), rv_policy::reference_internal, extra...);
        set_p = cpp_function((detail::forward_t<Setter>) setter, scope(*this), is_method(), arg_, extra...);
```

2. The `nb::arg().none()` is always passed to the setter. This may be preferable to having `nb::arg` specified by users since the name or default value of arg for properties is meaningless.

```c++
class_ &def_prop_rw(const char *name_, Getter &&getter, Setter &&setter, const Extra &...extra) {
        get_p = cpp_function((detail::forward_t<Getter>) getter, scope(*this), is_method(), rv_policy::reference_internal, extra...);
        set_p = cpp_function((detail::forward_t<Setter>) setter, scope(*this), is_method(), nb::arg().none(), extra...);
```

3. Implement compile time removal of `nb::arg` in extra arguments and set them to getters. We need additional meta classes and functions. I can try it but it may seem exaggerated implementation.

@wjakob  Do you have any idea?

### DIFF
--- a/include/nanobind/nb_class.h
+++ b/include/nanobind/nb_class.h
@@ -440,7 +440,9 @@ public:
         return *this;
     }
 
-    template <typename Getter, typename Setter, typename... Extra>
+    template <
+        typename Getter, typename Setter, typename... Extra,
+        detail::enable_if_t<((!std::is_same_v<Extra, arg> && !std::is_same_v<Extra, arg_v>) && ...)> = 0>
     NB_INLINE class_ &def_prop_rw(const char *name_, Getter &&getter,
                                   Setter &&setter, const Extra &...extra) {
         object get_p, set_p;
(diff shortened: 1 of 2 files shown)


wjakob:
It seems too much of a specific case to warrant adding this much code to the mainline repository. I'm open to fixing this issue, but the solution will need to be much more compact. Alternatively, this issue should be easy to work around on your end by providing a custom `def_prop_rw` in a subclass of `class_` that you use instead in your project.

yosh-matsuda:
OK. I don't have a compact solution at the moment. I will use a custom `def_prop_rw` you suggested for now. Thanks!

wjakob:
I found a simpler way to implement this: https://github.com/wjakob/nanobind/commit/11a9b3c318f2f262509ccf4090a57699ea860cad

yosh-matsuda:
@wjakob 
That's great! Thank you very much.

PR #237, opened 2023-06-19
Add missing cmake parameter

robquant:
This fixes a small bug in the cmake function `nanobind_compile_options`, where the `name` parameter was missing.
I also noticed a small error in the documentation of `nanobind_add_module`, not sure if this should go into a separate pull request.


wjakob:
Oh wow, I am surprised that this worked previously. Thanks for both corrections!d

PR #242, opened 2023-06-24
Changes nb::tuple() to construct an empty tuple.

### LINKED ISSUE #241: nanobind::tuple default constructor creates a null handle
When porting some old pybind11 code, I found that `nb::tuple{}` creates a `tuple` class with a nullptr handle, which promptly segfaults if doing anything with it. I feel like in order to be consistent with `list` and `dict`, this should have a default constructor like:

```
tuple() : object(PyTuple_New(0), detail::steal_t()) { }
```

stellaraccident:
This is consistent with both `list` and `dict`. However, it looks like this might have been done this way for a reason so I'm posting this patch for comment.

Default initializing to nullptr differs from what pybind11 does, and I found this difference due to a segfault while porting some existing code.

Fixes #241

### DIFF
--- a/include/nanobind/nb_types.h
+++ b/include/nanobind/nb_types.h
@@ -413,7 +413,8 @@ class bytes : public object {
 };
 
 class tuple : public object {
-    NB_OBJECT_DEFAULT(tuple, object, "tuple", PyTuple_Check)
+    NB_OBJECT(tuple, object, "tuple", PyTuple_Check)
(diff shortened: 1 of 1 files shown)


wjakob:
Closed via a more comprehensive fix in https://github.com/wjakob/nanobind/commit/506185dca821c9cc1268c33b4cc867ae20f0fc4b.

wjakob:
(Thanks for reporting!)

PR #253, opened 2023-08-07
fix MSVC compilation error in DEBUG mode due to free macro

polyvertex:
Thanks for your library!

Using nanobind release 1.5.0, I faced the following compilation error with MSVC 2022 in DEBUG mode (`_DEBUG` defined) only due to macro `free` being defined as `_free_dbg` by Windows headers:

```
nb_func.cpp(94,20): error C2039: '_free_dbg': is not a member of 'nanobind::detail::func_data'
```

I am aware you wish to keep the project easily maintainable and this is a quick and least intrusive fix, however I am not yet familiar with nanobind so not sure it is the most ideal or correct approach.

Anyway here it is just in case :)

### DIFF
--- a/include/nanobind/nb_attr.h
+++ b/include/nanobind/nb_attr.h
@@ -108,7 +108,7 @@ enum class func_flags : uint32_t {
     is_implicit = (1 << 12),
     /// Is this function an arithmetic operator?
     is_operator = (1 << 13),
-    /// When the function is GCed, do we need to call func_data_prelim::free?
+    /// When the function is GCed, do we need to call func_data_prelim::free_capture?
     has_free = (1 << 14),
     /// Should the func_new() call return a new reference?
     return_ref = (1 << 15),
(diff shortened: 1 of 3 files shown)


wjakob:
This isn't a new issue, correct? I suppose that variable has been named `free` for a long time (since commit 765253b1ade5e9ce4f7394c723632833896fb254 in Feb 2022). It was called `free_capture` before then. Can you rename it to that? (instead of `freemem`)

polyvertex:
> This isn't a new issue, correct?

If you refer to the definition of macro `free` by Windows headers, no it is not new. No pun intended :)

> It was called `free_capture` before then. Can you rename it to that?

Done!

wjakob:
With "new", I was referring to the nanobind version. I assume that the failure also occurs on v1.4.0?

polyvertex:
> With "new", I was referring to the nanobind version. I assume that the failure also occurs on v1.4.0?

I did not test, but a quick glance at the source of 1.4.0 shows that it certainly does yes. For the same reason indeed. I.e. struct `func_data_prelim` has a member named `free`, which name collides with the macro defined by Windows headers

PR #254, opened 2023-08-08
Fix `std::shared_ptr` code example in documentation

nicholasjng:
Eliminates the small typo `share_ptr ` ->  `shared_ptr`.


wjakob:
Good catch, thanks!

Issue #255, opened 2023-08-09
Type Caster for std::vector<bool>


maximiliank:
The Type Caster for STL `std::vector<bool>` is not working due to the following range based for loop:

https://github.com/wjakob/nanobind/blob/90b267785f23dc91ea0912c9343780df3f23c695/include/nanobind/stl/detail/nb_list.h#L62

See https://stackoverflow.com/a/34085592/9759769.

An example demonstrating the error would be:

```cpp
#include <nanobind/nanobind.h>
#include <nanobind/stl/vector.h>
#include <vector>

namespace nb = nanobind;

using namespace nb::literals;


// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables, misc-use-anonymous-namespace, readability-identifier-length)
NB_MODULE(nanobind_example_ext, m)
{
    m.def("vector_bool", []() { return std::vector<bool>{false, true, false, true}; });
}
```

which yields on GCC 12.3.0:

```shell
[build] /nanobind_example/.venv/lib/python3.11/site-packages/nanobind/include/nanobind/stl/detail/nb_list.h:62:13: error: cannot bind non-const lvalue reference of type ‘std::_Bit_reference&’ to an rvalue of type ‘std::_Bit_iterator::reference’
[build]    62 |             for (auto &value : src) {
[build]       |             ^~~
[build] ninja: build stopped: subcommand failed.
```

wjakob:
Closed via #256.

PR #256, opened 2023-08-09
Update nb_list.h to allow for std::vector<bool>, see #255

maximiliank:
See https://stackoverflow.com/a/34079497/9759769


wjakob:
Looks great! Could you add a test as well (which would not compile without the change).

Issue #259, opened 2023-08-13
[BUG]: Fail to compile on Linux x86, pp310


laggykiller:
### Problem description

nanobind failed to compile on Linux x86, pp310.

I am using nanobind as a submodule in my project.
```
 /project/nanobind/src/nb_type.cpp:426:12: error: ‘Py_am_send’ was not declared in this scope
    426 |     Ei<i1, Py_##p2##_##name,                           \
        |            ^~~
  /project/nanobind/src/nb_type.cpp:517:5: note: in expansion of macro ‘E’
    517 |     E(81, as_async, am, send),
        |     ^
  /project/nanobind/src/nb_type.cpp:428:52: error: no matching function for call to ‘Ei<81, <expression error>, 208, 220>()’
    428 |        offsetof(PyHeapTypeObject, p1.p2##_##name)>()
        |                                                    ^
  /project/nanobind/src/nb_type.cpp:517:5: note: in expansion of macro ‘E’
    517 |     E(81, as_async, am, send),
        |     ^
  /project/nanobind/src/nb_type.cpp:398:83: note: candidate: ‘template<unsigned int I1, unsigned int I2, unsigned int Offset1, unsigned int Offset2> constexpr nanobind::detail::nb_slot nanobind::detail::Ei()’
    398 | template <size_t I1, size_t I2, size_t Offset1, size_t Offset2> nb_slot constexpr Ei() {
        |                                                                                   ^~
  /project/nanobind/src/nb_type.cpp:398:83: note:   template argument deduction/substitution failed:
  /project/nanobind/src/nb_type.cpp:428:52: error: template argument 2 is invalid
    428 |        offsetof(PyHeapTypeObject, p1.p2##_##name)>()
        |                                                    ^
  /project/nanobind/src/nb_type.cpp:517:5: note: in expansion of macro ‘E’
    517 |     E(81, as_async, am, send),
        |     ^
  gmake[2]: *** [CMakeFiles/nanobind-static.dir/nanobind/src/nb_type.cpp.o] Error 1
  gmake[2]: *** Waiting for unfinished jobs....
  gmake[1]: *** [CMakeFiles/nanobind-static.dir/all] Error 2
  gmake: *** [all] Error 2
```

Full log from my GitHub runner: https://github.com/laggykiller/apngasm-python/actions/runs/5847075688/job/15852936378

laggykiller:
@tttapa I tried to use your fork but the problem is still not solved...

[...]

PR #263, opened 2023-08-14
Exclude Py_am_send for PyPy

tttapa:
Addresses #259 by not using `Py_am_send` when compiling for PyPy, even if the Python version is >= 3.10.


wjakob:
Resolves #259. A more fine-grained workaround could be added once this issue on the PyPy issue tracker is resolved: https://foss.heptapod.net/pypy/pypy/-/issues/3990.

PR #264, opened 2023-08-14
clang-tidy fixups

NAThompson:
nanobind is built in userspace, and userspace warning policies are applied. If `clang-tidy` warnings are enabled, we see messages such as:

```
~venv/lib/python3.9/site-packages/nanobind/src/nb_type.cpp:1587:5: warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
    nb_inst *nbi = (nb_inst *) o;
    ^~~~~~~
    auto
~venv/lib/python3.9/site-packages/nanobind/src/error.cpp:213:20: warning: use '= default' to define a trivial destructor [modernize-use-equals-default]
builtin_exception::~builtin_exception() { }
                   ^                    ~~~
                                        = default;
~venv/lib/python3.9/site-packages/nanobind/src/nb_type.cpp:1167:18: warning: an exception may be thrown in function 'nb_type_put_common' which should not throw exceptions [bugprone-exception-escape]
static PyObject *nb_type_put_common(void *value, type_data *t, rv_policy rvp,
                 ^
```

The first two messages are nitpicks, but the third seems to be a genuine problem.

Fix these `clang-tidy` warnings.

### DIFF
--- a/include/nanobind/nb_lib.h
+++ b/include/nanobind/nb_lib.h
@@ -246,7 +246,7 @@ NB_CORE PyObject *capsule_new(const void *ptr, const char *name,
 // ========================================================================
 
 /// Create a Python function object for the given function record
-NB_CORE PyObject *nb_func_new(const void *data) noexcept;
+NB_CORE PyObject *nb_func_new(const void *data);
 
 // ========================================================================
 
@@ -357,11 +357,11 @@ NB_CORE std::pair<bool, bool> nb_inst_state(PyObject *o) noexcept;
 
 // Create and install a Python property object
 NB_CORE void property_install(PyObject *scope, const char *name,
-                              PyObject *getter, PyObject *setter) noexcept;
+                              PyObject *getter, PyObject *setter);
 
 NB_CORE void property_install_static(PyObject *scope, const char *name,
                                      PyObject *getter,
-                                     PyObject *setter) noexcept;
+                                     PyObject *setter);
 
 // ========================================================================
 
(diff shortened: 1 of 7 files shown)


wjakob:
My issue with clang-tidy is that many of its warnings are opinionated, and I usually don't agree with its opinions. For all of the ones you reported, my answer is a categorical "no" ;).

1. Presumably incorrect `noexcept` declaration: functions that can throw require stack unwind code + tables (`.eh_frame` section in ELF binary) to clean up, which can add significant binary size overheads (which is an area of concern in C++ metatemplate-based binding tools like nanobind).

   The function `nb_func_new` you wanted to change tends to be called thousands of times to create the individual function bindings, so it's absolutely crucial that this one stays `noexcept`. If an exception is thrown inside, then that is *not* undefined behavior but rather a severe error condition that will invoke `abort()`, which is the preferable behavior in this case.

3. `auto` to avoid duplication: whenever a type name isn't overly long (i.e. *not*  something like ``std::map<key_type, value_type>::iterator``), I find it vastly more readable to just write out what it is called.

4. This is a polymorphic type, which means that destructor+vtable+RTTI type info have to be generated somewhere. If you declare everything in the header file, those three items will be generated in every single compilation unit and then have to be merged.

I don't think it's wrong to occasionally look at what clang-tidy reports, but my experience with it is that fixing reported warnings often makes things worse. Incorporating clang-tidy into a project (e.g. via CI) adds significant churn.

PR #280, opened 2023-08-24
Minor grammatical fix in api_extra.rst

torokati44:


wjakob:
I don't see a grammatical problem. Can you point to a source saying why this expression is incorrect?

torokati44:
Sure:

https://www.grammarphobia.com/blog/2013/03/whether-or-not.html
> Pat discusses this in her grammar and usage book [Woe Is I](https://www.grammarphobia.com/books-html/books_woe-html). Here’s the passage:
> “When you’re talking about a choice between alternatives, use whether: Richie didn’t know whether he should wear the blue suit or the green one. The giveaway is the presence of or between the alternatives. But when there’s a whether or not choice (Richie wondered whether or not he should wear his green checked suit), you can usually drop the or not and use either whether or if: Richie wondered if [or whether] he should wear his green checked suit. You’ll need or not, however, if your meaning is ‘regardless of whether’: Richie wanted to wear the green one, whether it had a gravy stain or not. (Or, if you prefer, whether or not it had a gravy stain.)”

https://dictionary.cambridge.org/dictionary/english/whether-or-not
> used for saying that it is not important which of two possibilities is true

----

The "whether or not" phrasing, at least to me, means:
 - either that it doesn't matter which case it is, which is not right in the corrected sentence,
 - or that the two possibilities are logical opposites (with only one of them written down), which is also not the case.

The corrected sentence presents two alternatives (`self` on the left side vs. `self` on the right side), in which case a simple "whether" is more appropriate in my opinion.
Of course, one could argue that left and right are logical opposites, but in that sentence, both using "whether or not", and then spelling out the two alternatives, comes across as strange to me.
It would also be fine as "Depending on whether or not `self` is the left argument of a binary operation", but I think the other variant fits better here.

wjakob:
Got it, that makes sense -- thank you. Could you rebase the PR based on master? Looks like it bundles some unrelated changes.

wjakob:
Thanks!

PR #282, opened 2023-08-25
Mark c_str() methods of nb::str and nb::bytes as const.

hawkinsp:


wjakob:
Could you update the documentation as well? (`api_core.rst`)

hawkinsp:
> Could you update the documentation as well? (`api_core.rst`)

Done.

wjakob:
Thanks!

Issue #286, opened 2023-08-31
Passing vector<complex<float>> to function argument


gillesdegottex:
Hi,

I struggle in passing a vector of complex values to a function.

For example, the following works as expected for `float` type:
```
#include <nanobind/nanobind.h>
#include <nanobind/stl/vector.h>
#include <iostream>

namespace nb = nanobind;

NB_MODULE(pybar, m) {
    m.def("foo", [](std::vector<float>& x){
        std::cout << "x[0]=" << x[0] << std::endl;
    });
}
```
leads to:
```
> python3 -c 'import pybar; import numpy as np; x = np.array([1],dtype=np.single); pybar.foo(x)'
x[0]=1
```

However, if using `std::complex<float>` type instead of `float`:
```
#include <nanobind/nanobind.h>
#include <nanobind/stl/vector.h>
#include <iostream>

namespace nb = nanobind;

NB_MODULE(pybar, m) {
    m.def("foo", [](std::vector<std::complex<float>>& x){
        std::cout << "x[0]=" << x[0] << std::endl;
    });
}
```
leads to:
```
> python3 -c 'import pybar; import numpy as np; x = np.array([1+1j],dtype=np.csingle); pybar.foo(x)'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: foo(): incompatible function arguments. The following argument types are supported:
    1. foo(arg: list[std::complex<float>], /) -> None

Invoked with types: ndarray
```

I did try to add an explicit `type_caster` in ` stl/vector.h` (defined before the generic `typecaster<Type>`):
```
template <typename Type, typename Alloc> struct type_caster<std::vector<std::complex<Type>, Alloc>>
 : list_caster<std::vector<std::complex<Type>, Alloc>, std::complex<Type>> { };
```
which leads to the exact same error as above.

Any idea how this should be done in nanobind?

(thanks to this new C++/Python binder!)

wjakob:
Nanobind currently lacks a type caster for `std::complex<T>``. Help on this front is appreciated. This is not an issue but a feature request, so I am closing the ticket.

gillesdegottex:
Ok!

A few questions:
* Where can I open a feature request? For follow up on this subject.
* To add support for std::complex, I guess I should add a `stl/complex.h`, and maybe dig into `bn_cast.h`, right? Am I in the proper direction?

wjakob:
[...]

PR #292, opened 2023-09-05
Add type caster for complex numbers

gillesdegottex:
This is an attempt to add support for complex numbers using a `type_caster` (previously discussed as #286 ).

With test code:
```
#include <nanobind/stl/vector.h>
#include <nanobind/stl/complex.h>

namespace nb = nanobind;

#include <iostream>

typedef double value_t;
// typedef float value_t;

std::complex<value_t> foo(const std::complex<value_t>& x){
    std::cout << "x=" << x << " sizeof(complex value)=" << 8*sizeof(x) << "b" << std::endl;
    return x;
}

std::vector<std::complex<value_t>> foo_array(const std::vector<std::complex<value_t>>& x){
    std::cout << "x[0]=" << x[0] << " sizeof(complex element)=" << 8*sizeof(x[0]) << "b" << std::endl;
    std::cout << "x[1]=" << x[1] << " sizeof(complex element)=" << 8*sizeof(x[1]) << "b" << std::endl;
    return x;
}

NB_MODULE(pybar, m) {
    m.def("foo", &foo);
    m.def("foo_array", &foo_array);
}
```

This works with test commands:
```
>> python3 -c 'import pybar; import numpy as np; x = 2.7-3.2j; print(pybar.foo(x))'
>> python3 -c 'import pybar; x = [2.7-3.2j, -1j]; print(pybar.foo_array(x))'
>> python3 -c 'import pybar; import numpy as np; x = np.array([2.7-3.2j, -1j]); print(pybar.foo_array(x))'
>> python3 -c 'import pybar; import numpy as np; x = np.array([2.7-3.2j, -1j],dtype=np.complex128); print(pybar.foo_array(x))'
```

However, it fails when numpy input is 32b float (i.e. np.complex64), ex:
```
>> python3 -c 'import pybar; import numpy as np; x = np.array([2.7-3.2j, -1j],dtype=np.complex64); print(pybar.foo_array(x))'
<string>:1: ComplexWarning: Casting complex values to real discards the imaginary part
[...]
```
This happens, no matter if I'm using `std::complex<float>` or `std::complex<double>` in test code.
My understanding is that a `PyComplex` seems to be built at some point assuming double size (because base python (not numpy) complex type is always double) and I can't find where this is built (I guess before call to `type_caster<std::complex<T1>>::from_python(.)` ).
If anybody has an idea about this issue, it is very much welcome!

Even though forcing to 64b float could be enough, this is clearly not optimal both in terms of efficiency and memory.

Elements to review:
- [ ] See copyright notice. I don't know how you handle this in nanobind.
- [x] Tests: Done.
- [x] `np.complex64` inputs, as mentioned above: Now supported.
- [x] Unnecessary use of `make_caster<T1>` I think. See TODO tag in file `stl/complex.h`: Cleaned.

### DIFF
--- a/include/nanobind/stl/complex.h
+++ b/include/nanobind/stl/complex.h
@@ -19,10 +19,7 @@ NAMESPACE_BEGIN(detail)
 
 template <typename T1> struct type_caster<std::complex<T1>> {
 
-    // Sub type caster
-    // TODO This definition seems useless, but I can't find a neat way to replace `Caster::Name` below.
-    using Caster = make_caster<T1>; 
-    NB_TYPE_CASTER(std::complex<T1>, const_name("complex [") + Caster::Name + const_name("]") )
+    NB_TYPE_CASTER(std::complex<T1>, const_name("complex") )
 
     bool from_python(handle src, uint8_t flags,
                      cleanup_list *cleanup) noexcept {
--- a/include/nanobind/stl/complex.h
+++ b/include/nanobind/stl/complex.h
@@ -12,8 +12,6 @@
 #include <nanobind/nanobind.h>
 #include <complex>
 
-#include <iostream>
-
 NAMESPACE_BEGIN(NB_NAMESPACE)
 NAMESPACE_BEGIN(detail)
 
@@ -36,6 +34,8 @@ template <typename T1> struct type_caster<std::complex<T1>> {
     template <typename T>
     static handle from_cpp(T &&value, rv_policy policy,
                            cleanup_list *cleanup) noexcept {
+        (void)policy;
+        (void)cleanup;
 
         // Conversion from 2xfloat to 2xdouble if needs be
         return PyComplex_FromDoubles(value.real(), value.imag());
--- a/include/nanobind/stl/complex.h
+++ b/include/nanobind/stl/complex.h
@@ -30,8 +30,6 @@ template <typename T1> struct type_caster<std::complex<T1>> {
         (void)cleanup;
 
         PyObject* obj = src.ptr();
-        // Real and Imaginary parts of a python complex object are always double:
-        //   https://docs.python.org/3/c-api/complex.html#c.Py_complex
         value.real(PyComplex_RealAsDouble(obj));
         value.imag(PyComplex_ImagAsDouble(obj));
 
@@ -43,9 +41,7 @@ template <typename T1> struct type_caster<std::complex<T1>> {
                            cleanup_list *cleanup) noexcept {
 
         // Conversion from 2xfloat to 2xdouble if needs be
-        PyObject* r = PyComplex_FromDoubles(value.real(), value.imag());
-
-        return r;
+        return PyComplex_FromDoubles(value.real(), value.imag());
     }
 };
 
--- a/include/nanobind/stl/complex.h
+++ b/include/nanobind/stl/complex.h
@@ -1,0 +1,53 @@
+/*
+    nanobind/stl/complex.h: type caster for std::complex<...>
+
+    Copyright (c) 2023 Degottex Gilles
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <nanobind/nanobind.h>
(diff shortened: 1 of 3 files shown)


include/nanobind/stl/complex.h
```diff
@@ -0,0 +1,53 @@
...
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+template <typename T1> struct type_caster<std::complex<T1>> {
+
+    // Sub type caster
+    // TODO This definition seems useless, but I can't find a neat way to replace `Caster::Name` below.
+    using Caster = make_caster<T1>; 
+    NB_TYPE_CASTER(std::complex<T1>, const_name("complex [") + Caster::Name + const_name("]") )
```
wjakob, line 25: Can remove ``Caster::Name`` here and just denote the type name as `"complex"` (anything else would be a pretty weird type signature in Python).

include/nanobind/stl/complex.h
```diff
@@ -0,0 +1,53 @@
...
+
+        return true;
+    }
+
+    template <typename T>
+    static handle from_cpp(T &&value, rv_policy policy,
+                           cleanup_list *cleanup) noexcept {
+
+        // Conversion from 2xfloat to 2xdouble if needs be
+        PyObject* r = PyComplex_FromDoubles(value.real(), value.imag());
```
wjakob, line 46: return directly

include/nanobind/stl/complex.h
```diff
@@ -0,0 +1,53 @@
...
+    using Caster = make_caster<T1>; 
+    NB_TYPE_CASTER(std::complex<T1>, const_name("complex [") + Caster::Name + const_name("]") )
+
+    bool from_python(handle src, uint8_t flags,
+                     cleanup_list *cleanup) noexcept {
+        (void)flags;
+        (void)cleanup;
+
+        PyObject* obj = src.ptr();
+        // Real and Imaginary parts of a python complex object are always double:
```
wjakob, line 33: simple enough not to need a comment.

wjakob (changes requested):
Could you add test coverage to the official test suite? (this should validate the casting in both directions for  `std::complex<float>` and `std:::complex<double>`).

gillesdegottex:
I'll see for coding the tests, it might take me a bit of time.
Should I get inspired by a specific test file? (otherwise `tests/test_ndarray.*` seems a descent starting point to me)
This would go in `tests/test_stl_complex.*`, correct?

include/nanobind/stl/complex.h
wjakob: Those type annotations are intended to make sense in Python, and there isn't a separate single precision float object.

include/nanobind/stl/complex.h
gillesdegottex: done

wjakob:
You can put tests into the existing STL type casters (it is part of that). I don't think it needs to be that complicated: show that you can pass a complex number to C++ and back, for single and double versions.

include/nanobind/stl/complex.h
gillesdegottex: done

include/nanobind/stl/complex.h
gillesdegottex: done

wjakob:
I'm checking in on the status of this PR -- will you be able to add test coverage @gillesdegottex ?

gillesdegottex:
Give me up to next Monday plz, and just close it if I can't do it by then.

gillesdegottex:
All done.
Side bonus: complex64 is also handled now. I suspect it could be faster though (see TODO).

The 9 failing tests do so because of numpy module not installed on the corresponding docker image.
I can add a condition to run only the numpy-related tests if the module is present.
I think it would be better to install the packages. Complex numbers are very likely to be used when numpy is also used.

wjakob:
I made a number of changes to this PR and merged it as commit dcbed4fe1500383ad1f4dff47cacbf0f2e6b1d3f.

gillesdegottex:
Great thx!

PR #299, opened 2023-09-18
support exec, eval

nschloe:
I took pybind11's `eval.h` plus tests and adapted it for nanobind. I had to remove `eval_file` as I couldn't find a stable-ABI equivalent for opening files (done with `_Py_fopen_obj` in pybind11).

@wjakob I believe this is ready for the review. The test failures seem unrelated.


include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+#include <nanobind/nanobind.h>
+
+#include <string>
+#include <utility>
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+
+/// \ingroup python_builtins
+/// Return a dictionary representing the global variables in the current execution frame,
+/// or ``__main__.__dict__`` if there is no frame (usually when the interpreter is embedded).
```
wjakob, line 25: nanobind doesn't support embedding, so the special case here can be removed.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+inline dict globals() {
+    PyObject *p = PyEval_GetGlobals();
+    return borrow<dict>(p ? p : module_::import_("__main__").attr("__dict__").ptr());
+}
+
+enum eval_mode {
+    /// Evaluate a string containing an isolated expression
+    eval_expr,
```
wjakob, line 33: let's set this equal to `Py_eval_input` (ditto for the other cases below). That way, the `switch` statement in the `eval()` implementation can disappear.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <nanobind/nanobind.h>
+
+#include <string>
```
wjakob, line 18: Let's remove the `string` header here, we can do without.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+object eval(const str &expr, dict global = globals(), object local = object()) {
+    if (!local) {
+        local = global;
+    }
+
+    /* PyRun_String does not accept a PyObject / encoding specifier,
+       this seems to be the only alternative */
+    std::string buffer = std::string("# -*- coding: utf-8 -*-\n") + std::string(expr.c_str());
+
+    int start = 0;
```
wjakob, line 52: Can be removed if enumeration entries equal to `Py_*` values.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+
+template <eval_mode mode = eval_expr>
+object eval(const str &expr, dict global = globals(), object local = object()) {
+    if (!local) {
+        local = global;
+    }
+
+    /* PyRun_String does not accept a PyObject / encoding specifier,
+       this seems to be the only alternative */
+    std::string buffer = std::string("# -*- coding: utf-8 -*-\n") + std::string(expr.c_str());
```
wjakob, line 50: I think that this whole line can just disappear, it's a Python 2.x thing. Python 3.x is UTF8 by default. So replace any use of `buffer` with `expr` below.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+            break;
+        default:
+            detail::fail("invalid evaluation mode");
+    }
+
+    // This used to be PyRun_String, but that function isn't in the stable ABI.
+    PyObject *codeobj = Py_CompileString(buffer.c_str(), "<string>", start);
+    if (!codeobj)
+    {
+        throw python_error();
```
wjakob, line 71: Instead of `throw python_error()`: call `detail::raise_python_error();`

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,97 @@
...
+    }
+    PyObject *result = PyEval_EvalCode(codeobj, global.ptr(), local.ptr());
+    if (!result) {
+        throw python_error();
```
wjakob, line 75: ditto

wjakob (changes requested):
This looks like a great start! Here is some feedback on this first version.

wjakob:
The CMake thing happens sometimes, unfortunately. Not sure what that is all about -- the next time you update the PR, it will likely succeed.

include/nanobind/nb_eval.h
nschloe: Sure? Because the script is using `std::string`. I'm a bit rusty on C++, but I faintly remember to "include what you use". Will go ahead with the removal once you confirm.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+    // Evaluate a string containing a single statement. Returns \c none
+    eval_single_statement = Py_single_input,
+
+    // Evaluate a string containing a sequence of statement. Returns \c none
+    eval_statements = Py_file_input
+};
+
+template <eval_mode start = eval_expr>
+object eval(const str &expr, object global = object(), object local = object()) {
+    if (!local) {
```
wjakob, line 35: Minor stylistic comment (applies to this line and a few others): throughout the codebase, the opening/closing brace for simple `if` statements with a 1-liner `then` branch is omitted.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
+/*
+    nanobind/nb_eval.h: Support for evaluating Python expressions and statements
+                        from strings and files
+
+    Adapted by Nico Schlömer from pybind11's eval.h with
+
+    Copyright (c) 2016 Klemens Morgenstern <klemens.morgenstern@ed-chemnitz.de> and
```
wjakob, line 7: At this point, the functionality is such a thin wrapper around stock Python functions that I don't think you need to include this extended copyright header. Maybe for the testcases part ("Testcases adapted from [..]"), which I now see you added already.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+};
+
+template <eval_mode start = eval_expr>
+object eval(const str &expr, object global = object(), object local = object()) {
+    if (!local) {
+        local = global;
+    }
+
+    // This used to be PyRun_String, but that function isn't in the stable ABI.
+    PyObject *codeobj = Py_CompileString(expr.c_str(), "<string>", start);
```
wjakob, line 40: This code looks to me like it will leak a reference to `codeobj` (is that a bug in the pybind11 version as well?!). 

Can you change it to `object codeobj = steal(Py_CompileString(expr.c_str(), "<string>", start));` and then access via `codeobj.ptr()` below?

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+    if (!codeobj)
+    {
+        detail::raise_python_error();
+    }
+    PyObject *result = PyEval_EvalCode(codeobj, global.ptr(), local.ptr());
+    if (!result) {
+        detail::raise_python_error();
+    }
+
+    return steal<object>(result);
```
wjakob, line 50: Can just write `steal(`, no need to specify `object` as template parameter.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+
+    // Evaluate a string containing a single statement. Returns \c none
+    eval_single_statement = Py_single_input,
+
+    // Evaluate a string containing a sequence of statement. Returns \c none
+    eval_statements = Py_file_input
+};
+
+template <eval_mode start = eval_expr>
+object eval(const str &expr, object global = object(), object local = object()) {
```
wjakob, line 34: Please change the interface so that `global` and `local` are accepted as `handle`. The caller already holds a reference, so taking objects as input is overkill.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+    PyObject *result = PyEval_EvalCode(codeobj, global.ptr(), local.ptr());
+    if (!result) {
+        detail::raise_python_error();
+    }
+
+    return steal<object>(result);
+}
+
+template <eval_mode start = eval_expr, size_t N>
+object eval(const char (&s)[N], object global = object(), object local = object()) {
```
wjakob, line 54: Analogous change for `object` -> `handle`. Can remove the `std::move` calls in the function.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+    // Support raw string literals by removing common leading whitespace
+    auto expr = (s[0] == '\n') ? str(module_::import_("textwrap").attr("dedent")(s)) : str(s);
+    return eval<start>(expr, std::move(global), std::move(local));
+}
+
+inline void exec(const str &expr, object global = object(), object local = object()) {
```
wjakob, line 60: Analogous change for `object` -> `handle`. Can remove the `std::move` calls in the function.

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,69 @@
...
+    eval<eval_statements>(expr, std::move(global), std::move(local));
+}
+
+template <size_t N>
+void exec(const char (&s)[N], object global = object(), object local = object()) {
```
wjakob, line 65: Analogous change for `object` -> `handle`. Can remove the `std::move` calls in the function.

wjakob (changes requested):
This looks like it is close to mergeable. One remaining thing that is missing is basic Sphinx documentation that could likely be repurposed from pybind11.

include/nanobind/nb_eval.h
nschloe: Instead of `Py_CompileString` and `PyEval_EvalCode`, pybind11 uses `PyRun_String` (not part of the stable ABI), so it's not affected by this.

include/nanobind/nb_eval.h
nschloe: I think it used to be `object` so you can pass in `None`. Will change to `handle`.

include/nanobind/nanobind.h
```diff
@@ -51,6 +51,7 @@
 #include "nb_func.h"
 #include "nb_class.h"
 #include "nb_misc.h"
+#include "nb_eval.h"
```
wjakob, line 54: Let's make this an optional feature of nanobind. Users will need to include the file `nanobind/eval.h` (so it needs to be renamed and not included from `include/nanobind/nanobind.h`).

include/nanobind/nb_eval.h
```diff
@@ -0,0 +1,63 @@
...
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <nanobind/nanobind.h>
+
+#include <utility>
```
wjakob, line 15: Are you using anything from `<utility>`?

wjakob:
One last thing I remembered: besides copying the docs from nanobind, can you add the function signatures and document then added enum? This should go to `docs/api_extra.rst`, which documents opt-in parts of the nanobind API.

include/nanobind/nb_eval.h
nschloe: Now that the `std::move`s are gone, not anymore.

nschloe:
I could actually make `eval_file` work as well. Let me try and have a look.

nschloe:
Let's forget about that; I'm getting entangled in segfaults here. To be added as a later step. :)

wjakob:
Ok. Ready for another review or potentially merge?

include/nanobind/eval.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+template <eval_mode start = eval_expr>
+object eval(const str &expr, handle global = handle(), handle local = handle()) {
+    if (!local)
+        local = global;
+
+    // This used to be PyRun_String, but that function isn't in the stable ABI.
+    object codeobj = steal(Py_CompileString(expr.c_str(), "<string>", start));
+    if (!codeobj)
```
wjakob, line 35: `codeobj.is_valid()`

include/nanobind/eval.h
```diff
@@ -0,0 +1,61 @@
...
+    if (!result)
+        detail::raise_python_error();
+
+    return steal(result);
+}
+
+template <eval_mode start = eval_expr, size_t N>
+object eval(const char (&s)[N], handle global = handle(), handle local = handle()) {
+    // Support raw string literals by removing common leading whitespace
+    auto expr = (s[0] == '\n') ? str(module_::import_("textwrap").attr("dedent")(s)) : str(s);
```
wjakob, line 48: `str expr` (even shorter than auto :D)

docs/utilities.rst
```diff
@@ -0,0 +1,58 @@
...
+        if x == 42:
+            print('Hello World!')
+        else:
+            print('Bye!')
+        )", scope
+    );
+
+.. note::
+
+    `eval` accepts a template parameter that describes how the string/file
```
wjakob, line 53: Can you add sphinx prefixes (e.g. `:py:func:..` to referenced function and enumerations values to ensure that those are linked to the right parts of the API documentation?

include/nanobind/eval.h
```diff
@@ -0,0 +1,61 @@
...
+    // Evaluate a string containing a single statement. Returns \c none
+    eval_single_statement = Py_single_input,
+
+    // Evaluate a string containing a sequence of statement. Returns \c none
+    eval_statements = Py_file_input
+};
+
+template <eval_mode start = eval_expr>
+object eval(const str &expr, handle global = handle(), handle local = handle()) {
+    if (!local)
```
wjakob, line 30: `if (!local.is_valid())`

tests/test_eval.cpp
```diff
@@ -0,0 +1,81 @@
+#include <nanobind/nanobind.h>
+#include <nanobind/eval.h>
+#include <nanobind/stl/pair.h>
+
+// #include "pybind11_tests.h"
```
wjakob, line 5: Remove commented include directives.

docs/api_extra.rst
```diff
@@ -904,3 +904,43 @@ converting to Python.
...
+.. code-block:: cpp
+
+   #include <nanobind/eval.h>
+
+Detailed documentation including example code is provided in a :ref:`separate
+section <eval>`.
+
+.. cpp:enum-class:: eval_mode
+
+   This enumeration characterizes the start symbol for different kinds of
```
wjakob, line 924: "start symbol" -- really (maybe in a Chomsky grammar kind of way)? This is likely confusing to users, I would say: "This enumeration specifies how the content of a string should be interpreted."

docs/api_extra.rst
```diff
@@ -904,3 +904,43 @@ converting to Python.
...
+   expressions. Used in Py_CompileString().
+
+   .. cpp:enumerator:: eval_expr = Py_eval_input
+
+      Evaluate a string containing an isolated expression
+
+   .. cpp:enumerator:: eval_single_statement = Py_single_input
+
+      Evaluate a string containing a single statement. Returns \c none
```
wjakob, line 933: `\c none` -> \`\`None\`\`

wjakob (commented):
Just a few minor remaining things I noticed.

nschloe:
> Ready for another review

Always ready! :) (Thanks for the good reviews.)

docs/api_extra.rst
nschloe: Took the expression from [here](https://docs.python.org/3/c-api/veryhigh.html#c.Py_eval_input). Will change as suggested.

docs/utilities.rst
wjakob: (I don't see the change for this resolved comment)

nschloe: I'd only changed it for the first occurence of eval/exec; fixed now for that line too.

wjakob:
Merged, thank you for your work!

nschloe:
Thanks again for walking me through the process!

wjakob:
You'll have to be a lot more patient ;). There is a huge issue that I want to fix in an upcoming release (#307), and which requires some stabilization until then. If you cannot wait, just reference a particular commit of this repository (supported by pip, requirements.txt, etc.)

PR #311, opened 2023-10-03
add nb::globals

nschloe:


include/nanobind/nb_globals.h
```diff
@@ -0,0 +1,19 @@
...
+    Copyright (c) 2022 Wenzel Jakob
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+NAMESPACE_BEGIN(NB_NAMESPACE)
+
+/// \ingroup python_builtins
+/// Return a dictionary representing the global variables in the current execution frame,
+/// or ``__main__.__dict__`` if there is no frame (usually when the interpreter is embedded).
```
wjakob, line 13: nanobind does not support embedding, so the alternative implementation seems superfluous. Can you add this to one of the main header files instead of creating such a small one for just one function? (e.g. `nb_types.h`)

wjakob (changes requested):
Looks good, I have two change requests.

nschloe:
The test failure is unrelated I believe.

include/nanobind/nb_misc.h
```diff
@@ -39,4 +39,12 @@ inline void set_implicit_cast_warnings(bool value) noexcept {
     detail::set_implicit_cast_warnings(value);
 }
 
+/// \ingroup python_builtins
+/// Return a dictionary representing the global variables in the current execution frame,
+/// or ``__main__.__dict__`` if there is no frame.
+inline dict globals() {
+    PyObject *p = PyEval_GetGlobals();
```
wjakob, line 46: isn't a simple: `return borrow<dict>(PyEval_GetGlobals());` enough? My understanding is that this will always work when nanobind code runs as an extension module, which is the case in nanobind projects.

nschloe: No idea. `PyEval_GetGlobals` _can_ return `NULL`, so I guess we should handle this case somehow? Perhaps throwing an error would be more appropriate.

wjakob: Yes, the way to handle it in that case is via ``nb::detail::raise_python_error()``.

include/nanobind/nb_misc.h
```diff
@@ -39,4 +39,14 @@ inline void set_implicit_cast_warnings(bool value) noexcept {
     detail::set_implicit_cast_warnings(value);
 }
 
+/// \ingroup python_builtins
+/// Return a dictionary representing the global variables in the current execution frame,
```
wjakob, line 43: That whole comment is now out of sync and can simply be deleted. (FWIW nanobind also don't use the `\ingroup` syntax since the API docs are separately built using Sphinx.

wjakob:
I made a small change: `nb::detail::raise_python_error()` did not apply since `PyEval_GetGlobals` does not set a Python error status when it returns `NULL`..

wjakob:
(also, this is a new feature rather than a bugfix, so it will go into the next minor release, not the patch one.)

PR #317, opened 2023-10-09
nullptr from cpp

huangweiwu:


include/nanobind/nb_cast.h
```diff
@@ -218,6 +218,12 @@ template <> struct type_caster<char> {
 
     static handle from_cpp(const char *value, rv_policy,
                            cleanup_list *) noexcept {
+        if (value == nullptr)
```
wjakob, line 221: All good, except can you put the opening brace on the previous line?

huangweiwu: Done.

PR #318, opened 2023-10-09
Fix None check in variant caster

yosh-matsuda:
This PR fixes the missing `None` check in the `std::variant` caster. 

The following code produces SIGABRT at runtime since the change in 1220156961ce2d0c96a525f3c27b88e824b997ce;

```c++
m.def("variant_copyable_none", [](std::variant<int, Copyable, std::monostate> &) {}, nb::arg("x").none());
```

```python
t.variant_copyable_none(None)
```

This PR include

* Improve test coverage
* Revert the `None` check
* Refactoring of redundant code

### DIFF
--- a/include/nanobind/stl/variant.h
+++ b/include/nanobind/stl/variant.h
@@ -28,9 +28,7 @@ template <> struct type_caster<std::monostate> {
     NB_TYPE_CASTER(std::monostate, const_name("None"));
 
     bool from_python(handle src, uint8_t, cleanup_list *) noexcept {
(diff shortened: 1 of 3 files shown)


wjakob:
Thank you!

wjakob:
This was kind of a serious issue that also needed to be fixed elsewhere. See commit [5f25ae0](https://github.com/wjakob/nanobind/commit/5f25ae0eb9691fbe03a20bcb9f604277ccc1884b).

PR #325, opened 2023-10-17
Made type_caster<Eigen::Map<T>> correctly consider T's constness

WKarel:
There is a flaw in type_caster<Eigen::Map<T>>: instead of testing T's constness, it tests the constness of Map itself, which can never be const. 
The writability of numpy.ndarray's returned for Eigen::Map is not tested yet, and so I have added respective tests.


wjakob:
Thank you @WKarel !

PR #338, opened 2023-10-24
Minor fixes and additions for complex `ndarray`

garth-wells:
Consistency fixes for complex ndarray and remove duplicated trait code.

Came across these issues when hunting down a problem passing read-only complex Numpy arrays to C++ (real-valued, read-only arrays work fine), which appears to be an issue with the Numpy  `__dlpack__` function, which fails for read-only complex.

### DIFF
--- a/include/nanobind/ndarray.h
+++ b/include/nanobind/ndarray.h
@@ -13,6 +13,7 @@
 #pragma once
 
 #include <nanobind/nanobind.h>
+#include <nanobind/stl/complex.h>
 #include <initializer_list>
 
 NAMESPACE_BEGIN(NB_NAMESPACE)
@@ -65,13 +66,6 @@ struct dltensor {
 
 NAMESPACE_END(dlpack)
 
-NAMESPACE_BEGIN(detail)
-
-template <typename T>
-struct is_complex : public std::false_type { };
-
-NAMESPACE_END(detail)
-
 constexpr size_t any = (size_t) -1;
 
 template <size_t... Is> struct shape {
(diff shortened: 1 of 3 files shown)


include/nanobind/ndarray.h
```diff
@@ -13,6 +13,7 @@
 #pragma once
 
 #include <nanobind/nanobind.h>
+#include <nanobind/stl/complex.h>
```
wjakob, line 16: This undoes an important change introduced in the previous review of the ndarray headers: please, please don't include `complex` into all compilation units that require NDarray. This pulls in a huge amount of STL header code. You will need to make changes using partial template overloads for complex arrays, as was done before.

wjakob (changes requested):

src/nb_ndarray.cpp
```diff
@@ -211,6 +211,9 @@ static PyObject *dlpack_from_buffer_protocol(PyObject *o, bool ro) {
             case 'f':
             case 'd': dt.code = (uint8_t) dlpack::dtype_code::Float; break;
 
+            // case 'Zf':
```
garth-wells, line 214: This needs looking at since for complex the id isn't a single `char`.

include/nanobind/ndarray.h
garth-wells: Fixed.

When making this fix I noticed https://github.com/wjakob/nanobind/pull/338#discussion_r1369836868.

src/nb_ndarray.cpp
wjakob: Before line 191, add:

```cpp
bool is_complex = format_str[0] == 'Z';
if (is_complex)
    format_str++;
```
Then remove the two 'Z' cases and below the switch statement, add:

```cpp
if (is_complex) {
    fail =  dt.code != (uint8_t) dlpack::dtype_code::Float;
    dt.code = dlpack::dtype_code::Complex;
}
```

wjakob:
Is this good to merge now? Does it address the problems you discovered when testing with fenics?

garth-wells:
> Is this good to merge now? Does it address the problems you discovered when testing with fenics?

Yes, I think it's good to merge.

No, it didn't fix the problem I'm having with read-only Numpy complex arrays. The line
```
capsule = steal(PyObject_CallMethod(o, "__dlpack__", nullptr));
```
doesn't return a valid object for const `std::complex`.  I need to dig deeper. Does `nanobind` cast const away somewhere to avoid Numpy `__dlpack__` errors. I can dig deeper later.

garth-wells:
See #342 with failing example.

Issue #342, opened 2023-10-25
[BUG]: `ndarray` cast fails for read-only complex Numpy arrays


garth-wells:
### Problem description

Passing a read-only complex Numpy array to a function expecting a `nb::ndarray<const std::complex<double>>` leads to a `TypeError`. 

It works fine for reals. See https://github.com/garth-wells/nanobind_example.

### Reproducible example code

https://github.com/garth-wells/nanobind_example/blob/master/src/nanobind_example_ext.cpp
```cpp  
#include <nanobind/nanobind.h>
#include <nanobind/ndarray.h>
#include <nanobind/stl/complex.h>
#include <complex>
#include <iostream>

namespace nb = nanobind;

NB_MODULE(nanobind_example_ext, m)
{
    m.def("array_d_const", [](nb::ndarray<const double> x)
          { std::cout << "Const double array" << std::endl; });
    m.def("array_c_const", [](nb::ndarray<const std::complex<double>> x)
          { std::cout << "Const complex array" << std::endl; });
}
```

https://github.com/garth-wells/nanobind_example/blob/master/bug.py
```python
import numpy as np
from nanobind_example import array_c_const, array_d_const


x = np.array([1, 2], dtype=np.float64)
array_d_const(x)  # works
x.flags.writeable = False
array_d_const(x)  # works

c = np.array([1, 2], dtype=np.complex128)
array_c_const(c)  # works
c.flags.writeable = False
array_c_const(c)  # fails with TypeError
```

garth-wells:
In case it's easier, see PR #343 with failing test.

wjakob:
Fixed by ba8c7fa55f2d0ad748cad1dd4af2b22979ebc46a

PR #343, opened 2023-10-25
Add failing Numpy read-only test to test suite

garth-wells:
Following #https://github.com/wjakob/nanobind/pull/338#issuecomment-1779217448, tests added that fail with read-only complex Numpy array.

See #342.


wjakob:
Should be resolved now.

Issue #345, opened 2023-10-27
[BUG]: Nanobind throws various warnings when building with `-Werror -Wall`


jorgensd:
### Problem description

It is not possible to build nanobind with -Werror (or any software depending on nanobind) which unfortunate as it makes it harder to spot warnings in other packages

[log_nanobind.txt](https://github.com/wjakob/nanobind/files/13186018/log_nanobind.txt)

### Reproducible example code

```text
FROM ubuntu:22.04

ENV DEBIAN_FRONTEND=noninteractive
ARG TARGET_PLATFORM
RUN apt-get update && \
    apt-get install -y software-properties-common
RUN apt-get install -y git \
    cmake \
    gcc \
    g++ \
    ninja-build
RUN apt-get install -y python3-pip
RUN python3 -m pip install -U setuptools pip
ENV CXX_FLAGS "-Werror -Wall"
WORKDIR /src
RUN git clone --recursive https://github.com/wjakob/nanobind
RUN cd nanobind && \
    cmake -G Ninja -B build-dir -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DCMAKE_BUILD_TYPE=Debug -S . && \
    cmake --build build-dir && \ 
    cmake --install build-dir

RUN cd nanobind && \
    python3 -m pip install .
```

jorgensd:
The first issue that comes out of this is due to: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94428
i.e. using `-Werror` with gcc 11.4.0 yields:
```bash
In file included from /root/shared/include/nanobind/nanobind.h:48,
                 from /root/shared/src/common.cpp:10:
/root/shared/include/nanobind/nb_attr.h:202:32: error: unknown option after '#pragma GCC diagnostic' kind [-Werror=pragmas]
  202 | #pragma GCC diagnostic ignored "-Wzero-length-array"
      |                                ^~~~~~~~~~~~~~~~~~~~~
/root/shared/include/nanobind/nb_attr.h:202:32: note: did you mean '-Wzero-length-bounds'?
```
making it tricky to use GCC with nanobind and strict compile options.

francesco-ballarin:
The fix for https://github.com/wjakob/nanobind/pull/336 will probably fix this one too

jorgensd:
> The fix for #336 will probably fix this one too

Yes. it Seems like one has to use fairly recent versions of gcc to have a replacement for the flag in question. Problem is what to do with the older versions of the compiler.

PR #347, opened 2023-10-29
Fix for zero length array warning and gcc

garth-wells:
GCC doesn't have `"-Wzero-length-array"` (unlike clang) [1], which caused #345. This change allows `-Werror -pedantic` builds with GCC and clang.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94428


wjakob:
Sorry, I really don't like this change.

PR #350, opened 2023-11-01
Add type_caster for `std::nullopt`

zhqrbitee:
Currently if any code uses `std::nullopt`, it will throw a `bast_cast` exception as no type_caster has been defined for it: https://github.com/wjakob/nanobind/discussions/274. 
As `std::nullopt` is a standard way to assign empty value to `std::optional`, so I think it is better to add the type_caster for it in `stl/optional.h`


wjakob:
Thanks!

PR #351, opened 2023-11-01
For hashing type name, using `std::hash<std::string_view>` instead of `std::hash<std::string>` 

zhqrbitee:
Switch from `std::hash<string>` to `std::hash<string_view>` to avoid creating new std::string (which could involve memory allocation and copy) when doing hash for type name.


wjakob:
Nice!

PR #356, opened 2023-11-07
cmake: Add install rules and option to use submodule dependencies

ekilmer:
Package ecosystems that know how to install CMake projects can now package nanobind for distribution without using submodule'd dependencies.

nanobind does not install any compiled library artifacts. Instead, all necessary source files and supporting CMake files are installed to build nanobind within another CMake project.

A CMake user can now adjust the CMake option `NB_USE_SUBMODULE_DEPS` to choose whether nanobind will use and/or install the submodule'd external dependencies. This is `ON` by default, allowing nanobind's CMake logic to continue operating as it did previously.

Another CMake option was introduced, `NB_CREATE_INSTALL_RULES`, to allow users to choose whether to enable or disable the installation rules. Notably, if nanobind is added to a CMake project with `add_subdirectory`, the installation rules are disabled, because installing source files probably isn't the expected result of the parent project(s). Install-rule generation still respects the standard CMake `CMAKE_SKIP_INSTALL_RULES` variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ else()
 endif()
 
 option(NB_CREATE_INSTALL_RULES "Create installation rules" ${NB_MASTER_PROJECT})
-option(NB_USE_SUBMODULE_DEPS   "Use submodule'd dependencies" ON)
+option(NB_USE_SUBMODULE_DEPS   "Use the nanobind dependencies shipped as a git submodule of this repository" ON)
 
 option(NB_TEST              "Compile nanobind tests?" ${NB_MASTER_PROJECT})
 option(NB_TEST_STABLE_ABI   "Test the stable ABI interface?" OFF)
@@ -72,9 +72,7 @@ if(NB_CREATE_INSTALL_RULES AND NOT CMAKE_SKIP_INSTALL_RULES)
       DESTINATION "${NB_INSTALL_EXTDIR}/robin_map/include"
     )
     install(
-      FILES
-        "ext/robin_map/tsl-robin-map.natvis"
-        "ext/robin_map/CMakeLists.txt"
+      FILES ext/robin_map/CMakeLists.txt
       DESTINATION "${NB_INSTALL_EXTDIR}/robin_map"
     )
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,15 +44,16 @@ endif()
 # ---------------------------------------------------------------------------
 
 if(NB_CREATE_INSTALL_RULES AND NOT CMAKE_SKIP_INSTALL_RULES)
-  set(CMAKE_INSTALL_DATAROOTDIR "share"
-    CACHE PATH "Installation path for read-only architecture-independent data root")
+  include(GNUInstallDirs)
+  set(NB_INSTALL_DATADIR "${CMAKE_INSTALL_DATADIR}/nanobind"
+    CACHE PATH "Installation path for read-only architecture-independent nanobind data files")
 
   # Normally these would be configurable by the user, but we can't allow that
   # because the lookup paths are hard-coded in 'cmake/nanobind-config.cmake'
-  set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/include")
-  set(NB_INSTALL_SRCDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/src")
-  set(NB_INSTALL_EXTDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/ext")
-  set(NB_INSTALL_CMAKEDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/cmake")
+  set(CMAKE_INSTALL_INCLUDEDIR "${NB_INSTALL_DATADIR}/include")
+  set(NB_INSTALL_SRCDIR "${NB_INSTALL_DATADIR}/src")
+  set(NB_INSTALL_EXTDIR "${NB_INSTALL_DATADIR}/ext")
+  set(NB_INSTALL_CMAKEDIR "${NB_INSTALL_DATADIR}/cmake")
 
   install(
     DIRECTORY include/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,9 @@ else()
   set(NB_MASTER_PROJECT OFF)
 endif()
 
+option(NB_CREATE_INSTALL_RULES "Create installation rules" ${NB_MASTER_PROJECT})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -30,12 +33,57 @@ endif()
...
+if(NB_CREATE_INSTALL_RULES AND NOT CMAKE_SKIP_INSTALL_RULES)
+  set(CMAKE_INSTALL_DATAROOTDIR "share"
+    CACHE PATH "Installation path for read-only architecture-independent data root")
+
+  # Normally these would be configurable by the user, but we can't allow that
+  # because the lookup paths are hard-coded in 'cmake/nanobind-config.cmake'
+  set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/include")
+  set(NB_INSTALL_SRCDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/src")
+  set(NB_INSTALL_EXTDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/ext")
+  set(NB_INSTALL_CMAKEDIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind/cmake")
```
ekilmer, line 55: I guess these could be refactored a bit more to introduce `NB_INSTALL_DATADIR` as the base for these variables and let the user decide where that goes (in case they want to have multiple versions of nanobind `-DNB_INSTALL_DATADIR=share/nanobind-1.8.0`) Thoughts?

```cmake
  set(NB_INSTALL_DATADIR "${CMAKE_INSTALL_DATAROOTDIR}/nanobind"
    CACHE PATH "Installation path for read-only architecture-independent nanobind data files")
    
  # Normally these would be configurable by the user, but we can't allow that
  # because the lookup paths are hard-coded in 'cmake/nanobind-config.cmake'
  set(CMAKE_INSTALL_INCLUDEDIR "${NB_INSTALL_DATADIR}/include")
  set(NB_INSTALL_SRCDIR "${NB_INSTALL_DATADIR}/src")
  set(NB_INSTALL_EXTDIR "${NB_INSTALL_DATADIR}/ext")
  set(NB_INSTALL_CMAKEDIR "${NB_INSTALL_DATADIR}/cmake")
```

ekilmer: Added a new commit to include this and also use CMake's [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)

CMakeLists.txt
```diff
@@ -30,12 +33,58 @@ endif()
...
+  )
+
+  if(NB_USE_SUBMODULE_DEPS)
+    install(
+      DIRECTORY ext/robin_map/include/
+      DESTINATION "${NB_INSTALL_EXTDIR}/robin_map/include"
+    )
+    install(
+      FILES
+        "ext/robin_map/tsl-robin-map.natvis"
```
wjakob, line 76: The natvis part seems overkill given that this is an implementation detail dependency of nanobind.

CMakeLists.txt
```diff
@@ -11,6 +11,9 @@ else()
   set(NB_MASTER_PROJECT OFF)
 endif()
 
+option(NB_CREATE_INSTALL_RULES "Create installation rules" ${NB_MASTER_PROJECT})
+option(NB_USE_SUBMODULE_DEPS   "Use submodule'd dependencies" ON)
```
wjakob, line 15: "Use submodule'd dependencies" -> `Use the nanobind dependencies shipped as a git submodule of this repository`

cmake/nanobind-config.cmake
```diff
@@ -202,8 +202,20 @@ function (nanobind_build_library TARGET_NAME)
...
-    ${NB_DIR}/ext/robin_map/include)
+  # If nanobind was installed without submodule dependencies, then the
+  # dependencies directory won't exist and we need to find them.
+  # However, if the directory _does_ exist, then the user is free to choose
+  # whether nanobind uses them (based on `NB_USE_SUBMODULE_DEPS`), with a
+  # preference to choose them if `NB_USE_SUBMODULE_DEPS` is not defined
+  if (NOT IS_DIRECTORY ${NB_DIR}/ext/robin_map/include OR
+      (DEFINED NB_USE_SUBMODULE_DEPS AND NOT NB_USE_SUBMODULE_DEPS))
+    include(CMakeFindDependencyMacro)
+    find_dependency(tsl-robin-map)
```
wjakob, line 213: The `tsl-robin-map` package and `tsl::robin_map` target show up. This, I presume, is an alternative package you have created in the vcpkg ecosystem?

I'm wondering here if there is a way to make this more general. Perhaps, instead of `NB_USE_SUBMODULE_DEPS `, we can let the user specify a target name? (e.g. `NB_TSL_ROBIN_MAP`)

wjakob (commented):
That smaller set of changes looks much more reasonable, thanks! A few comments from me.

CMakeLists.txt
ekilmer: Agreed. Not sure why I included this in the first place.

cmake/nanobind-config.cmake
ekilmer: The [installation directions](https://github.com/Tessil/robin-map#installation) for robin-map indicate that CMake users can find it in their project with `find_package(tsl-robin-map)`, after it has been installed, and then link against the exported `tsl::robin_map` target. vcpkg has packaged this project, but so have many [other projects](https://repology.org/project/robin-map).

> I'm wondering here if there is a way to make this more general. Perhaps, instead of `NB_USE_SUBMODULE_DEPS`, we can let the user specify a target name? (e.g. `NB_TSL_ROBIN_MAP`)

Sorry, I don't follow what you mean by this. What would this code look like instead?

CMakeLists.txt
ekilmer: Fixed. Not sure how/if you want to format that line, now that it's quite a bit longer than the rest of the other CMake lines. It looks like ~80 columns is the style.

cmake/nanobind-config.cmake
Ahajha: Random comment from a lurker: Conan's tsl-robin-map also uses this name for find_package and its target:
```cmake
find_package(tsl-robin-map REQUIRED CONFIG)
target_link_libraries(${PROJECT_NAME} PRIVATE tsl::robin_map)
```

wjakob: Ok, I was just confused. I did not realize that `tsl::robin_map` has an official CMake installation mechanism. I had been wondering: why is *this* the right target name to use, but it makes sense now.

CMakeLists.txt
wjakob: no big deal.

wjakob:
Looks all good to me now, thank you!

PR #358, opened 2023-11-09
type_caster<std::optional>: support non-assignable types

WKarel:
So far, type_caster<std::optional<T>>::from_python requires T to support copy-assignment. Removed this requirement by replacing std::optional<T>::operator= with std::optional<T>::emplace, which always uses in-place construction.

There should be no other change in behaviour, because std::optional<T>::operator= uses
- in-place construction if empty, or
- copy-assignment if non-empty,
and since type_caster<std::optional<T>>::from_python is called only once on each instance, its member `value` is always empty, anyway.


wjakob:
Nice, thank you!

PR #359, opened 2023-11-09
Minimize required CMake dependencies when only installing

ekilmer:
For packagers, these changes fix an issue caused by the `find_package(Python)` call that cannot be avoided without patching. This call places a requirement on Python during CMake configuration, even though it's not a requirement for building and installing nanobind. While this point is subtle, it's ideal to install/package nanobind without requiring unnecessary external dependencies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.15...3.27)
-project(nanobind LANGUAGES CXX)
+project(nanobind LANGUAGES NONE)
 
 # ---------------------------------------------------------------------------
 # Only build tests by default if this is the top-level CMake project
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ endif()
 # ---------------------------------------------------------------------------
 
 if(NB_CREATE_INSTALL_RULES AND NOT CMAKE_SKIP_INSTALL_RULES)
+  # Silence warning in GNUInstallDirs due to no enabled languages
+  set(CMAKE_INSTALL_LIBDIR "")
```
wjakob, line 48: Can you explain this line a bit more? The effect of this is not obvious to me.

ekilmer: The comment above is attached to this (so that should be changed to provide more clarity), but if you remove this line, you will see a warning from the `GNUInstallDirs` CMake module.
```sh
$ cmake -B build -S . -DNB_TEST=OFF
-- nanobind: setting build type to 'Release' as none was specified.
CMake Warning (dev) at /opt/homebrew/Cellar/cmake/3.27.7/share/cmake/Modules/GNUInstallDirs.cmake:243 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:48 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring done (0.0s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/ekilmer/src/nanobind/build
```
Setting `CMAKE_INSTALL_LIBDIR` is, unfortunately, necessary here to silence the warning.

ekilmer: The warning comes now because I've set `LANGUAGES NONE` in the `project` line.

wjakob: Aha! Got it, thanks!

PR #360, opened 2023-11-09
Use provided CMake Python targets instead of variables

ekilmer:
When available, usage of CMake targets to propogate information like include directories, compilation and linker flags, etc. is more idiomatic than usage of variables.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -189,12 +189,10 @@ function (nanobind_build_library TARGET_NAME)
     target_compile_options(${TARGET_NAME} PRIVATE -fno-strict-aliasing)
   endif()
 
-  if (WIN32)
-    if (${TARGET_NAME} MATCHES "abi3")
-      target_link_libraries(${TARGET_NAME} PUBLIC Python::SABIModule)
-    else()
-      target_link_libraries(${TARGET_NAME} PUBLIC Python::Module)
-    endif()
+  if (${TARGET_NAME} MATCHES "abi3")
+    target_link_libraries(${TARGET_NAME} PUBLIC Python::SABIModule)
+  else()
+    target_link_libraries(${TARGET_NAME} PUBLIC Python::Module)
   endif()
 
   # Nanobind performs many assertion checks -- detailed error messages aren't
@@ -217,9 +215,7 @@ function (nanobind_build_library TARGET_NAME)
       ${NB_DIR}/ext/robin_map/include)
   endif()
 
-  target_include_directories(${TARGET_NAME} PUBLIC
-    ${Python_INCLUDE_DIRS}
-    ${NB_DIR}/include)
+  target_include_directories(${TARGET_NAME} PUBLIC ${NB_DIR}/include)
 
   target_compile_features(${TARGET_NAME} PUBLIC cxx_std_17)
   nanobind_set_visibility(${TARGET_NAME})


cmake/nanobind-config.cmake
```diff
@@ -189,12 +189,10 @@ function (nanobind_build_library TARGET_NAME)
     target_compile_options(${TARGET_NAME} PRIVATE -fno-strict-aliasing)
   endif()
 
-  if (WIN32)
-    if (${TARGET_NAME} MATCHES "abi3")
-      target_link_libraries(${TARGET_NAME} PUBLIC Python::SABIModule)
-    else()
-      target_link_libraries(${TARGET_NAME} PUBLIC Python::Module)
-    endif()
+  if (${TARGET_NAME} MATCHES "abi3")
```
wjakob, line 192: This seems dangerous. On macOS and Linux, we never want to link against Python's shared library. IIRC those targets caused such linking behavior. at least some CMake versions ago.

ekilmer: Ensuring we don't link against the Python shared library is a valid concern, so in addition to looking through CMake's source code, I've also performed experiments with the minimum and latest CMake versions to confirm that none of the created test modules link against Python's shared library when using `Python::Module` on Linux and macOS.

For CMake source code, [v3.15.7 of the `Modules/FindPython/Support.cmake`](https://github.com/Kitware/CMake/blob/v3.15.7/Modules/FindPython/Support.cmake#L1752-L1775) file shows that the `Python::Module` library is an `INTERFACE` library target on non-Windows (meaning only compiler flags are carried, and no physical library location is added). In the latest CMake version [v3.27.7 of the `Modules/FindPython/Support.cmake`](https://github.com/Kitware/CMake/blob/v3.27.7/Modules/FindPython/Support.cmake#L3925-L3967), we can also see similar code showing creation of an `INTERFACE` library target on non-windows machines (this library creation has been turned into a macro, and it's called with `Python::Module` and `Python::SABIModule` in the latter lines.

To check experimentally, I downloaded CMake v3.15.7 (for both Linux and macOS) and ran (for both CMake v3.15.7 and latest v3.27.7)
```sh
cmake -B build -S . -G Ninja
```
on both `master` and this branch, making sure to save off the resulting `build/build.ninja` files for comparison later (naming them something like `build-master-3.15.7.ninja`, `build-module-3.15.7.ninja`, etc.). I use the same build directory to eliminate differences in build directory paths. The only differences I saw were some additional link flags (`-Xlinker -undefined -Xlinker dynamic_lookup` on macOS) and slightly modified include path for Python (`-isystem /usr/include/python3.8` now, instead of `-I /usr/include/python3.8`).

---------------------

I believe the only way to link against Python's shared library is to use the `Python::Python` target, which is created when finding the [`Development.Embed` component](https://cmake.org/cmake/help/latest/module/FindPython.html#imported-targets).

wjakob: Right, I remembered there was something. It's the `-Xlinker -undefined -Xlinker dynamic_lookup`. We don't want that. See 2f29ec7d5fbebd5f55fb52da297c8d197279f659 and the huge discussion on the Python issue tracker linked from that commit.

ekilmer: Oh wow. TIL! Thank you for that link. I will close this PR.

PR #372, opened 2023-11-25
type_caster<ref>::from_cpp: use ref->self_py

WKarel:
`type_caster<ref>::from_cpp` missed an optimization so far: if no copy shall be made and `intrusive_base::self_py` returns a valid pointer, then return it directly and spare the time to lookup the instance map.

### DIFF
--- a/include/nanobind/intrusive/ref.h
+++ b/include/nanobind/intrusive/ref.h
@@ -136,6 +136,11 @@ template <typename T> struct type_caster<nanobind::ref<T>> {
 
     static handle from_cpp(const ref<T> &value, rv_policy policy,
                            cleanup_list *cleanup) noexcept {
+        if constexpr (std::is_base_of_v<intrusive_base, T>)
+            if (policy != rv_policy::copy && value.get())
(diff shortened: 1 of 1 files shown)


wjakob:
Can this be tested on the Python end? (e.g. with a ... ` is `.. assertion)

wjakob:
(never mind my earlier comment, I had completely misunderstood the context)

include/nanobind/intrusive/ref.h
```diff
@@ -136,6 +136,11 @@ template <typename T> struct type_caster<nanobind::ref<T>> {
 
     static handle from_cpp(const ref<T> &value, rv_policy policy,
                            cleanup_list *cleanup) noexcept {
+        if constexpr (std::is_base_of_v<intrusive_base, T>)
+            if (policy != rv_policy::copy && value.get())
```
wjakob, line 140: I think that we might want to exclude `rv_policy::move` as well

WKarel: I've done so.

wjakob:
Thanks!

PR #382, opened 2023-12-12
Add support for bool

matthewdcong:
Ported over from the corresponding pybind11 implementation

### DIFF
--- a/include/nanobind/nb_types.h
+++ b/include/nanobind/nb_types.h
@@ -350,7 +350,7 @@ class bool_ : public object {
         : object(value ? Py_True : Py_False, detail::borrow_t{}) { }
 
     explicit operator bool() const {
-        return (m_ptr != nullptr) && PyLong_AsLong(m_ptr) != 0;
+        return m_ptr == Py_True;
     }
 };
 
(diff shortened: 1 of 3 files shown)


include/nanobind/nb_types.h
```diff
@@ -340,6 +340,20 @@ class capsule : public object {
...
+    NB_OBJECT_DEFAULT(bool_, object, "bool", PyBool_Check)
+
+    explicit bool_(handle h)
+        : object(detail::bool_from_obj(h.ptr()), detail::borrow_t{}) { }
+
+    explicit bool_(bool value)
+        : object(value ? Py_True : Py_False, detail::borrow_t{}) { }
+
+    explicit operator bool() const {
+        return (m_ptr != nullptr) && PyLong_AsLong(m_ptr) != 0;
```
wjakob, line 353: I believe this could be implemented by a comparison against ``Py_True``.

matthewdcong: Do you mean something along the lines of `return m_ptr == Py_True;`?

matthewdcong: It works on my end. Changed, thanks for the suggestion.

PR #383, opened 2023-12-13
Fix incompatible MSVC options being passed through to nvcc

matthewdcong:
nanobind adds `/bigobj and /MP` as compiler options for MSVC which results in this issue previously seen in pybind11.

https://github.com/pybind/pybind11/issues/3883

The same fix has been ported over in this PR.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -231,9 +231,9 @@ endfunction()
 
 function(nanobind_opt_size name)
   if (MSVC)
-    target_compile_options(${name} PRIVATE $<${NB_OPT_SIZE}:/Os>)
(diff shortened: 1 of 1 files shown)


wjakob:
Should something similar be done related to size optimizations? I remember vaguely that nvcc doesn't like `-Os`

matthewdcong:
The current workaround for that is passing `NOMINSIZE` to `nanobind_add_module`. However, that applies to the whole module as opposed to individual translation units so it really depends if we want the finer grained control.

wjakob:
Could you modify the patch so that the default (pass `-Os` unless NOMINSIZE is specified) is handled differently so that nvcc works out of the box? Then this PR would be complete in the sense that it handles all nvcc-related oddities.

matthewdcong:
Sure, sounds good

wjakob:
Thanks!

PR #386, opened 2023-12-14
Avoid a crash in nb::try_cast<T>(nb::none())

oremanj:
Previously it could raise an exception from the `type_caster`'s conversion operator, which would then crash the program due to `try_cast` being `noexcept`.

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -396,7 +396,7 @@ bool try_cast(const detail::api<Derived> &value, T &out, bool convert = true) no
         try {
             out = caster.operator detail::cast_t<T>();
             return true;
-        } catch (...) {
+        } catch (const builtin_exception&) {
             return false;
         }
     }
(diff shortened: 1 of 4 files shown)


include/nanobind/nb_cast.h
```diff
@@ -393,8 +393,12 @@ bool try_cast(const detail::api<Derived> &value, T &out, bool convert = true) no
     if (caster.from_python(value.derived().ptr(),
                            convert ? (uint8_t) detail::cast_flags::convert
                                    : (uint8_t) 0, nullptr)) {
-        out = caster.operator detail::cast_t<T>();
-        return true;
+        try {
+            out = caster.operator detail::cast_t<T>();
+            return true;
+        } catch (...) {
```
wjakob, line 399: Can you be more specific in catching a particular exception here? I think it should be `nb::next_overload`.

oremanj: Sure. It's `nb::builtin_exception` since all the more-specific choices turned into wrappers around that one.

wjakob:
Thanks!

PR #390, opened 2023-12-31
Add built-in constant types

yosh-matsuda:
I would like to propose adding [built-in constant](https://docs.python.org/3/library/constants.html) types in this PR.

This enables to write literal-based overloads like:

```python
@overload
func(arg: Literal[True], /) -> Literal[True]: ...
@overload
func(arg: Literal[False], /) -> Literal[False]: ...
@overload
func(arg: Literal[None] = None) -> Literal[None]: ...
```

From:

```c++
m.def("func", [](nb::true_) { return nb::true_(); });
m.def("func", [](nb::false_) { return nb::false_(); });
m.def("func", [](nb::none_) { return nb::none_(); }, "arg"_a = nb::none_());
```

This is very convenient for me. It makes binding and typing simpler in some cases.

If it is OK, I found that the `nb::none()` function can be replaced with the constant `none` class without major changes for users (54d87f357bb539760dd0aaed08c43aab7cc15193).

### DIFF
--- a/include/nanobind/nb_types.h
+++ b/include/nanobind/nb_types.h
@@ -593,7 +593,6 @@ inline void print(const char *str, handle end = handle(), handle file = handle()
     print(nanobind::str(str), end, file);
 }
 
-inline object none() { return borrow(Py_None); }
 inline dict builtins() { return borrow<dict>(PyEval_GetBuiltins()); }
 
 inline iterator iter(handle h) {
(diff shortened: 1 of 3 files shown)


wjakob:
Dear @yosh-matsuda,

I see how this can be useful in a particular project, I think the need for this is too specific for the change to be a part of nanobind. Such an extension can also easily be added in user code.

Regards,
Wenzel

yosh-matsuda:
@wjakob I understand it. Thank you for reviewing.

PR #395, opened 2024-01-08
Add datetime wrapper class

WillAyd:
Per discussion in https://github.com/wjakob/nanobind/discussions/394

I couldn't find a test suite for the similarly designed classes so not sure how to best assert the functionality here. Open to any pointers you might have

### DIFF
--- a/include/nanobind/nb_lib.h
+++ b/include/nanobind/nb_lib.h
@@ -135,7 +135,8 @@ NB_CORE PyObject *bytes_from_cstr_and_size(const char *c, size_t n);
 
 // ========================================================================
 
-NB_CORE PyObject *datetime_from_time_point(const std::chrono::system_clock::time_point &tp);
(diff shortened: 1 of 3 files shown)


wjakob:
nanobind core header files absolutely cannot depend on large headers like `<chrono>`, which seems fundamental to this PR. 

The rationale is explained in the [documentation](https://nanobind.readthedocs.io/en/latest/why.html):

> Smaller headers: #include <pybind11/pybind11.h> pulls in a large portion of the STL (about 2.1 MiB of headers with Clang and libc++). nanobind minimizes STL usage to avoid this problem. Type casters even for for basic types like std::string require an explicit opt-in by including an extra header file (e.g. #include <nanobind/stl/string.h>).

WillAyd:
Ah ok thanks for that context. As an alternative would you consider a custom struct that is like struct tm but with microsecond support? 

That would avoid the stl import issues but still give users some way to create a datetime object directly

PR #405, opened 2024-01-17
Pyarrow_bindings

maximiliank:
Adding type caster for pyarrow bindings


wjakob:
Dear @maximiliank,

while I appreciate the significant effort that must have gone into this, I don't think this a component that needs to be part of nanobind (of which I try to keep the core as small as possible). My suggestion would be that you set up a separate repository. The nanobind documentation could then link to this repository and include a sort of directory of repositories with additional type casters.

Best,
WEnzel

maximiliank:
Dear @wjakob 

thank you for your suggestion. I created a separate repository https://github.com/maximiliank/nanobind_pyarrow

Best,
Max

wjakob:
I added a commit referencing your repository from the documentation just now.

PR #407, opened 2024-01-23
Fix minor typos in documentation

njroussel:
Found a few typos when reading through the documentation.
The custom type slots example was also wrong/flipped.


wjakob:
Thank you 👍

PR #412, opened 2024-01-31
Fix `intrusive_base` documentation

njroussel:
Source documentation for `instrusive_base::set_self_py()` and `intrusive_base::self_py()` were swapped.

The `api_extra.rst` documentation does not need to be changed - it is already correct.


wjakob:
Thanks!

PR #424, opened 2024-02-22
Changes to the `intrusive_base` and `ref<T>` interfaces

njroussel:
This PR introduces the following changes:
* Reference count changes on `intrusive_base` should be considered `const` operations
* Allow implicit casts from `ref<T>` to `T*` and `const T*`
* C-style casts on any use of `inc_ref` and `dec_ref`  to avoid type-checking.


wjakob:
Thank you @njroussel!

PR #428, opened 2024-02-27
Fix leak of `nb_translator_seq` objects.

hawkinsp:
Exception handlers are kept in a linked list, but that list is never freed. Free exception translators immediately before deleting `nb_internals`.


wjakob:
Thanks!

PR #437, opened 2024-02-29
Add typing in python scripts

laggykiller:
Add typing in python scripts, reducing chance of making mistakes.

### DIFF
--- a/include/nanobind/stl/variant.h
+++ b/include/nanobind/stl/variant.h
@@ -22,7 +22,7 @@ struct concat_variant<std::variant<Ts1...>, std::variant<Ts2...>, Ts3...>
     : concat_variant<std::variant<Ts1..., Ts2...>, Ts3...> {};
 
 template <typename... Ts> struct remove_opt_mono<std::variant<Ts...>>
-    : concat_variant<std::conditional_t<std::is_same_v<std::monostate, Ts>, std::variant<>, std::variant<Ts>>...> {};
+    : concat_variant<std::conditional_t<std::is_same_v<std::monostate, Ts>, std::variant<>, std::variant<remove_opt_mono_t<Ts>>>...> {};
 
 template <> struct type_caster<std::monostate> {
     NB_TYPE_CASTER(std::monostate, const_name("None"))


src/stubgen.py
```diff
@@ -94,13 +95,13 @@ def __init__(
...
         # A stack to avoid infinite recursion
-        self.stack = []
+        self.stack: list[Any] = []
 
         # An associated sequence of identifier
         self.prefix = module.__name__ if module else ""
 
         # Dictionary to keep track of import directives added by the stub generator
-        self.imports = {}
+        self.imports: dict[str, dict[Union[str, tuple[Optional[str], str]], Optional[str]]] = {}
```
laggykiller, line 104: Complex type. While this works, consider using a class instead of dictionary?

src/stubgen.py
```diff
@@ -62,11 +63,11 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
-        include_docstrings=True,
-        include_private=False,
-        module=None,
-        max_expr_length=50,
-        patterns={},
+        include_docstrings: bool = True,
+        include_private: bool = False,
+        module: Optional[types.ModuleType] = None,
+        max_expr_length: int = 50,
+        patterns: "dict[re.Pattern[str], tuple[list[str], int]]" = {},
```
laggykiller, line 70: Note that list becomes tuple, as the first item must be list[str] while second item must be int. Using list[Union[list[str], int]] would make type checkers unhappy.

Also, this is complex type. While this works, consider using a class instead of dictionary?

src/stubgen.py
```diff
@@ -545,12 +550,13 @@ def put(self, value, module=None, name=None, parent=None):
                     if not match:
                         continue
 
-                    query_v[1] += 1
+                    self.patterns[query] = (query_v[0], query_v[1] + 1)
```
laggykiller, line 553: Workaround for changing from list to tuple

src/stubgen.py
```diff
@@ -589,8 +595,8 @@ def put(self, value, module=None, name=None, parent=None):
                         groups = match.groups()
                         for i in reversed(range(len(groups))):
                             l = l.replace(f"\\{i+1}", groups[i])
-                        for k, v in match.groupdict():
-                            l = l.replace(f"\\{k}", v)
+                        for k, v in match.groupdict(): # type: ignore
```
laggykiller, line 598: mypy complains of `Unpacking a string is disallowed`

wjakob:
Although I appreciate your work on this, I'm in the midst of a bunch of significant changes and improvements to stubgen. I fear that there will be lots of conflicts. Would it make sense to put this on pause and wait until stubgen is ready to be released? (e.g. closer to the 2.0.0 release)

laggykiller:
I thought adding typing can help with development by offering autocompletes and prevent bugs. But sure, just notify my me when ready!

wjakob:
I wrapped up the refactoring of stubgen and also did a stab at adding type information to it.

laggykiller:
Looks good, it passes pylance standard type checking. I can help add typing so it passes strict type checking, but I will wait until you are ready.

wjakob:
Please go ahead. I am done with this script for now.

wjakob:
(I pushed one more tiny change)

laggykiller:
Please review, thanks!

src/stubgen.py
```diff
@@ -493,7 +524,7 @@ def put_value(self, value: object, name: str, parent=None, abbrev=True) -> None:
 
             # Catch a few different typing.* constructs
             if issubclass(tp, typing.TypeVar) or (
-                hasattr(typing, "TypeVarTuple") and issubclass(tp, typing.TypeVarTuple)
+                sys.version_info >= (3, 11) and issubclass(tp, typing.TypeVarTuple)
```
laggykiller, line 527:
```python
>>> import timeit
>>> timeit.timeit("sys.version_info >= (3, 11)", setup="import sys")
0.029883108043577522
>>> timeit.timeit("hasattr(typing, 'TypeVarTuple')", setup="import typing")
0.04435727297095582
```

src/stubgen.py
```diff
@@ -81,11 +82,45 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+# package_name -> ((name, desired_as_name) -> actual_as_name)
+PackagesDict = Dict[str, ImportDict]
+
+# (signature_str, doc_str, default_arg_1, default_arg_2, ...)
+# We could type as Tuple[Optional[str], Optional[str], *Tuple[Any, ...]]
+# but this is only valid in python >= 3.11
+NbSignature = Tuple[Optional[str], Optional[str]]
+NbGetterSetterSignature = Tuple[str, str]
+
+
```
laggykiller, line 96: We can do:
```python
@typing.runtime_checkable
class NbObject(Protocol):
    __nb_signature__: Tuple[NbSignature, ...]
```

Which allows us to check if a variable has a type that contains all the required field (i.e. `__nb_signature__`) by `isinstance(some_nb_function, NbObject)`. I think this is very neat, but not sure if you have any uses of this.

src/stubgen.py
```diff
@@ -81,11 +82,45 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+class NbObject(Protocol):
```
laggykiller, line 97: We can do:
```python
@typing.runtime_checkable
class NbObject(Protocol):
    __nb_signature__: Tuple[NbSignature, ...]
```

Which allows us to check if a variable has a type that contains all the required field (i.e. `__nb_signature__`) by `isinstance(some_nb_function, NbObject)`. I think this is very neat, but not sure if you have any uses of this.

wjakob:
This is getting kind of large. Would you mind splitting this PR into smaller ones that are easy to accept individually? Many parts of it are edits on lines without code additions. Relatively unobjectionable.

I am less positive about adding quite a bit of typing-related code and data structures (protocols) at the beginning of the file.

laggykiller:
I think using protocol can also serve as documenting the data structures of nanobind objects, as well as providing IDE autocompletions.

However, if you insist on not adding protocol-related code, I can prepare another PR with that removed.

wjakob:
Hm, let me do a more thorough pass over it. I do see the value in documenting the data structures.

It's really important that everything makes sense on Python 3.8 even if it is very old.

src/stubgen.py
wjakob: Fair enough :)

laggykiller:
Rest assure that it is python 3.8 compatible, the non-compatible typings are wrapped in quotes.

src/stubgen.py
wjakob: It's nice. And luckily supported on 3.8

src/stubgen.py
```diff
@@ -81,11 +82,63 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+# but this is only valid in python >= 3.11
+NbSignature = Tuple[Optional[str], Optional[str], Any]
+NbGetterSetterSignature = Tuple[str, str]
+
+
+class NbObject(Protocol):
+    __module__: Literal["nanobind"]
+    __name__: str
+    __nb_signature__: Tuple[NbSignature, ...]
+    __bases__: Tuple[Any, ...]
```
wjakob, line 101: Does the presence of ``__bases__`` mean that this is a type?

src/stubgen.py
```diff
@@ -81,11 +82,63 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+PackagesDict = Dict[str, ImportDict]
+
+# (signature_str, doc_str, default_arg_1, default_arg_2, ...)
+# We could type as Tuple[Optional[str], Optional[str], *Tuple[Any, ...]]
+# but this is only valid in python >= 3.11
+NbSignature = Tuple[Optional[str], Optional[str], Any]
+NbGetterSetterSignature = Tuple[str, str]
+
+
+class NbObject(Protocol):
```
wjakob, line 97: I think that these protocol all need a 1-line comment explaining what they could be. For example, this one is called `NbObject`, but I'm not sure if all nanobind-created objects have a ``_nb_signature__``. Type objects do not, and neither does the metaclass.

src/stubgen.py
```diff
@@ -81,11 +82,63 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+    __name__: Literal["nb_func", "nb_method"]
+    __nb_signature__: Tuple[NbSignature, ...]
+    __func__: "NbFunction"
+    __call__: Callable[..., Any]
+
+
+@typing.runtime_checkable
+class NbGetter(Protocol):
+    __nb_signature__: Tuple[NbGetterSetterSignature, ...]
+    __call__: Callable[[Any], Any]
```
wjakob, line 115: Is it important to have all of these things listed here? For example, I'm pretty sure we are not calling the getter.

I think there is a more fundamental question: are we trying to create some kind of object abstraction layer for nanobind in general? Then stubgen probably isn't the right place for it. I would keep it minimally short and simple, which has also been the goal for other parts of this file.

src/stubgen.py
```diff
@@ -279,7 +331,7 @@ def put_nb_func(self, fn, name: Optional[str] = None) -> None:
                 self.write_ln(f"@{overload}")
                 self.put_nb_overload(fn, s, name)
 
-    def put_function(self, fn, name: Optional[str] = None, parent=None):
+    def put_function(self, fn: NbFunction, name: Optional[str] = None, parent: Optional[object] = None):
```
wjakob, line 334: This function is also used to process pure Python functions without any of the nanobind bits. I am surprised if this type-checks with the `NbFunction` signature.

src/stubgen.py
```diff
@@ -312,8 +364,9 @@ def put_function(self, fn, name: Optional[str] = None, parent=None):
 
         if name is None:
             name = fn.__name__
+            assert name
 
-        overloads: Sequence[Callable] = []
+        overloads: Sequence[Callable[..., Any]] = []
```
wjakob, line 369: Stupid question, isn't `Callable` and `Callable[..., Any]` the same? I don't see the latter type adding any useful extra information.

src/stubgen.py
```diff
@@ -564,7 +621,7 @@ def process_general(m: re.Match) -> str:
 
         return s
 
-    def apply_pattern(self, value: object, pattern: Pattern, match: re.Match) -> None:
+    def apply_pattern(self, value: object, pattern: Pattern, match: "re.Match[str]") -> None:
```
wjakob, line 624: From: https://stackoverflow.com/a/70847364

``for Python 3.8, use typing.Pattern[bytes]``

laggykiller:
All changes done, please review, thanks!

laggykiller:
Pushing from above, should we change:
```python
if (
    type(fget).__module__ == "nanobind"
    and type(fset).__module__ == "nanobind"
):
    assert isinstance(fget, NbGetter)
    assert isinstance(fset, NbSetter)
    ...
```

To:
```python
if isinstance(fget, NbGetter) and isinstance(fset, NbSetter):
    ...
```

`NbGetter` and `NbSetter` has `__nb_signature__`, we can use this to determine if a property is from nanobind.

wjakob:
That looks cleaner indeed.

wjakob:
> All changes done, please review, thanks!

Did you see the other comments above?

laggykiller:
All those comments that have been replied to have been taken action. Anything I missed?

laggykiller:
Wait I see a `re.Pattern`, let me patch that

src/stubgen.py
wjakob: See also this.

src/stubgen.py
wjakob: See also this.

src/stubgen.py
wjakob: See also this.

src/stubgen.py
wjakob: See also this.

src/stubgen.py
wjakob: See also this.

wjakob:
There were a few more (I responded `See also this.` to each one)

laggykiller:
Oh I have taken action in the "Multiple improvements" (bf482084282a9777d6c2190ea2f110c35170d2dd) commit, did that address the problem?

src/stubgen.py
```diff
@@ -81,11 +87,49 @@ class and repeatedly call ``.put()`` to register modules or contents within the
 ]
 # fmt: on
 
+# (name, desired_as_name) -> actual_as_name
```
wjakob, line 90: Add comment before:
```
# This type is used to track per-module imports (``import name as desired_name``) during stub generation. The actual name in the stub is given by the value element.
```

src/stubgen.py
```diff
@@ -81,11 +87,49 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+ImportDict = Dict[Tuple[Optional[str], Optional[str]], Optional[str]]
+# package_name -> ((name, desired_as_name) -> actual_as_name)
```
wjakob, line 92:
```
# This type maps a module name to an `ImportDict` tuple that tracks the import declarations from that module.
```

src/stubgen.py
```diff
@@ -81,11 +87,49 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+PackagesDict = Dict[str, ImportDict]
+
+# (signature_str, doc_str, default_arg_1, default_arg_2, ...)
+# We could type as Tuple[Optional[str], Optional[str], *Tuple[Any, ...]]
```
wjakob, line 96: That is neat but doesn't seem like a useful strategy for now because of the minumum Python requirement. Let's remove that tangential comment.

src/stubgen.py
```diff
@@ -81,11 +87,49 @@ class and repeatedly call ``.put()`` to register modules or contents within the
 ]
 # fmt: on
 
+# (name, desired_as_name) -> actual_as_name
+ImportDict = Dict[Tuple[Optional[str], Optional[str]], Optional[str]]
+# package_name -> ((name, desired_as_name) -> actual_as_name)
+PackagesDict = Dict[str, ImportDict]
+
+# (signature_str, doc_str, default_arg_1, default_arg_2, ...)
```
wjakob, line 95: Add comment:
```
# Type of an entry of the ``__nb_signature__`` tuple of nanobind functions. It stores a function signature string, docstring, and a tuple of default function values.
```
A random thought: since this type seems to be so difficult to describe to Python, how about creating it differently with signature `Tuple[str, Optional[str], Optional[Tuple[Any, ...]]]`.

Here is a change to ``nb_func.cpp`` that accomplishes this:
```diff
diff --git a/src/nb_func.cpp b/src/nb_func.cpp
index 8c5602d..406fb70 100644
--- a/src/nb_func.cpp
+++ b/src/nb_func.cpp
@@ -1116,7 +1116,8 @@ static PyObject *nb_func_get_module(PyObject *self) {
 }
 
 PyObject *nb_func_get_nb_signature(PyObject *self, void *) {
-    PyObject *docstr = nullptr, *item = nullptr, *sigstr = nullptr;
+    PyObject *docstr = nullptr, *item = nullptr, *sigstr = nullptr,
+             *defaults = nullptr;
 
     func_data *f = nb_func_data(self);
     uint32_t count = (uint32_t) Py_SIZE(self);
@@ -1125,7 +1126,7 @@ PyObject *nb_func_get_nb_signature(PyObject *self, void *) {
         return nullptr;
 
     for (uint32_t i = 0; i < count; ++i) {
-        docstr = item = sigstr = nullptr;
+        docstr = item = sigstr = defaults = nullptr;
 
         const func_data *fi = f + i;
         if (fi->flags & (uint32_t) func_flags::has_doc && fi->doc[0] != '\0') {
@@ -1136,18 +1137,22 @@ PyObject *nb_func_get_nb_signature(PyObject *self, void *) {
         }
 
         buf.clear();
-        uint32_t n_default_args = nb_func_render_signature(fi, true),
-                 pos = 2;
+        uint32_t n_default_args = nb_func_render_signature(fi, true);
 
-        item = PyTuple_New(2 + n_default_args);
+        item = PyTuple_New(3);
         sigstr = PyUnicode_FromString(buf.get());
-        if (!docstr || !sigstr || !item)
-            goto fail;
+        if (n_default_args) {
+            defaults = PyTuple_New(n_default_args);
+        } else {
+            defaults = Py_None;
+            Py_INCREF(defaults);
+        }
 
-        NB_TUPLE_SET_ITEM(item, 0, sigstr);
-        NB_TUPLE_SET_ITEM(item, 1, docstr);
+        if (!docstr || !sigstr || !item || !defaults)
+            goto fail;
 
         if (n_default_args) {
+            size_t pos = 0;
             for (uint32_t j = 0; j < fi->nargs; ++j) {
                 const arg_data &arg = fi->args[j];
                 PyObject *value = arg.value;
@@ -1160,14 +1165,17 @@ PyObject *nb_func_get_nb_signature(PyObject *self, void *) {
                 } else {
                     Py_INCREF(value);
                 }
-                NB_TUPLE_SET_ITEM(item, pos, value);
+                NB_TUPLE_SET_ITEM(defaults, pos, value);
                 pos++;
             }
-        }
 
-        check(pos == n_default_args + 2,
-              "__nb_signature__: default argument counting inconsistency!");
+            check(pos == n_default_args,
+                  "__nb_signature__: default argument counting inconsistency!");
+        }
 
+        NB_TUPLE_SET_ITEM(item, 0, sigstr);
+        NB_TUPLE_SET_ITEM(item, 1, docstr);
+        NB_TUPLE_SET_ITEM(item, 2, defaults);
         NB_TUPLE_SET_ITEM(result, (Py_ssize_t) i, item);
     }
 
@@ -1176,6 +1184,7 @@ PyObject *nb_func_get_nb_signature(PyObject *self, void *) {
 fail:
     Py_XDECREF(docstr);
     Py_XDECREF(sigstr);
+    Py_XDECREF(defaults);
     Py_XDECREF(item);
     Py_DECREF(result);
     return nullptr;
```

src/stubgen.py
```diff
@@ -81,11 +87,49 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+# but this is only valid in python >= 3.11
+NbSignature = Tuple[Optional[str], Optional[str], Any]
+NbGetterSetterSignature = Tuple[str, str]
+
+
+class NbFunction(Protocol):
+    __module__: Literal["nanobind"]
+    __name__: Literal["nb_func", "nb_method"]
+    __nb_signature__: Tuple[NbSignature, ...]
+    __func__: "NbFunction"
```
wjakob, line 106: Could you explain why the ``__func__`` and ``__call__`` elements are needed here? Maybe I just don't fully understand how these protocols are used during type-checking.

wjakob (commented):
Here is some more feedback 🙇

laggykiller:
ok I tried to open this github PR page in incognito mode. Seems like my comments could not be seen...?

It has a `Pending` label on my comment. Pressing `Resolve conversation` did not make my comments appear...

laggykiller:
Oops I completely forgot about adding comments... Copied your comments and added in latest commit.

---

> how about creating it differently with signature `Tuple[str, Optional[str], Optional[Tuple[Any, ...]]]`

I was so upset that python did not accept `Tuple[Optional[str], Optional[str], Any, ...]` and had the same thought of changing C++ code so that the default args are in tuple, but I thought you would think this is too much work and did not mention about this.

Should I make this change for you? Or leave this to you?

---

> Could you explain why the __func__ and __call__ elements are needed here?

For the `__func__`, it **was** required as I typed the `fn` in `put_function()` as `NbFunction`, and there is access for `fn.__func__`. It is not required now, let me remove that.

For the `__call__` in `NbFunction`, it is necessary as this is what make a python object a function. I called `fn = cast(NbFunction, fn)` before `self.put_nb_func(fn, name)` in `put_function()` to inform type checker that `fn` is `NbFunction` so it does not complain that I am putting an incompatible `fn` arg to `self.put_nb_func`. If there is no `__call__` I cannot cast `fn` to `NbFunction` as `fn` is originally a `Callable[..., Any]` and it should cast to a `Callable`, in which `Callable` has `__call__`. If that sound confusing, you can try to remove `__call__` in `NbFunction` and see how a strict type checker react.

laggykiller:
If we are doing the change to `nb_func.cpp`:

```diff
diff --git a/src/stubgen.py b/src/stubgen.py
index 955cf68..2fea6d9 100755
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -99,17 +99,19 @@ PackagesDict = Dict[str, ImportDict]
 
 # Type of an entry of the ``__nb_signature__`` tuple of nanobind functions.
 # It stores a function signature string, docstring, and a tuple of default function values.
-# (signature_str, doc_str, default_arg_1, default_arg_2, ...)
-NbSignature = Tuple[Optional[str], Optional[str], Any]
+# (signature_str, doc_str, (default_arg_1, default_arg_2, ...))
+NbSignature = Tuple[Optional[str], Optional[str]]
+
+# Type of an entry of the ``__nb_signature__`` tuple of nanobind functions.
+NbFunctionSignature = Tuple[Optional[str], Optional[str], Optional[Tuple[Any, ...]]]
 
 # Type of an entry of the ``__nb_signature__`` tuple of nanobind getters and setters.
 NbGetterSetterSignature = Tuple[str, str]
 
-
 class NbFunction(Protocol):
     __module__: Literal["nanobind"]
     __name__: Literal["nb_func", "nb_method"]
-    __nb_signature__: Tuple[NbSignature, ...]
+    __nb_signature__: Tuple[NbFunctionSignature, ...]
     __call__: Callable[..., Any]
 
 
@@ -240,7 +242,7 @@ class StubGen:
         docstr = f'{raw_str}"""{docstr}"""\n'
         self.write_par(docstr)
 
-    def put_nb_overload(self, fn: NbFunction, sig: NbSignature, name: Optional[str] = None) -> None:
+    def put_nb_overload(self, fn: NbFunction, sig: NbFunctionSignature, name: Optional[str] = None) -> None:
         """
         The ``put_nb_func()`` repeatedly calls this method to render the
         individual method overloads.
@@ -256,45 +258,47 @@ class StubGen:
         sig_str = sig_str[:paren] + self.simplify_types(sig_str[paren:])
 
         # Substitute in string versions of the default arguments
-        for index, arg in enumerate(sig[2:]):
-            pos = -1
-            custom_signature = False
-            pattern = None
-
-            # First, handle the case where the user overrode the default value signature
-            if isinstance(arg, str):
-                pattern = f"\\={index}"
-                pos = sig_str.find(pattern, start)
-                if pos >= 0:
-                    custom_signature = True
-
-            # General case
-            if pos < 0:
-                pattern = f"\\{index}"
-                pos = sig_str.find(pattern, start)
-
-            if pos < 0:
-                raise Exception(
-                    "Could not locate default argument in function signature"
-                )
+        default_args = sig[2]
+        if default_args:
+            for index, arg in enumerate(default_args):
+                pos = -1
+                custom_signature = False
+                pattern = None
+
+                # First, handle the case where the user overrode the default value signature
+                if isinstance(arg, str):
+                    pattern = f"\\={index}"
+                    pos = sig_str.find(pattern, start)
+                    if pos >= 0:
+                        custom_signature = True
+
+                # General case
+                if pos < 0:
+                    pattern = f"\\{index}"
+                    pos = sig_str.find(pattern, start)
+
+                if pos < 0:
+                    raise Exception(
+                        "Could not locate default argument in function signature"
+                    )
 
-            if custom_signature:
-                arg_str = arg
-            else:
-                # Call expr_str to convert the default value to a string.
-                # Abbreviate with '...' if it is too long.
-                arg_str = self.expr_str(arg, abbrev=True)
-                if arg_str is None:
-                    arg_str = "..."
-
-            assert isinstance(arg_str, str)
-            assert (
-                "\n" not in arg_str
-            ), "Default argument string may not contain newlines."
-
-            assert pattern is not None
-            sig_str = sig_str[:pos] + arg_str + sig_str[pos + len(pattern) :]
-            start = pos + len(arg_str)
+                if custom_signature:
+                    arg_str = arg
+                else:
+                    # Call expr_str to convert the default value to a string.
+                    # Abbreviate with '...' if it is too long.
+                    arg_str = self.expr_str(arg, abbrev=True)
+                    if arg_str is None:
+                        arg_str = "..."
+
+                assert isinstance(arg_str, str)
+                assert (
+                    "\n" not in arg_str
+                ), "Default argument string may not contain newlines."
+
+                assert pattern is not None
+                sig_str = sig_str[:pos] + arg_str + sig_str[pos + len(pattern) :]
+                start = pos + len(arg_str)
 
         if type(fn).__name__ == "nb_func" and self.depth > 0:
             self.write_ln("@staticmethod")
```

```diff
diff --git a/tests/test_functions.py b/tests/test_functions.py
index f415e69..d0f3593 100644
--- a/tests/test_functions.py
+++ b/tests/test_functions.py
@@ -420,7 +420,7 @@ def test39_del():
 def test40_nb_signature():
     assert t.test_01.__nb_signature__ == ((r"def test_01() -> None", None),)
     assert t.test_02.__nb_signature__ == (
-        (r"def test_02(j: int = \0, k: int = \1) -> int", None, 8, 1),
+        (r"def test_02(j: int = \0, k: int = \1) -> int", None, (8, 1)),
     )
     assert t.test_05.__nb_signature__ == (
         (r"def test_05(arg: int, /) -> int", "doc_1"),
```

laggykiller:
I am committing anyway as I need to sleep. In case you don't want it, I will just revert the commit.

Please review. Feel free to merge when you are satisfied.

wjakob:
It looks great! Could I ask you to squash this into two commits and rebase on top of master? The first to adapt `nb_func.cpp` and related tests, and the second one which improves typing of stubgen. I will merge the PR following this. Thanks!

laggykiller:
Let me re-open another PR...

laggykiller:
Seems like I can reopen here

PR #438, opened 2024-03-01
Add `.name` and `.value` properties to `nb::enum_` objects.

hawkinsp:
This is to match both `enum.Enum`, which has these properties (https://docs.python.org/3/library/enum.html#enum.Enum) and to simplify porting code from pybind11, which has them also.


wjakob:
Thank you, that looks like a useful addition.

PR #440, opened 2024-03-01
Fix incorrect visibility when compiling on MinGW64 with gcc

laggykiller:
Whe compiling on ~~cygwin~~ mingw64 using gcc, errors like this occur:
```
nb_error.h:30:17: error: 'dllexport' implies default visibility, but 'class nanobind::python_error' has already been declared with a different visibility
     30 | class NB_EXPORT python_error : public std::exception {
        |                 ^~~~~~~~~~~~
```

This PR should fix this, with formatting changes in nearby code.

I made reference with https://gcc.gnu.org/wiki/Visibility

### DIFF
--- a/include/nanobind/nb_defs.h
+++ b/include/nanobind/nb_defs.h
@@ -22,22 +22,29 @@
 #  define NAMESPACE_END(name) }
 #endif
 
-#if defined(_WIN32)
-#  define NB_EXPORT        __declspec(dllexport)
-#  define NB_IMPORT        __declspec(dllimport)
-#  define NB_INLINE        __forceinline
+#if defined _WIN32 || defined __CYGWIN__
+#  ifdef __GNUC__
+#    define NB_EXPORT        __attribute__ ((visibility("default")))
+#    define NB_IMPORT        NB_EXPORT
+#    define NB_INLINE        inline __attribute__((always_inline))
+#    define NB_NOINLINE      __attribute__((noinline))
+#  else
+#    define NB_EXPORT        __declspec(dllexport)
+#    define NB_IMPORT        __declspec(dllimport)
+#    define NB_INLINE        __forceinline
+#    define NB_NOINLINE      __declspec(noinline)
+#  endif
 #  define NB_INLINE_LAMBDA
-#  define NB_NOINLINE      __declspec(noinline)
 #else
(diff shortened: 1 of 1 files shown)


wjakob:
Hmm, I have no plans to support CYGWIN officially as a compiler (This would require CI runners and then dealing with problems related to CYGWIN, which adds maintenance costs). I am also very hesitant to even mention its name anywhere  in the source code since that indicates to the reader that it is supported somehow.

Does the following also work? (i.e., does CYGWIN set `_WIN32`?)

```cpp
#if defined(_WIN32) || defined __CYGWIN__
#  defined(__GNUC__)
```
Let's stick to the `#if defined(condition)` syntax which is used throughout the codebase.

laggykiller:
Seems like newer version of cygwin no longer define _WIN32

Source: https://stackoverflow.com/questions/47150492/how-to-get-gcc-on-cygwin-to-define-win32

wjakob:
Then I don't understand the change you are making. If `_WIN32` is not defined, the preprocessor will use the second GCC-specific part of the definitions. Which of these definitions is it that's causing a problem for CYGWIN?

laggykiller:
I just copied boilerplate code without thinking... You are right about that, I have removed check for cygwin, please check!

wjakob:
> Seems like newer version of cygwin no longer define _WIN32 (https://stackoverflow.com/questions/47150492/how-to-get-gcc-on-cygwin-to-define-win32)
>
> MinGW has _WIN32 defined though (https://stackoverflow.com/questions/9025708/mingw-not-defining-win32-error-in-preprocessor-directives)

This makes sense. MinGW is supposed to be able to compile code written for the microsoft compiler, and Cygwin wants to behave more like stock GCC on Linux.

But then I don't understand the change you are making. It would seem that MinGW will parse the top set of definitions for the microsoft compiler, and the latest versions of Cygwin will use the bottom ones. You should not need to put ``__GNUC__`` specific parts into the `#if defined(_WIN32)` block.

laggykiller:
I think I made a mistake in the title, I actually discovered the problem in msys2 mingw64, not cygwin...

wjakob:
Great. This change is now something I can sign off on. Shall I merge it?

laggykiller:
Sure, thank you!

PR #443, opened 2024-03-02
Change output name of `nb::object` to `Any`

yosh-matsuda:
The `typing.Any` seems better for the output type names of `nb::object` and `nb::handle`.

The type checker may complain to the user's code if the return type is annotated with the `object`. This is because the `object` is the base class of all objects in Python, and only operations that are valid for all types are allowed on the `object` type.

ref: https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object


wjakob:
I'm not so sure about this change. Even the typing documentation says that one should prefer `object` if in doubt.

> If you’re not sure whether you need to use [object](https://docs.python.org/3/library/functions.html#object) or Any, use [object](https://docs.python.org/3/library/functions.html#object) – only switch to using Any if you get a type checker complaint.

> You can use different [type narrowing](https://mypy.readthedocs.io/en/latest/type_narrowing.html#type-narrowing) techniques to narrow [object](https://docs.python.org/3/library/functions.html#object) to a more specific type (subtype) such as int. Type narrowing is not needed with dynamically typed values (values with type Any).

What about adding a type ``nb::any`` that is equivalent to ``nb::object`` but renders differently in type signatures?

yosh-matsuda:
I thought that returning a (pure) nb::object in nanobind effectively abandons the type annotation so that the equivalent would be `Any` for that situation. Adding `nb::any_` seems flexible. This leads `Any` can be annotated for the arguments, too.

I can create a new PR about this. Can I ask you what name is good since `nb::any` is already defined in `nb_ndarray.h`. Implementation is like:

```cpp
class any_ : public object {
public:
    using object::object;
    using object::operator=;
    static constexpr auto Name = detail::const_name("typing.Any");
};
```

wjakob:
Great. I think in this case it's more important the ``Any`` type to be associated with `nb::any` than the ndarray size tag. The 2.0.0 release is the perfect opportunity for such a compatibility-breaking change.  I made a commit to a new `remove_any` branch. Could I ask you to make the change on top and then create a new commit combining everything? Please mention the new wrapper in the documentation and add a testcase that checks that it's rendered as expected in function signatures.

wjakob:
Closing this in favor of #460.

PR #444, opened 2024-03-02
Fix nested redundant `None` in `variant`

yosh-matsuda:
This PR fixes ' None' is still redundant when the `Optional` is in `std::variant`.

The argument of `std::variant<T, std::optional<U>>` should be typed as `T | U | None`, but it is `T | U | None | None `. This is my old mistake about `remove_opt_mono`.


wjakob:
Great, thank you!

PR #448, opened 2024-03-05
Add support for keyword-only arguments

oremanj:
I know this was in the "not going to support it" list, but I figured out how to do it with extremely minimal overhead, so I'm politely asking you to reconsider that position. :-) All ordering/etc checks happen at compile-time, the size of `func_data` doesn't change, and `nb_func.o` gets very slightly *smaller* on my laptop. The 'simple' function dispatcher is completely unaffected and the changes to the 'complex' one are very minimal.


oremanj:
I'm pretty sure the pypy failure is not my fault, as I see the same failure in #442.

docs/porting.rst
```diff
@@ -334,8 +334,8 @@ Removed features include:
   executable or run several independent Python interpreters in the same process
   is unsupported. Nanobind caters to bindings only. Multi-interpreter support
   would require TLS lookups for nanobind data structures, which is undesirable.
-- ○ **Function binding annotations**: the ``kw_only`` / ``pos_only`` argument
-  annotations were removed.
+- ○ **Function binding annotations**: The ``pos_only`` argument
+  annotation was removed.
```
wjakob, line 338: ADD: "However, the same behavior can be achieved by creating unnamed arguments (see the discussion in the section on :ref:`keyword-only arguments <kw_only>`)"

include/nanobind/nb_func.h
```diff
@@ -54,22 +60,61 @@ NB_INLINE PyObject *func_create(Func &&func, Return (*)(Args...),
         nargs = sizeof...(Args);
 
     // Determine the number of nb::arg/nb::arg_v annotations
-    constexpr size_t nargs_provided =
+    static constexpr size_t nargs_provided =
```
wjakob, line 63: I am curious -- is the addition of `static` for these `constexpr` parameters just general best practice, or do you think it has an actual impact here?

include/nanobind/nb_func.h
```diff
@@ -54,22 +60,61 @@ NB_INLINE PyObject *func_create(Func &&func, Return (*)(Args...),
...
+    static constexpr bool is_method_det =
         (std::is_same_v<is_method, Extra> + ... + 0) != 0;
-    constexpr bool is_getter_det =
+    static constexpr bool is_getter_det =
         (std::is_same_v<is_getter, Extra> + ... + 0) != 0;
+    static constexpr bool has_arg_annotations =
+        nargs_provided > 0 && !is_getter_det;
+
+    static constexpr size_t
+        kwonly_pos_1 = index_1_v<std::is_same_v<kw_only, Extra>...>,
```
wjakob, line 73: What's the difference between `kwonly_pos_1` and `kwonly_pos_n`? Did you test the assertions below in the case of invalid parameter specifications?

include/nanobind/nb_func.h
```diff
@@ -54,22 +60,61 @@ NB_INLINE PyObject *func_create(Func &&func, Return (*)(Args...),
...
-    static_assert(args_pos_1 == nargs || args_pos_1 + 1 == kwargs_pos_1,
-        "nb::args must follow positional arguments and precede nb::kwargs!");
+    static_assert(args_pos_1 == nargs || args_pos_1 < kwargs_pos_1,
+        "nb::args must precede nb::kwargs if both are present!");
+    static_assert(has_arg_annotations || (!implicit_kw_only && !explicit_kw_only),
+        "Keyword-only arguments must have names!");
+
+    // Avoid instantiating count_args_before_index unless the kw_only
+    // annotation is actually used
+    static constexpr size_t nargs_before_kw_only = []() constexpr {
```
wjakob, line 93: Not a fan of this one, it seems kind of complicated. I am suspicious of what a compiler like MSVC will make out of this (I've seen it struggle with things as basic as inlining a 1-line lambda function, and it is generally bad at constexpr computation).

include/nanobind/nb_attr.h
```diff
@@ -156,8 +157,17 @@ template <size_t Size> struct func_data_prelim {
...
 
-    /// Total number of function call arguments
-    uint32_t nargs;
+    /// Total number of arguments accepted by the C++ function; nb::args
+    /// and nb::kwargs parameters are counted as one each. If the
+    /// 'has_args' flag is set, then there is one arg_data structure
+    /// for each of these.
+    uint16_t nargs;
+
+    /// Number of arguments to the C++ function that may be filled from
```
wjakob, line 166: Just to clarify: this includes ``nb::arg()`` (unnamed) and nb::arg("name")` (named) arguments?

src/nb_func.cpp
```diff
@@ -277,13 +275,13 @@ PyObject *nb_func_new(const void *in_) noexcept {
     check(func, "nb::detail::nb_func_new(\"%s\"): alloc. failed (1).",
           name_cstr);
 
-    func->max_nargs_pos = f->nargs;
-    func->complex_call = has_args || has_var_args || has_var_kwargs || has_keep_alive;
+    func->max_nargs = f->nargs;
+    func->complex_call = f->nargs_pos < f->nargs || has_args || has_keep_alive;
```
wjakob, line 279: Is there any particular reason that we need a less than comparison, instead of `!=`? I'm assuming `nargs_pos` can never actually be larger than `nargs`, correct? No need to change anything here, I am just wondering.

include/nanobind/nb_func.h
```diff
@@ -165,6 +209,8 @@ NB_INLINE PyObject *func_create(Func &&func, Return (*)(Args...),
     f.descr = descr.text;
     f.descr_types = descr_types;
     f.nargs = nargs;
+    f.nargs_pos =   args_pos_1 < nargs ? args_pos_1 :
```
wjakob, line 212: Would you mind adding a comment that spells out the logic here in english?

src/nb_func.cpp
```diff
@@ -563,8 +561,15 @@ static PyObject *nb_func_vectorcall_complex(PyObject *self,
...
-            size_t nargs_pos = f->nargs - has_var_args - has_var_kwargs;
+            // Number of C++ arguments eligible to be filled from individual
+            // Python positional arguments
+            size_t nargs_pos = f->nargs_pos;
+
+            // Number of C++ arguments eligible to be filled from individual
+            // Python arguments (either positional or keyword), except that:
+            // if has_var_args, then the one at index nargs_pos is a nb::args
+            // that we'll fill in afterward.
+            size_t nargs_step1 = f->nargs - has_var_kwargs;
```
wjakob, line 572: Not sure I understand. Your comment mentions `has_var_args`, and that explanation made sense to me. But the code then uses `has_var_kwargs`.

src/nb_func.cpp
```diff
@@ -575,14 +580,17 @@ static PyObject *nb_func_vectorcall_complex(PyObject *self,
...
+            for (; i < nargs_step1; ++i) {
+                if (has_var_args && i == nargs_pos)
+                    continue; // skip nb::args argument, will be handled below
+
                 PyObject *arg = nullptr;
                 bool arg_convert  = pass == 1,
                      arg_none     = false;
 
-                if (i < nargs_in)
+                if (i < nargs_in && i < nargs_pos)
```
wjakob, line 593: didn't you already `continue` in the `i == nargs_po`s case just a few lines before?

wjakob (commented):
You are correct that this is a feature that isn't banned in any way. I just thought that the overload dispatch loop in pybind11 was way overkill. Here is some first feedback.

This actually looks really good -- remarkably simple as you say. I do have a few comments and questions though.

include/nanobind/nb_func.h
oremanj: This is mostly a habit I have because of past experience with compilers thinking constexpr non-statics need to be captured in lambdas. If I remove the lambda then these definitely aren't necessary. I was confused why some of them were previously static and some not, though.

include/nanobind/nb_func.h
oremanj: Switched to a version that doesn't use a lambda.

include/nanobind/nb_func.h
oremanj: ... just realized I accidentally was using index_1 in both. Fixed that and will test that multiple kw_only annotations produces an error.

src/nb_func.cpp
oremanj: Either would work; nargs_pos is always <= nargs, and always == for "simple" calls with no kwonly/varargs/varkwds.

src/nb_func.cpp
oremanj: Reworded for more clarity. A C++ parameter that takes variadic kwargs is excluded from this count. A C++ parameter that takes variadic positional args is included in the count but then skipped in the loop below. Rationale: if a nb::args parameter is present, it's between the pos-or-kw args and the kw-only args; it's better to deal with both of those categories in the same loop below than to duplicate the code for each type.

oremanj:
Thank you for the prompt review! I made some updates in response to your comments, mostly better comments with a couple functional fixes (detecting double kw_only annotations, handling kw_only before `**kwargs` with no `*args`). Please take another look when you have a moment.

include/nanobind/nb_func.h
```diff
@@ -60,16 +68,59 @@ NB_INLINE PyObject *func_create(Func &&func, Return (*)(Args...),
...
+        "nb::args must precede nb::kwargs if both are present!");
+    static_assert(has_arg_annotations || (!implicit_kw_only && !explicit_kw_only),
+        "Keyword-only arguments must have names!");
+
+    // Find the index in Args... of the first keyword-only parameter. Since
+    // the 'self' parameter doesn't get a nb::arg annotation, we must adjust
+    // by 1 for methods. Note that nargs_before_kw_only is only used if
+    // a kw_only annotation exists (i.e., if explicit_kw_only is true);
+    // the conditional is just to save the compiler some effort otherwise.
+    constexpr size_t nargs_before_kw_only =
```
wjakob, line 102: stupid question: what's the difference between keeping this code here as `constexpr size_t` or just making a `size_t` variable and then potentially overwriting it in the `constexpr` block? If efficiency (avoiding the template expansion) is the reason, then it would seem to me that the compiler must in any case be smart enough to not expand templates in a `if constexpr (false)` block to avoid potential exponential instantiation of templates in recursive constexpr programs. Anyways, this way of writing things down seems somehow more natural (closer to the behavior of a regular non-constexpr program).

include/nanobind/nb_func.h
wjakob: Perfection!

src/nb_func.cpp
wjakob: Awesome.

wjakob (commented):
Perfect! Only a few minor comments at this point.

include/nanobind/nb_func.h
oremanj: We need the `nargs_before_kw_only` value available as a constexpr in order to static_assert about it in the `explicit_kw_only` block, and we also need it down at the bottom of the function to initialize `nargs_pos`. The latter doesn't need to be constexpr, so I guess we could do something like
```
size_t nargs_before_kw_only = nargs;
if constexpr (explicit_kw_only) {
    constexpr size_t cx_nargs_before_kw_only = is_method_det + count_args_before_index<...>(...);
    // static_assert about cx_nargs_before_kw_only
    nargs_before_kw_only = cx_nargs_before_kw_only;
}
```
but that seemed like way too much trouble to go to when there's not any reason to believe the instantiation of `count_args_before_index` is particularly expensive. I was only trying to avoid it on principle, but the cure seemed worse than the disease in the end.

src/nb_func.cpp
wjakob: This is actually a quite important point, would you mind merging the comment into the code?

This was sort of the issue with the pybind11 version of the function call dispatcher -- it's a lot of tricky indexing. After a while, it became impossible understand the implementation without spending a significant amount of time on it -- comments like this go a long way in making the code more accessible.

include/nanobind/nb_func.h
wjakob: Got it. Let's leave it like this then.

wjakob:
I think this ready to go in. Could you add a changelog entry and then squash the commits?

oremanj:
Should be all set!

wjakob:
Thank you!

PR #450, opened 2024-03-06
Remove misplaced comment

tjstum:
Noticed this when upgrading our internal version. Looks like the comment accidentally got split.


wjakob:
Good catch, thanks!

PR #454, opened 2024-03-06
make_key/value_iterator: support iterators that return a non-reference from operator*

oremanj:
These can show up in certain map-like containers where the keys are stored implicitly. For example, you can imagine a map where the keys are relatively dense iterators, which chooses to store the values in a contiguous array/vector and have the key for each value be implied by its index in that vector. Dereferencing an iterator for such a map returns `pair<Key, Value&>` rather than `pair<const Key, Value>&`. Previously, nanobind's `make_key_iterator` and `make_value_iterator` functions would try to form a reference to a temporary (complete with compiler warning) if you passed them an iterator like that; after this change, they work correctly.

I verified that the unit test added here fails if I run it without the change to make_iterator.h that's being made here.

### DIFF
--- a/include/nanobind/make_iterator.h
+++ b/include/nanobind/make_iterator.h
@@ -25,6 +25,11 @@ struct iterator_state {
     bool first_or_done;
 };
 
+template <typename T>
+struct remove_rvalue_ref { using type = T; };
+template <typename T>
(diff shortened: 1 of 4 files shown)


wjakob:
This makes sense to me. I am surprised there isn't a `std::remove_rvalue_reference_t`.

PR #455, opened 2024-03-06
Support binding unions

oremanj:
This almost works already, except that `is_base_of` never returns true for unions, even when testing whether something is a base of itself. Add an `is_same` alternative, and a test to make sure this doesn't regress.

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -842,7 +842,7 @@ but once again each instantiation must be explicitly specified:
         template <typename V> T fn(V v);
     };
 
-    nb::class<MyClass<int>>(m, "MyClassT")
+    nb::class_<MyClass<int>>(m, "MyClassT")
         .def("fn", &MyClass<int>::fn<std::string>);
 
 .. _tag_based_polymorphism:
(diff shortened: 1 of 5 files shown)


oremanj:
I'll fix the type punning in the unit test. As for the broader question: Nothing prevents it directly, just like nothing prevents you from returning a pointer you don't own and getting a double-free due to the default `take_ownership` rv_policy. If we prevent people from ever committing UB, we also make it harder to bind some perfectly valid uses; for example, it's valid to write a union of standard-layout types and examine their common subsequence (which might be a tag indicating which member is active).

For reference, pybind11 supports unions (https://github.com/pybind/pybind11/pull/2320) and doesn't mention them at all except in the changelog entry introducing the feature, which included a brief warning about not using it to introduce UB. I tend to think that if someone reaches for binding a union, they probably have an unusual use case in mind (mine is formatting wire messages in a network protocol that includes a union) and it's better to leave them to it.

wjakob:
Ok, fair enough. Could I ask you to: add an entry about binding unions in the classes.rst part of the documentation explaining the feature, and that t this should be used analogously to a variant type? (can't write one entry and then read another)

oremanj:
Thanks! Added docs and squashed the commits.

wjakob:
Perfect, that's a nice description as well. 👍

PR #456, opened 2024-03-06
Fix returning pointer-to-const-T where T uses a type caster

oremanj:
The `from_cpp(T*)` function emitted by `NB_TYPE_CASTER` previously only accepted pointers-to-mutable-T, so lots of types would produce a compiler error if you returned pointer-to-const-T. This PR fixes it to accept pointer-to-cv-T with any cv qualification, and adds a regression test.


wjakob:
Thanks!

PR #459, opened 2024-03-07
docs: Update copyright dates

hpkfft:
This updates the copyright footer in the html documentation to "2023 - 2024" using an EnDash (rather than a hyphen) and using ThinSpace on both sides of it.


wjakob:
It seems fine the original way, I prefer not to have mysterious \u escape codes here.

Issue #420, opened 2024-02-14
Tracking issue: stub generation


wjakob:
Dear all (cc @cansik @torokati44 @qnzhou @tmsrise @rzhikharevich-wmt @njroussel @Speierers),

I am interested in providing a stub generation mechanism as part of nanobind. This is a tracking issue to brainstorm solutions.

**Context**: @cansik's [nanobind-stubgen](https://github.com/cansik/nanobind-stubgen) package is the only solution at the moment and works well in many cases. My goal is to overcome limitations discussed in discussion #163:

1. Enabling a better "out-of-the-box" experience by integrating stub generation into the CMake build system.

2. Stub generation currently involves complicated parsing, which is fragile and not always well-defined. Nanobind has this information in a more structured form and could provide it.

4. Stubs serve two purposes, and stub generation should cater to both needs:
  - To get autocomplete in VS Code and similar tools, which requires extracting function signatures and docstrings. I am mainly interested in this use case.

  - Type checkers like MyPy. I haven't used them before and know very little (hence this issue to exchange experience). It seems to me that stubs only need to contain typed signatures but no docstrings. But nanobind often generates type annotations that MyPy isn't happy with, so some sort of postprocessing may be needed.

Here is what I have in mind, before having actually having done anything. There may be roadblocks I haven't considered.

1. The CMake build system gets a new command `nanobind_add_stubs`. This will register a command that is run at _install time_. Basically we need the whole package to be importable, and doing that in a non-installed build might be tricky. 

```cmake
nanobind_add_stubs(
  PATH ${CMAKE_INSTALL_PREFIX}
  PACKAGE nanobind_example
  DEPENDS nanobind_example_ext
)
```
When the user installs the extension to `${CMAKE_INSTALL_PREFIX}`, this will run a Python file (shipped as part of the nanobind distribution) that imports the package and then generates `nanobind_example/__init__.pyi`.

Here, I am already getting confused because of unfamiliarity with stub generation. I've seen that packages sometimes contain multiple `.pyi` files. How does one decide where to put what? Can `.pyi` files import each other? What would be the best way to expose this in the `nanobind_add_stubs()` function?

2. I also wanted to modify nanobind's function class (`nb_func`) so that it exposes information in a more structured way, a bit like `__signature__` from `inspect.signature`. But `__signature__` is too limited because it (like Python) has no concept of overload chains.

[...]

PR #462, opened 2024-03-10
Sorted imports

laggykiller:
See https://github.com/wjakob/nanobind/issues/420#issuecomment-1986957935

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -59,6 +59,7 @@ from inspect import Signature, Parameter, signature, ismodule, getmembers
 import textwrap
 import importlib
 import importlib.machinery
+import importlib.util
 import types
 import typing
 from dataclasses import dataclass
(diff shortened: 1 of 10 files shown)


wjakob:
Regarding the PackagesDict data structure. Can this not be done in an easier way all the way at the end when the import list is being generated? Instead of traversing `sorted()`, you could use `sorted(.., key=)` where a key looks at the module name that distinguishes between `typing` + `collections`, anything starting with `self.module.__name__.split('.')[0]`, and everything else

laggykiller:
> Instead of traversing sorted(), you could use sorted(.., key=)

This would be the best solution, but I want to add extra line break between imports from different party, e.g.:
```
from typing import Sequence

from numpy.typing import ArrayLike
```

instead of:
```python
from typing import Sequence
from numpy.typing import ArrayLike
```

It would not be possible to tell where does stdlib imports end and 3rd party imports start -> Unable to add line break

However, I did simplify things a bit in latest commit, please check

laggykiller:
Alternative solution:
```python
    def get(self) -> str:
        """Generate the final stub output"""
        s = ""
        last_party = 0

        for module in sorted(self.imports, key=lambda module: str(self.check_party(module)) + module):
            imports = self.imports[module]
            items: List[str] = []
            party = self.check_party(module)

            for (k, v1), v2 in imports.items():
                if k is None:
                    if v1 and v1 != module:
                        s += f"import {module} as {v1}\n"
                    else:
                        s += f"import {module}\n"
                else:
                    if k != v2 or v1:
                        items.append(f"{k} as {v2}")
                    else:
                        items.append(k)
            
            items = sorted(items)
            if items:
                items_v0 = ", ".join(items)
                items_v0 = f"from {module} import {items_v0}\n"
                items_v1 = "(\n    " + ",\n    ".join(items) + "\n)"
                items_v1 = f"from {module} import {items_v1}\n"
                s += items_v0 if len(items_v0) <= 70 else items_v1
            
            if party != last_party:
                s += "\n\n"
                last_party = party

        s += "\n"
        s += self.put_abstract_enum_class()

        # Append the main generated stub
        s += self.output

        return s.rstrip() + "\n"
```

I am not sure if this is cleaner though

PR #467, opened 2024-03-11
Allow `ndarray_check` on objects of arbitrary modules

njroussel:
This PR changes `ndarray_check` such that it returns `True` if one of the following conditions is met:

* The given object has a `__dlpack__` attribute
* The given object supports the CPython buffer interface
* The given object is an object from NumPy (implicit through buffer protocol), PyTorch, XLA, TensorFlow

With this change, `ndarray_check` more closely matches the logic of supported imports through `ndarray_import`.

I've added a straightforward test which checks this new behavior on an object with a `__dlpack__` attribute.

### DIFF
--- a/docs/api_extra.rst
+++ b/docs/api_extra.rst
@@ -541,8 +541,11 @@ section <ndarrays>`.
 
 .. cpp:function:: bool ndarray_check(handle h) noexcept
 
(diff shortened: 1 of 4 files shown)


wjakob:
Thanks!

PR #472, opened 2024-03-12
Fix reference leak when adding a new function overload

oremanj:
Noticed this while reading through nb_func.cpp. I verified that the new test fails if run without the new `Py_CLEAR`.


oremanj:
Updated the test to not check the numeric refcount on PyPy.

wjakob:
Fair enough.

PR #473, opened 2024-03-12
Support customizing __new__

oremanj:
This was sort of already possible, but quite cumbersome, because:
* One needed to combine `.def_static("__new__", ...)` and `.def("__init__", ...)` with the same signature, and add a passthrough lambda for `__new__` that would discard the Python type object argument.
* Unpickling would break, since the custom `__new__` would hide the nanobind internal `__new__` with no public way to refer to the original. This could be ameliorated somewhat using `__getnewargs_ex__` but not while maintaining compatibility with existing pickles.

This PR fixes both issues, and adds detailed documentation of this feature. It should be helpful for people who are trying to figure out how to port pybind11 constructs like `.def(py::init([](...) -> std::shared_ptr<T> { ... }))`.

libnanobind.a size cost: 284 bytes


docs/classes.rst
```diff
@@ -1012,4 +1012,87 @@ custom ``__init__``-style constructors.
...
+produce Python object *initializers*. nanobind augments these with its own
+Python object constructor, which allocates a Python object that has
+space in its memory layout for the C++ object to slot in.
+The ``__init__`` method then fills in that space by calling a C++ constructor.
+
+This split between ``__new__`` and ``__init__`` has a lot of benefits,
+including a reduction in unnecessary allocations, but it does mean that
+anything created from Python must be able to control where its
+C++ innards are stored. Sometimes, as with the example of ``Pet`` above,
+that's not feasible. In such cases, you can duck down a layer and override
```
wjakob, line 1057: the expression "duck down" sounded odd to me (I was thinking that you are making some kind of reference to duck typing), perhaps "go down one level" or something of that sort?

docs/classes.rst
```diff
@@ -1012,4 +1012,87 @@ custom ``__init__``-style constructors.
...
+  of any other function you might write bindings for. In particular,
+  you can pass a :cpp:enum:`nb::rv_policy <rv_policy>` as an additional
+  argument to :cpp:func:`.def() <class_::def>` to control how this conversion
+  occurs.
+
+* A ``__init__`` that takes the same arguments as ``__new__`` but
+  performs no operation.  This is necessary because Python automatically
+  calls ``__init__`` on the object returned by ``__new__`` in most cases.
+
+If you need even more control, such as because you need to access the
```
wjakob, line 1085: "such as because" -> "perhaps because" or similar rephrasing

docs/classes.rst
```diff
@@ -1012,4 +1012,87 @@ custom ``__init__``-style constructors.
...
+type object that Python passes as the first argument of ``__new__``
+(which :cpp:struct:`nb::new_ <new_>` discards), you can write a
+``.def_static("__new__", ...)`` and matching ``.def("__init__", ...)``
+yourself.
+
+.. note::
+   Even if you provide a custom ``__new__`` that takes additional arguments,
```
wjakob, line 1092: I think there needs to be a black line between the RST directive and the body.

include/nanobind/nb_class.h
```diff
@@ -376,6 +376,29 @@ template <typename Arg> struct init_implicit {
...
+    template <typename Class, typename... Extra>
+    NB_INLINE void execute(Class &cl, const Extra&... extra) {
+        using Type = typename Class::Type;
+        cl.def_static(
+            "__new__",
+            [func = (detail::forward_t<Func>) func](type_object_t<Type>, Args... args) {
+                return func((detail::forward_t<Args>) args...);
+            },
+            extra...);
+        cl.def("__init__", [](handle_t<Type>, Args...) {}, extra...);
```
wjakob, line 397: can this just take a `nb::handle` for ``self``? I don't see the harm, and this avoids a type check done by nanobind.

include/nanobind/nb_traits.h
```diff
@@ -101,6 +101,20 @@ struct analyze_method<Ret (Cls::*)(Args...) const noexcept> {
...
 };
 
+template <typename F>
+struct strip_function_object {
+    using type = typename analyze_method<decltype(&F::operator())>::func;
+};
+
+// Extracts the function signature from a function, function pointer or lambda.
+template <typename Function, typename F = std::remove_reference_t<Function>>
+using function_signature_t = std::conditional_t<
```
wjakob, line 111: How does this relate to the other commit that is modifying function signatures? Are there simplification opportunities in merging them?

include/nanobind/nb_class.h
```diff
@@ -376,6 +376,29 @@ template <typename Arg> struct init_implicit {
...
+    std::remove_reference_t<Func> func;
+
+    new_(Func &&f) : func((detail::forward_t<Func>) f) {}
+
+    template <typename Class, typename... Extra>
+    NB_INLINE void execute(Class &cl, const Extra&... extra) {
+        using Type = typename Class::Type;
+        cl.def_static(
+            "__new__",
+            [func = (detail::forward_t<Func>) func](type_object_t<Type>, Args... args) {
```
wjakob, line 393: Do you need to take `type_object_t` as first argument? You aren't really using this argument at all, and this will add an extra check done by nanobind. The ``__new__`` signature is not added to the stubs.

wjakob (commented):
Some detailed comments and a more high-level one:

1. It seems that overriding ``__new__`` doesn't compose with the usual system of having multiple overloads. Shouldn't it? If not, why?

2. Can you explain what a "nullary" ``__new__`` is and why a special code path is needed in `nb_func.cpp`? FWIW pickling is considered a use case that is worth supporting, but not necessarily by adding code to libnanobind beyond the bare basics. Although I don't understand the nitty gritty details, my feeling is that `__new__` ought to support overloads, and then the user can create this "nullary" constructor on their end.

oremanj:
> It seems that overriding __new__ doesn't compose with the usual system of having multiple overloads. Shouldn't it? If not, why?

It should work fine to provide multiple overloads using `nb::new_`; if that wasn't clear from the docs, maybe I should provide an example of that?

> Can you explain what a "nullary" `__new__` is and why a special code path is needed in nb_func.cpp?

"Nullary" just means that it takes no arguments (well, none beyond the type object). The problem is that there are two competing desires for how `__new__` is to be used:

* Unpickling wants to do `__new__` + `__setstate__`. In nanobind, `__setstate__` is a constructor, so that means `__new__` must leave the C++ portion of the object uninitialized.
* The user who wrote a custom `__new__` wants most uses of it to produce a Python object whose C++ portion is not uninitialized.

I was trying to distinguish between these situations by letting the no-argument `__new__` produce an object with uninitialized C++ portion (just like it would without the `__new__` customization), while the user's custom `__new__` presumably takes arguments and can return a fully initialized object. If the user provides a custom `__new__` that takes no arguments, then they're creating an inherent conflict with unpickling and are on their own.

> FWIW pickling is considered a use case that is worth supporting, but not necessarily by adding code to libnanobind beyond the bare basics. Although I don't understand the nitty gritty details, my feeling is that `__new__` ought to support overloads, and then the user can create this "nullary" constructor on their end.

I would love to write it that way if possible! Unfortunately, pretty much everything the user (or the `nb::new_::execute()` function) would need to do this themselves is private to libnanobind. After writing their own `__new__` to the class, the only way they can cause `inst_new_int` to be called is by saving the old `tp_new` pointer (from before their `__new__` assignment) and calling it with the appropriate convention. `nb::new_` couldn't even easily tell whether the incoming `__new__` is overwriting the stock nanobind one, versus overloading with another custom `__new__`, because the `nb_func` type is also private. I guess it could do a string comparison on the type name? But that felt awkward enough that it seemed warranted to provide some help from libnanobind.

If you'd rather see an approach that doesn't need any libnanobind changes, I can write one; it will just be pretty awkward in expectation.

wjakob:
Can you tell me the flaw in the following reasoning? Usually, the construction sequence is that nanobind will allocate an object (still in invalid state), and then run a constructor (`__init__` or `__setstate__`) that transitions the instance to valid state. `nb::new_` would provide a new strategy to create the initial object, but I presume it would still be in "invalid state" even though it is already constructed. A subsequent dummy call to `__init__` then transitions the instance to a valid state.

For unpickling, I would expect this to work similarly. The user could bind a nullary ``nb::new_([]{return new Inst(); }) which would allocate space for a default-constructed instance. To support unpickling for such an object, one would then need to run an in-place destructor (given that `new` already ran an actual constructor) and subsequently reconstruct the object with the pickle data, e.g.:

```cpp
cls.def("__setstate__", [](Inst *i, std::pair<int, float> value) { i->~Inst(); new (inst) Inst(value.first, value.second); })
```

These things could be wrapped up in a nice way without impacting `libnanobind.a`. 

The `new_` construction method looks like a nice addition, but the interaction with pickling is a more advanced use case where several special requirements coincide to create a situation that's likely out of scope for the larger nanobind project. Perhaps we can find a way to support `nb::new_` _and_ unpickling in your codebase with minimal additions that can be maintained outside of the project (e.g. a helper function to create the ``_setstate__`` construction above).

oremanj:
> nb::new_ would provide a new strategy to create the initial object, but I presume it would still be in "invalid state" even though it is already constructed. A subsequent dummy call to __init__ then transitions the instance to a valid state.

This is fine if our conception of "validity" is entirely theoretical, but I do think the object returned by `new_` needs to have its `nb_inst::ready` flag set immediately, because someone could call `__new__` without `__init__` and that shouldn't be allowed to cause a C++-side leak. (Also, there's no practical way to _not_ set the `ready` flag; it's set by `nb_type_put` in the type caster for `__new__`'s return value, which occurs after anything that the user's custom `__new__` can control.)

> To support unpickling for such an object, one would then need to run an in-place destructor (given that new already ran an actual constructor) and subsequently reconstruct the object with the pickle data, e.g.:

That's an option I hadn't considered, which I suppose is workable except that nanobind won't allow `__setstate__` to be invoked if the object is already initialized (because it's considered a constructor). How would you propose getting around that?

I'll try and figure out an approach that doesn't break pickling and doesn't rely on a libnanobind change.

docs/classes.rst
oremanj: It looks fine locally. I don't think there even needs to be a newline; I've written `.. note:: words words words` in other docs before.

oremanj:
D'oh, of course, `inst_new_int` *is* (effectively) exposed as `nb::inst_alloc()`. I can just use that. Update forthcoming.

include/nanobind/nb_class.h
oremanj: Yep, I was trying to get the signature correct but as you point out it doesn't matter.

oremanj:
Please take another look; this is now implemented without modifications to libnanobind.

wjakob:
This is great and a big improvement.

Can you include an example with overloads? Regarding the RST bit, the [spec](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives) says:

> Arguments and options must form a contiguous block beginning on the first or second line of the directive; a blank line indicates the beginning of the directive content block.

All examples there also use the convention with the intermediate newline. The Sphinx RST parser may be more forgiving, but this may then creates issues with syntax highlighting in editors or other RST parsers (e.g. the GitHub builtin web interface). Better to stick to the spec.

wjakob:
Would it not be more logical to factor out the nullary constructor into a separate construction? I think that some users may not actually _want_ this feature.

wjakob:
e.g. `class.def(nb::new())` without arguments

include/nanobind/nb_class.h
```diff
@@ -376,6 +376,46 @@ template <typename Arg> struct init_implicit {
...
+            // which would make it difficult to support pickling for this type.
+            // To maximize flexibility here, we provide a cpp_function __new__
+            // (which can be overloaded) to wrap the slot __new__ (which can't).
+            // If the incoming __new__ has no arguments, then this overload
+            // would be ambiguous with the user-supplied one, so we don't add
+            // it; the user will have to figure out something else for pickle.
+            cl.def_static("__new__",
+                          [](handle type) {
+                              if (!type_check(type))
+                                  detail::raise_cast_error();
```
wjakob, line 404: perhaps a type error is more appropriate?

oremanj: This is what you hit if you write e.g. `t.UniqueInt.__new__(int)`. I think someone who gets here probably knows what they did wrong; my main priority is that it shouldn't crash the interpreter. Calling `raise_cast_error` produces more compact code than constructing a new exception, so I went with that.

oremanj:
> Would it not be more logical to factor out the nullary constructor into a separate construction? I think that some users may not actually want this feature.

I guess I'm not seeing how it would harm someone who doesn't need it. There's no corresponding nullary `__init__`, so writing `SomeType()` still produces a type error.
```
>>> t.UniqueInt()
TypeError: __init__(): incompatible function arguments. The following argument types are supported:
    1. __init__(self, arg: int, /) -> None

Invoked with types: test_classes_ext.UniqueInt

>>> t.UniqueInt("hello")
TypeError: __new__(): incompatible function arguments. The following argument types are supported:
    1. __new__(arg: object, /) -> object
    2. __new__(arg0: object, arg1: int, /) -> test_classes_ext.UniqueInt

Invoked with types: nanobind.nb_type_0, str
```
Meanwhile factoring it out has costs of its own: we have to document that `new_` can be used in these two different ways, and users who don't pick up on the subtlety will have a lot of confusion if they use pickling. Is there some cost or hazard I'm missing here, beyond the presumably-small increment to compilation time and binary size from the extra overload?

oremanj:
Updated the examples, factored out the non-dependent parts into a separate inline function, and added enforcement that any zero-arg `new_` is first in the overload set (so we can know not to create the wrapper). I think this is ready.

tests/test_classes.py
```diff
@@ -825,3 +825,42 @@ def test45_hidden_base():
     assert s.value == 20
     assert s.get_answer() == 200
     assert s.polymorphic() == 20
+
+
+def test46_custom_new():
```
wjakob, line 830: Could you add an example with multiple ``__new__`` overloads to the test suite?

include/nanobind/nb_class.h
```diff
@@ -508,6 +562,12 @@ class class_ : public object {
         return *this;
     }
 
+    template <typename Func, typename... Extra>
+    NB_INLINE class_ &def(new_<Func> &&init, const Extra &... extra) {
```
wjakob, line 566: could the ``init`` argument just be called ``arg``? Let's also make this change for the `init` and `init_implicit` constructor.

docs/api_core.rst
```diff
@@ -1989,6 +1989,13 @@ Class binding
       This constructor generates more compact code than a separate call to
       :cpp:func:`implicitly_convertible`, but is otherwise equivalent.
 
+   .. cpp:function:: template <typename Func, typename... Extra> class_ &def(new_<Func> new_, const Extra &... extra)
```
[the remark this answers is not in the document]
wjakob: Can you make a similar change also for the `init` and `init_implicit` methods above? (The same change is requested below for the C++ bits).

wjakob (commented):
Miscellaneous minor requests

wjakob:
Perfect, thank you!

PR #474, opened 2024-03-12
RFC: low-level API for changing the C++ object managed by a Python object

oremanj:
I'm submitting this initially without tests and with minimal documentation because I want to know if you like the direction before I invest more time in it.

This extends the low-level API with two functions that allow the creation of nanobind instances that can manage different heap-allocated C++ objects over the course of a single Python object's lifetime:
* `inst_alloc_indirect()` creates an instance that can participate in this system, which initially manages no C++ object (`inst_ptr` returns null for it)
* `inst_reset()` changes the C++ object associated with an instance, which must have originally been created by `inst_alloc_indirect()`. The new object can be either owned or referenced. The previously-associated C++ object is destroyed if it had been advertised as owned in the previous `inst_reset()` call. keep_alive callbacks are also cleared by `inst_reset()` in order to be compatible with shared ownership techniques such as are used by the `shared_ptr` type caster.

I think this is the smallest coherent core change that would allow me to handle some tricky situations I'm running into. (Illustrative example: support unpickling for a class that can only be constructed on the heap.) It might be possible to make smaller, but at the cost of being more specialized to my specific problem.

Size cost to libnanobind is currently 472 bytes text + 128 bytes rodata. (I'm not sure where the latter is coming from, so it might be something easily fixable.)

### DIFF
--- a/include/nanobind/nb_class.h
+++ b/include/nanobind/nb_class.h
@@ -284,6 +284,9 @@ inline object inst_alloc(handle h) {
 inline object inst_alloc_zero(handle h) {
     return steal(detail::nb_inst_alloc_zero((PyTypeObject *) h.ptr()));
 }
+inline object inst_alloc_indirect(handle h) {
+    return steal(detail::nb_inst_alloc_indirect((PyTypeObject *) h.ptr()));
+}
 inline object inst_take_ownership(handle h, void *p) {
     return steal(detail::nb_inst_take_ownership((PyTypeObject *) h.ptr(), p));
 }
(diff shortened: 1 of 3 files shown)


wjakob:
As-is, this is too intrusive and specific of a feature for me to consider, sorry. Splitting up functions on the hot path like `inst_new_ext` or `inst_dealloc` also comes with a tiny but nonzero cost, which I would not want to pay for.

oremanj:
All right, thanks for giving it a look!

PR #475, opened 2024-03-13
Rename `forward_like` to avoid conflict with STL

yosh-matsuda:
When nanobind user code enables C++23, I found that `nanobind::detail::forward_like` has a naming conflict with  `std::forward_like` in STL due to ADL.

The situation is like [this](https://godbolt.org/z/98xP78v7o):

```c++
#include <utility>

namespace nanobind {
    template <typename T, typename U>
    auto forward_like(U&& t) { return t; }

    template <typename T>
    auto func(T&& t) { return forward_like<T>(t); }
}

int main() {
    auto p = std::pair(0, 0);
    nanobind::func(p);
    return 0;
}
```

```
x86-64 clang 17.0.1
x86-64 clang 17.0.1
-std=c++23 -stdlib=libc++ -Wall -Wextra 
123
<Compilation failed>

# For more information see the output window
x86-64 clang 17.0.1 - cached
<source>:8:31: error: call to 'forward_like' is ambiguous
    8 |     auto func(T&& t) { return forward_like<T>(t); }
      |                               ^~~~~~~~~~~~~~~
```

In particular, this will potentially occur when casting std object to a Python object using the nanobind STL caster.  
The simplest solution is to rename `forward_like` (in this PR). We can also avoid ADL in other ways.

### DIFF
--- a/include/nanobind/nb_traits.h
+++ b/include/nanobind/nb_traits.h
@@ -55,7 +55,7 @@ using forwarded_type = std::conditional_t<std::is_lvalue_reference_v<T>,
 
 /// Forwards a value U as rvalue or lvalue according to whether T is rvalue or lvalue; typically
 /// used for forwarding a container's elements.
-template <typename T, typename U> NB_INLINE forwarded_type<T, U> forward_like(U &&u) {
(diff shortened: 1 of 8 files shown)


wjakob:
Fair enough, thanks!

PR #480, opened 2024-03-17
Fix problem where `std::complex` caster always succeeded

### LINKED ISSUE #479: [BUG]: Complex type does not work in variant
### Problem description

If `std::complex<float>` is included in a variant then it causes the variant to improperly detect non complex types as complex (see example below).

### Reproducible example code

```c++
#include <iostream>
#include <nanobind/stl/complex.h>                                              
#include <nanobind/stl/variant.h>                                              

namespace nb = nanobind;                                                       

using MyType = std::variant<
    nb::float_,
    std::complex<float>>;

void print_type(MyType v) {
  if (auto pv = std::get_if<std::complex<float>>(&v); pv) {                    
    std::cout << "GOT COMPLEX " << std::endl;                                  
  } else if (auto pv = std::get_if<nb::float_>(&v); pv) {                      
    std::cout << "GOT FLOAT" << std::endl;                                     
  }                                                                            
}

NB_MODULE(test, m) {
    m.def("print_type", &print_type);                                          
}
```

And in Python:

```python
import test

# Correctly routes to float
test.print_type(1.0)

# Incorrectly routes to cmplex
test.print_type(list())
```

yosh-matsuda:
Fix #479

### DIFF
--- a/include/nanobind/stl/complex.h
+++ b/include/nanobind/stl/complex.h
@@ -55,7 +55,7 @@ template <typename T> struct type_caster<std::complex<T>> {
             return true;
         }
 
-        return true;
+        return false;
     }
 
     template <typename T2>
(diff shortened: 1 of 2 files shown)


wjakob:
Oh, that's very bad. Thank you for catching the issue.

PR #482, opened 2024-03-18
Allow custom dying words in fail_unspecified().

hpkfft:
It seems impolite for an extension library developer (such as myself) to have an error message that would likely direct problems with a particular extension to nanobind rather than directly to the extension developer.
This commit allows the fail message to be customized by setting a CMake cache variable.

Since the default message is giving CMake build mode advice, I think it makes sense for this string to be initialized in the CMake layer anyway.   If someone were to support an alternate build tool/environment, they would want to provide a different message.

With this commit, the default message is:
```
Critical nanobind error: encountered an unrecoverable error condition.
Recompile using the 'Debug' or 'RelWithDebInfo' modes to obtain further information about this problem.
Aborted
```

An extension can, for example, do the following before configuring nanobind:
```
set(NB_COMPACT_ASSERTION_MESSAGE
    "For help regarding ${CMAKE_PROJECT_NAME}, please visit ${CMAKE_PROJECT_HOMEPAGE_URL}"
    CACHE STRING "Printed in Release builds if internal checks fail")
```
In my project, a failure then would print the following:
```
Critical nanobind error: encountered an unrecoverable error condition.
For help regarding hpk.fft, please visit https://hpkfft.com
Aborted
```

If you like this idea, I will add something to the docs in a separate PR in a week or two.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,11 @@ if (NB_MASTER_PROJECT AND NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES
     "MinSizeRel" "RelWithDebInfo")
 endif()
 
+set(NB_COMPACT_ASSERTION_MESSAGE
+    "Recompile using the 'Debug' or 'RelWithDebInfo' modes to obtain further information about this problem."
+    CACHE STRING
+    "Printed in Release or MinSizeRel builds if internal checks fail")
+
 # ---------------------------------------------------------------------------
 # Check whether all dependencies are present
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


wjakob:
By the way, your hpkfft project looks very cool and something I'd love to use as an alternative to FFTW. But it is closed source and will continue to be so, correct?

hpkfft:
Thanks for the kind words regarding hpkfft.  It is closed source now, and although that may change in the future, you should not count on it becoming open source.

PR #484 is offered as an alternative to this, and, I think, does as you suggest.  If you merge either, please reject the other one.

I would like to mention that the CMake-related footprint of this PR is only relevant to CMake.   Ultimately, CMake only adds the compilation flag `-DNB_COMPACT_ASSERTION_MESSAGE=....`
Other build systems (e.g., bazel) can ignore this entirely.   If they do, the output will be:
```
Critical nanobind error: encountered an unrecoverable error condition.

Aborted
```
This may be preferable to PR #484, which would print CMake-specific advice using words like `RelWithDebInfo` that I assume do not apply in a bazel environment.

If either PR is accepted, the bazel maintainers are advised to provide a default string for `-DNB_COMPACT_ASSERTION_MESSAGE=....`
If neither PR is accepted, the message cannot be changed via macro definition.

wjakob:
See 94d09cacbe594698d6f2d1daf728e239a34075b7.

PR #485, opened 2024-03-19
Decouple complex ndarray from stl/complex.h.

hpkfft:
As an extension developer using complex numbers and having custom bindings, I do not want to use a type caster for std::complex.  However, I do want to use nb::ndarray for arrays of std::complex.

This PR adds `is_complex_v<T>` to nb_traits.h so that it is not necessary to include nanobind/stl/complex.h in order to use nb::ndarray with complex numbers.  Note that the standard library `<complex>` is **not** included so that those not using complex numbers do not pay the cost of its inclusion.


wjakob:
LGTM, thanks.

PR #489, opened 2024-03-22
ndarray import from buffer protocol requires integer stride.

hpkfft:
In the numpy buffer protocol, strides are measured in bytes.  If a stride (in any dimension) is not an integer multiple of itemsize, then the stride cannot be correctly converted to number of elements.

With this PR, a function taking such an nb::ndarray argument will raise a TypeError exception.


wjakob:
This looks great, thank you! No worries about the failing tests.

PR #492, opened 2024-03-23
enum metaclass methods

rzhikharevich:
This PR implements the following metaclass methods for enum types: `__iter__`, `__contains__`, `__getitem__`, `__len__` and `__reversed__`, similar to those of standard Python enums.

Unfortunately, this also reintroduces entanglement between `nb_type.cpp` and `nb_enum.cpp`, so I'm marking this as a draft for now, but posting to start the discussion of possible implementations. To me it seems like keeping them detangled would require adding some kind of metaclass API, but maybe there is a more simple way?

I've also noticed that `stubgen.py` crashes when it encounters an empty enum due to them not having an `@entries` attribute and therefore not being recognized as an enum. Should we perhaps always create this attribute? Or handle `__doc__` more carefully in `stubgen.py`?

```
[27/29] Generating test_enum_ext.pyi
FAILED: tests/test_enum_ext.pyi /home/roman/projects/nanobind/build/tests/test_enum_ext.pyi

cd /home/roman/projects/nanobind/build/tests && /home/roman/venv/master-3.12/bin/python3 /home/roman/projects/nanobind/src/stubgen.py -q -i /home/roman/projects/nanobind/build/tests -m test_enum_ext -o test_enum_ext.pyi
  File "/home/roman/projects/nanobind/src/stubgen.py", line 1463, in <module>
    main()
  File "/home/roman/projects/nanobind/src/stubgen.py", line 1434, in main
    sg.put(mod_imported)
  File "/home/roman/projects/nanobind/src/stubgen.py", line 838, in put
    self.put(child, name=name, parent=value)
  File "/home/roman/projects/nanobind/src/stubgen.py", line 845, in put
    self.put_type(value, name)
  File "/home/roman/projects/nanobind/src/stubgen.py", line 544, in put_type
    self.put_docstr(docstr)
  File "/home/roman/projects/nanobind/src/stubgen.py", line 283, in put_docstr
    docstr = textwrap.dedent(docstr).strip()
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/roman/.pyenv/versions/3.12.2/lib/python3.12/textwrap.py", line 435, in dedent
    text = _whitespace_only_re.sub('', text)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'getset_descriptor'
```

### DIFF
--- a/include/nanobind/nb_class.h
+++ b/include/nanobind/nb_class.h
@@ -86,7 +86,11 @@ enum class type_init_flags : uint32_t {
     /// and/or 'type_slots_callback' members of type_init_data
     has_type_slots           = (1 << 23),
 
-    all_init_flags           = (0x1f << 19)
+    // This type provides extra metaclass type slots via the 'metaclass_type_slots' member of
+    // type_init_data, as well as a custom metaclass name via 'metaclass_name'.
+    has_metaclass_type_slots = (1 << 24),
+
+    all_init_flags           = (0x3f << 19)
 };
 
 // See internals.h
@@ -96,7 +100,7 @@ struct nb_alias_chain;
 struct type_data {
     uint32_t size;
     uint32_t align : 8;
-    uint32_t flags : 24;
+    uint32_t flags : 25;
     const char *name;
     const std::type_info *type;
     nb_alias_chain *alias_chain;
@@ -123,6 +127,8 @@ struct type_init_data : type_data {
     const PyType_Slot *type_slots;
     void (*type_slots_callback)(const type_init_data *d, PyType_Slot *&slots, size_t max_slots);
     size_t supplement;
+    const PyType_Slot *metaclass_type_slots;
+    const char *metaclass_name;
 };
 
 NB_INLINE void type_extra_apply(type_init_data &t, const handle &h) {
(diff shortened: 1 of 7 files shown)


wjakob:
cc @oremanj

wjakob:
Having a dependence from `nb_type` -> `nb_enum` is something I feared and would like to avoid.

How about giving `nb_type_tp` the ability to add custom slots? Rather than exposing this externally to the user, the enum constructor could internally use this to create an enums-specific metaclass that adds the functions you listed above.

Also, one stupid question: why are all these sequence-related methods needed? (``__reversed__``, ``__iter__``), etc. Isn't any object with ``__len__`` and ``__getitem__`` implicitly reversible and iterable? The smaller this all can be, the better.

wjakob:
Any thoughts on this @rzhikharevich? Note that if API-breaking changes are needed, then it would be good to still include them in 2.0.0. (And that window of opportunity will be closing in the next weeks.)

rzhikharevich:
> How about giving nb_type_tp the ability to add custom slots? Rather than exposing this externally to the user, the enum constructor could internally use this to create an enums-specific metaclass that adds the functions you listed above.

Sounds good to me, I'll try to implement this in the next few days.

> Isn't any object with __len__ and __getitem__ implicitly reversible and iterable?

I think the problem is that we define `mp_subscript` (accepts an object) instead of `sq_item` (accepts an index), so it is not automatically reversible / iterable. I'll double-check this later.

rzhikharevich:
@wjakob I've updated the PR to set up the metaclass through fields in `type_init_data`: `metaclass_type_slots` for extra slots and `metaclass_name` for setting a custom metaclass name (as well as using it as a key in the metaclass dictionary).

wjakob:
closing in favor of #533

PR #496, opened 2024-03-28
Fix possible ValueError during stubgen

laggykiller:
Fixes https://github.com/wjakob/nanobind/issues/420#issuecomment-2023992428


wjakob:
Thanks!

PR #501, opened 2024-04-02
use qualified name when aliasing types

skallweitNV:
Use qualified name when writing type aliases.

If you have some bindings like this:
```c++
nb::class_<A> a(m, "A");
nb::enum_<E> e(a, "E");
nb::class_<B> b(m, "B");
b.attr("E") = a.attr("E");
```

you currently get:

```python
class A:
    class E(_Enum):
        ...

class B:
    E: TypeAlias = E  
```

with this fix you get:
```python
class B:
    E: TypeAlias = A.E
```


wjakob:
LGTM, thanks!

PR #509, opened 2024-04-08
Use a checked cast() in accessor cast operator.

hawkinsp:
Currently the following code pattern performs an unchecked cast:
```
nb::object o = ...;
nb::list l = o.attr("a");
```

The type of the attribute is not checked, which leads to surprising errors when the type of the attribute is something else, e.g., a tuple.

It seems like a footgun to have implicit unchecked casts; casts should either be implicit or unchecked but not both.


wjakob:
This restriction of this change to accessors seems somewhat arbitrary. Based on the logic underlying this change
```
nb::object o = ...;
nb::any_other_wrapper_type o2 = o;
```
should also perform a check. While I would not want to pay the cost of such extra checks, I am open to doing so in debug builds.

So in summary: I would be open to a PR (assuming it doesn't add an unreasonable amount of code) that checks for wrapper cast errors in a more general sense.

hawkinsp:
Hmmm. This is definitely something I don't want to do accidentally, and currently it's very easy to do by mistake.

If the cost of doing these checks is too high, would you be open to marking the operator as `explicit`, at least for subclasses of object if not `object` itself? This would be a breaking change to client code.

wjakob:
What's wrong with adding such instrumentation in debug builds?

hawkinsp:
Nothing. But I'm wondering if I can do better and get more type safety in optimized builds, also.

If you require an explicit cast, then the user can either `static_cast<T>` if they don't want checking, or `nanobind::cast<T>`, if they do.

wjakob:
I took another look at this commit and will close it for now. As a strategy for checking implicit casts, it is not general enough since it only applies to accessors (e.g. dictionary lookups), which seems confusing if people rely on the library to provide safety guarantees. The other point is that, I'd really prefer if users can opt out of these checks, which come at a cost.

Issue #517, opened 2024-04-11
[BUG]: importing my_ext fails with "ModuleNotFoundError: No module named 'my_ext'" on Windows 


nurpax:
### Problem description

Following the docs in https://nanobind.readthedocs.io/en/latest/basics.html, I created a directory with the following files:

```
CMakeLists.txt
my_ext.cpp
```

I followed the docs to build it.  It builds fine but import fails:

```
C:\Users\janne\dev\nbtest>cmake -S . -B build
C:\Users\janne\dev\nbtest>cmake --build build
C:\Users\janne\dev\nbtest>cd build

C:\Users\janne\dev\nbtest\build>python
Python 3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import my_ext
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'my_ext'```
```

However, it works if I cd into `build\Debug` instead:

```
C:\Users\janne\dev\nbtest\build>cd debug

C:\Users\janne\dev\nbtest\build\Debug>python
Python 3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import my_ext
>>> my_ext.add(3,5)
8
>>>
```

Not sure if this happens only on Windows but it wasn't immediately obvious what the problem is.

Perhaps this could be called out in the docs?

### Reproducible example code

```text
CMakeLists.txt:


cmake_minimum_required(VERSION 3.15...3.27)
project(my_ext) # Replace 'my_project' with the name of your project
find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)

if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
  set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
endif()

# Detect the installed nanobind package and import it into CMake
execute_process(
  COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
[...]

PR #521, opened 2024-04-12
add a note about running on Windows

nurpax:
The cmake outputs for the built extensions go under build\Debug so the run instructions are not the same as on other OSs.  In addition, python is just `python.exe` on Windows, not `python3`.

for #517


wjakob:
Thanks!

PR #531, opened 2024-04-16
Swap target/source in implicitly_convertible example code

nurpax:
Match the order of `implicitly_convertible()` `Source` and `Target` template parameters in the example code.

Here's a screenshot of the docs for context:

image


wjakob:
Thanks!

PR #533, opened 2024-04-17
Switch ``nb::enum_`` to Python's builtin ``enum.*`` package

wjakob:
nanobind's use of custom enumeration types has caused some friction in the past. This experimental commit replaces them with builtin Python types (in particular, ``enum.Enum`` and ``enum.IntEnum``).

The implementation uses two internal maps to accelerate value->element and element->value queries. Another detail is that ``.value()`` extends a partially built enumeration on the fly, which requires access to some ``enum.Enum`` internals. This resembles ``aenum.extend_enum`` by Ethan Furman (who also developed the ``enum.*`` classes).


wjakob:
An update: I deployed this change in one of my main nanobind-based binding project, and it worked fine. I will merge this for now but am still happy to discuss the change further in case anyone runs into issues.

hpkfft:
Is it possible to write a custom comparison for enum?  I had been using `nb::type_slots()` to accomplish this.

Alternatively, might you be willing to implement [enum.Flag](https://docs.python.org/3/library/enum.html#enum.Flag)

I'm not sure what is the most pythonic way to allow my users to query whether a given "thing" is present in an enum.
I had been writing my own comparison so that `lhs >= rhs` means "`lhs` is a superset of the bits set in `rhs`".
That is, the following, after casting to integers:  `(lhs & rhs) == rhs`

I can also use `__contains__`.   Maybe that's actually better:
```
from enum import Flag, auto
class Color(Flag):
    RED = auto()
    GREEN = auto()
    BLUE = auto()

purple = Color.RED | Color.BLUE
white = Color.RED | Color.GREEN | Color.BLUE

Color.GREEN in purple  # False
Color.GREEN in white   # True
purple in white    # True
white in purple    # False
```

I'm a bit new to python--I want to do whatever seems most natural to the users of my extension.
Any advice/help you can offer would be appreciated.

PR #551, opened 2024-04-28
Fix building tests with Ninja on Windows

patstew:
Putting outputs in Release/ folders etc is a result of using a multi-config cmake generator like Visual Studio, not of using MSVC. This fixes using ninja on windows, and presumably using multi config generators with compilers other than MSVC.


wjakob:
Thanks!

PR #555, opened 2024-04-30
Fix visibility for GCC on Windows (mingw64)

lkeegan:
- revert change from https://github.com/wjakob/nanobind/pull/440
  - `dllexport` was/is correct on Windows, including with mingw64 gcc compiler
- don't set `visibility("hidden")` attribute for namespaces on windows
  - this resolves the original mingw64 compilation error seen in #440

### DIFF
--- a/include/nanobind/nb_defs.h
+++ b/include/nanobind/nb_defs.h
@@ -22,7 +22,7 @@
 #  define NAMESPACE_END(name) }
 #endif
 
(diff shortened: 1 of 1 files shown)


wjakob:
cc @laggykiller  -- do you have any feedback on this change?

laggykiller:
@wjakob Not expert, but looks good to me

wjakob:
Ok, thanks! -- merged.

PR #562, opened 2024-05-05
add support for std::wstring following std::string

kayama-shift:
In some cases, wchar_t strings are preferable to UTF-8 strings for non-ascii string.
So add support for std::wstring following std::string using provided Python API.

### DIFF
--- a/include/nanobind/stl/wstring.h
+++ b/include/nanobind/stl/wstring.h
@@ -1,0 +1,39 @@
+/*
+    nanobind/stl/string.h: type caster for std::string
+
+    Copyright (c) 2022 Wenzel Jakob
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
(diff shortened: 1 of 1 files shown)


wjakob:
The caster looks good, but this PR will need test coverage to be considered for inclusion.

wjakob:
(nvm, I will ad them myself)

valgur:
Doesn't the `PyUnicode_AsWideCharString(src.ptr(), &size)` call in this conversion leak memory?

> PyUnicode_AsWideCharString returns a buffer allocated by [PyMem_New](https://docs.python.org/3/c-api/memory.html#c.PyMem_New) (use [PyMem_Free()](https://docs.python.org/3/c-api/memory.html#c.PyMem_Free) to free it) on success. On error, returns NULL and *size is undefined. Raises a [MemoryError](https://docs.python.org/3/library/exceptions.html#MemoryError) if memory allocation is failed.

For context, I'm writing a caster for std::u32string and std::u32_string_view types and was thinking of using this std::wstring caster as a reference.

PR #585, opened 2024-05-15
Update stubgen of enums to include literal values

Seairth:
This MR addresses the comment made [here](https://github.com/wjakob/nanobind/issues/420#issuecomment-2108221414).


Seairth:
Hmm.. one check failed, but it doesn't look related to my MR.

wjakob:
LGTM, thanks. That failure is something unrelated.

PR #592, opened 2024-05-23
Improve check_party for 1st party module

laggykiller:
This is the original `check_party()` in `stubgen.py`

```python
    def check_party(self, module: str) -> Literal[0, 1, 2]:
        """
        Check source of module
        0 = From stdlib
        1 = From 3rd party package
        2 = From the package being built
        """
        if module.startswith(".") or module == self.module.__name__.split('.')[0]:
            return 2

        try:
            spec = importlib.util.find_spec(module)
        except (ModuleNotFoundError, ValueError):
            return 1

        if spec:
            if spec.origin and "site-packages" in spec.origin:
                return 1
            else:
                return 0
        else:
            return 1
```

See https://github.com/wjakob/nanobind/issues/420#issuecomment-2043426430

Somehow `module` in `check_party` could be `my_sanitized_module.BASE` instead of `.my_sanitized_module` or `my_sanitized_module`. https://github.com/wjakob/nanobind/pull/496 only prevent crash due to `ValueError`, but all those modules are all misclassified as 3rd party modules instead of 1st party. This PR improve check for such condition.


wjakob:
Thanks!

PR #594, opened 2024-05-23
Add cupy support

JoOkuma:
Hi, this PR adds cupy support.

What additional tests do you recommend adding?

Do you think it would make sense for `np.ndarray` with `nb::device::cuda` be considered `cp.ndarray` automatically?

Reference: #510

### DIFF
--- a/docs/api_extra.rst
+++ b/docs/api_extra.rst
@@ -644,7 +644,7 @@ N-dimensional array type
 ------------------------
 
 The following type can be used to exchange n-dimension arrays with frameworks
-like NumPy, PyTorch, Tensorflow, JAX, and others. It requires an additional
+like NumPy, PyTorch, Tensorflow, JAX, CuPy, and others. It requires an additional
 include directive:
 
 .. code-block:: cpp
@@ -1005,6 +1005,8 @@ convert into an equivalent representation in one of the following frameworks:
 .. cpp:class:: pytorch
 
 .. cpp:class:: jax
+
+.. cpp:class:: cupy
 
 Eigen convenience type aliases
 ------------------------------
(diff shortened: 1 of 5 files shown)


wjakob:
This looks great. It's tricky to test these CUDA frameworks on GitHub actions. If it works for you, it is good enough for me. Please list the CuPy enumeration entry in the docs (both the ndarray overview file and the api_extra.rst file), then it's good to go.

JoOkuma:
Thanks, @wjakob; I updated the docs.

I'm getting an error on the cupy side in some environments, lets wait for me to understand this before merging this PR.

```python
E           RuntimeError: nanobind::detail::ndarray_wrap(): could not import ndarray: Traceback (most recent call last):
E             File "cupy/_core/dlpack.pyx", line 186, in cupy._core.dlpack.DLPackMemory.__init__
E           RuntimeError: CuPy is built against CUDA, different from the backend that backs the incoming DLPack tensor
```

wjakob:
Probably an attempt to convert a CPU dlpack tensor into a CuPy tensor. Perhaps they don't implement the copy back to the GPU.

JoOkuma:
You're rig. It does support a CPU dlpack.

And the error was on my side. I did not set the `nb::device::cuda::value` to the returning cuda array.
I think it's ready to merge.

Thanks!

wjakob:
> And the error was on my side. I did not set the nb::device::cuda::value to the returning cuda array.
> I think it's ready to merge.

Does this have any implications for the nanobind PR? I did not see further commits/amends since you raised this issue.

JoOkuma:
No, @wjakob. It worked once I fixed how I was using nanobind in my other codebase.

wjakob:
Got it. Thanks for the PR!

PR #595, opened 2024-05-23
Don't leak enum names.

hawkinsp:
The ->name field of the typeinfo is never freed at the moment.


wjakob:
Oops, you are right. Thanks!

PR #599, opened 2024-05-24
Add support for Flag enums

mwittgen:
With the major overhaul of enums in nanobind 2.0, the `Flag` enum type could be easily supported.
Some existing pybind11 applications rely on enums supporting bitwise operations with the result being an `enum` and not an `int`.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1930,9 +1930,9 @@ The following annotations can be specified using the variable-length
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
-   with two enumeration as operands.
-   The result will an enumeration of the same type.
-   So ``Shape(2) | Shape(1) -> Shepe(3)`.
+   with two with two enumerators as operands.
+   The result will have the same enumeration type as the operands.
+   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
 
 Function binding
 ----------------
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1926,7 +1926,7 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
-   .. cpp:struct:: is_flag_enum
+   .. cpp:struct:: is_flag
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1926,6 +1926,14 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
+   .. cpp:struct:: is_flag_enum
+
+   Indicate that the enumeration may be used with bitwise
+   operations.  This enables the bitwise operators ``| & ^ ~``
+   with two enumeration as operands.
+   The result will an enumeration of the same type.
+   So ``Shape(2) | Shape(1) -> Shepe(3)`.
+
 Function binding
 ----------------
 
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1930,7 +1930,7 @@ The following annotations can be specified using the variable-length
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
-   with two with two enumerators as operands.
+   with two enumerators as operands.
    The result will have the same enumeration type as the operands.
    So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
 
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1926,7 +1926,7 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
-   .. cpp:struct:: is_flag
+   .. cpp:struct:: flag_enum
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,8 +17,12 @@ below inherit that of the preceding release.
 
 Version 2.1.0 (TBA)
 -------------------
-* The :cpp:class:`nb::enum_\<T\>() <enum_>` can now create an``enum.Flag``-derived type
-  with  flag :cpp:class:`nb::is_flag_enum() <is_flag_enum>`.
+
+* nanobind now allows a :cpp:class:`nb::enum_\<T\>() <enum_>` to specify the
+  new class binding annotation :cpp:class:`nb::is_flag_enum() <is_flag_enum>`,
+  which produces an enumeration type derived from `enum.Flag`. Instances of such
+  an enumeration type represent a bitwise combination of the defined enumerators,
+  and they may be combined using bitwise operators ``& | ^ ~``.
 
 Version 2.0.1 (TBA)
 ---------------------------
@@ -143,11 +147,9 @@ according to `SemVer <http://semver.org>`__. The following changes are
 noteworthy:
 
 * The :cpp:class:`nb::enum_\<T\>() <enum_>` binding declaration is now a
-  wrapper that creates either a ``enum.Enum``, ``enum.IntEnum`` or ``enum.Flag``-derived type.
+  wrapper that creates either a ``enum.Enum`` or ``enum.Flag``-derived type.
   Previously, nanobind relied on a custom enumeration base class that was a
   frequent source of friction for users.
-  A new flag :cpp:class:`nb::is_flag_enum() <is_flag_enum>`
-  creates a ``enum.Flag``-derived type.
 
   This change may break code that casts entries to integers, which now only
   works for arithmetic (``enum.IntEnum``-derived) enumerations. Replace
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,11 @@ case, both modules must use the same nanobind ABI version, or they will be
 isolated from each other. Releases that don't explicitly mention an ABI version
 below inherit that of the preceding release.
 
+Version 2.1.0 (TBA)
+-------------------
+* The :cpp:class:`nb::enum_\<T\>() <enum_>` can now create an``enum.Flag``-derived type
+  with  flag :cpp:class:`nb::is_flag_enum() <is_flag_enum>`.
+
 Version 2.0.1 (TBA)
 ---------------------------
 
@@ -138,9 +143,11 @@ according to `SemVer <http://semver.org>`__. The following changes are
 noteworthy:
 
 * The :cpp:class:`nb::enum_\<T\>() <enum_>` binding declaration is now a
-  wrapper that creates either a ``enum.Enum`` or ``enum.IntEnum``-derived type.
+  wrapper that creates either a ``enum.Enum``, ``enum.IntEnum`` or ``enum.Flag``-derived type.
   Previously, nanobind relied on a custom enumeration base class that was a
   frequent source of friction for users.
+  A new flag :cpp:class:`nb::is_flag_enum() <is_flag_enum>`
+  creates a ``enum.Flag``-derived type.
 
   This change may break code that casts entries to integers, which now only
   works for arithmetic (``enum.IntEnum``-derived) enumerations. Replace
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,7 +19,7 @@ Version 2.1.0 (TBA)
 -------------------
 
 * nanobind now allows a :cpp:class:`nb::enum_\<T\>() <enum_>` to specify the
-  new class binding annotation :cpp:class:`nb::is_flag() <is_flag>`,
+  new class binding annotation :cpp:class:`nb::flag_enum() <flag_enum>`,
   which produces an enumeration type derived from `enum.IntFlag`. Instances of such
   an enumeration type represent a bitwise combination of the defined enumerators,
   and they may be combined using bitwise operators ``& | ^ ~``.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,8 +19,8 @@ Version 2.1.0 (TBA)
 -------------------
 
 * nanobind now allows a :cpp:class:`nb::enum_\<T\>() <enum_>` to specify the
-  new class binding annotation :cpp:class:`nb::is_flag_enum() <is_flag_enum>`,
-  which produces an enumeration type derived from `enum.Flag`. Instances of such
+  new class binding annotation :cpp:class:`nb::is_flag() <is_flag>`,
+  which produces an enumeration type derived from `enum.IntFlag`. Instances of such
   an enumeration type represent a bitwise combination of the defined enumerators,
   and they may be combined using bitwise operators ``& | ^ ~``.
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -147,7 +147,7 @@ according to `SemVer <http://semver.org>`__. The following changes are
 noteworthy:
 
 * The :cpp:class:`nb::enum_\<T\>() <enum_>` binding declaration is now a
-  wrapper that creates either a ``enum.Enum`` or ``enum.Flag``-derived type.
+  wrapper that creates either a ``enum.Enum`` or ``enum.IntEnum``-derived type.
   Previously, nanobind relied on a custom enumeration base class that was a
   frequent source of friction for users.
 
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -296,10 +296,12 @@ C++11-style strongly typed enumerations.
    When the annotation :cpp:class:`nb::is_arithmetic() <is_arithmetic>` is
    passed to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type
    will support arithmetic and bit-level operations (and, or,
-   xor, negation).
-   Passing the annotation :cpp:class:`nb::is_flag_enum() <is_flag_enum>` to
-   to :cpp:class:`nb::enum_\<T\> <enum_>`, will result in a Python type
-   `enum.Flags`, that supports bit operations without losing their `Flag` membership.
+   xor, negation).  The operands of these operations may be either enumerators
+   When the annotation :cpp:class:`nb::is_flag_enum() <is_flag_enum>` is passed to
+   to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type will be an
+   `enum.Flag`, meaning its enumerators can be combined using bitwise operators
+   in a type-safe way: the result will have the same enumeration type as the operands,
+   and only enumerators of the same type can be combined.
 
    .. code-block:: cpp
 
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -295,8 +295,11 @@ C++11-style strongly typed enumerations.
 
    When the annotation :cpp:class:`nb::is_arithmetic() <is_arithmetic>` is
    passed to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type
-   will support arithmetic and bit-level operations like comparisons, and, or,
-   xor, negation, etc.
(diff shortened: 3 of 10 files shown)


chrreisinger:
Adding `Flag` support would be highly appreciated, since we use flags in our application and the new enum behaviour is blocking us from upgrading to 2.0.

docs/api_core.rst
```diff
@@ -1926,6 +1926,14 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
+   .. cpp:struct:: is_enum_flag
```
wjakob, line 1929: flag_enum

docs/api_core.rst
```diff
@@ -1926,6 +1926,14 @@ The following annotations can be specified using the variable-length
...
+
+   Indicate that the enumeration may be used with bitwise
+   operations.  This enables the bitwise operators ``| & ^ ~``
+   with two enumeration as operands.
+   The result will an enumeration of the same type.
+   So ``Shape(2) | Shape(1) == Shepe(3)`.
```
wjakob, line 1935: Shepe -> Shape

docs/changelog.rst
```diff
@@ -129,9 +129,11 @@ according to `SemVer <http://semver.org>`__. The following changes are
 noteworthy:
 
 * The :cpp:class:`nb::enum_\<T\>() <enum_>` binding declaration is now a
-  wrapper that creates either a ``enum.Enum`` or ``enum.IntEnum``-derived type.
+  wrapper that creates either a ``enum.Enum``, ``enum.IntEnum`` or ``enum.Flag``-derived type.
   Previously, nanobind relied on a custom enumeration base class that was a
   frequent source of friction for users.
+  A new flag :cpp:class:`nb::is_flag_enum() <is_flag_enum>`
```
wjakob, line 135: This will need to go to a separate changelog entry for 2.1.0

docs/classes.rst
```diff
@@ -295,8 +295,12 @@ C++11-style strongly typed enumerations.
 
    When the annotation :cpp:class:`nb::is_arithmetic() <is_arithmetic>` is
    passed to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type
-   will support arithmetic and bit-level operations like comparisons, and, or,
-   xor, negation, etc.
+   will support arithmetic and bit-level operations and, or,
```
wjakob, line 298: (and, or ..) (in parentheses)

include/nanobind/nb_class.h
```diff
@@ -62,9 +62,12 @@ enum class type_flags : uint32_t {
...
 
     /// Is the number type underlying the enumeration signed?
-    is_signed                = (1 << 17)
+    is_signed                = (1 << 17),
 
-    // One more flag bits available (18) without needing
+    /// Is the underlying enumeration type Flag?
+    is_flag_enum                  = (1 << 18)
+
+    // No more flag bits available (18). Needs
```
wjakob, line 70: 🥲

wjakob (commented):
A small first round of comments

chrreisinger:
> @chrreisinger
> 
> > Adding Flag support would be highly appreciated, since we use flags in our application and the new enum behaviour is blocking us from upgrading to 2.0.
> 
> Could you tell me which behavior changed? The new enums should mostly behave like the old ones.

Sure.
  
A simplified version of the code is:
```
enum class Foo {
    A = 1,
    B = 2,
    C = 4,
    D = 8,
    E = 16,
    F = 32,
    ...
};
```

```
nb::enum_<pynf::Foo>(m, "Foo", nb::is_arithmetic())
        .value("A", pynf::Foo::A)
        .value("B", pynf::Foo::B)
        .value("C", pynf::Foo::C)
        .value("D", pynf::Foo::D)
        .value("E", pynf::Foo::E)
        .def("__bool__", [](const pynf::Foo val) { return pynf::to_underlying(val) != 0; })
        .def("__or__",
             [](const pynf::Foo a, const pynf::Foo b) {
                 return pynf::Foo(pynf::to_underlying(a) | pynf::to_underlying(b));
             })
        .def("__and__", [](const pynf::Foo a, const pynf::Foo b) {
            return pynf::Foo(pynf::to_underlying(a) & pynf::to_underlying(b));
        });
```

An at runtime we get an error that a combination of the flags is not valid.

` | ValueError: 31095 is not a valid Foo.`

wjakob:
Aha! Yes, that makes sense. The fact that it worked previously is effectively undefined behavior. We expect that the user returns a valid/bound enum but do not check if that is actually the case.

wjakob:
What's the status of this PR? Do you still plan to make changes? Would you like me to review it?

mwittgen:
There are some more unresolved issues, for example
`Flag.A | Flag.B` can't be passed to a C++ function `void func(enum_type e)`
Looks like this needs some type casters for `enum.Flag`

wjakob:
Yeah, that seems tricky. The existing type caster does a hash table lookup that looks for enumerants. In this case, you'd need to either populate the table with all possible combinations (combinatorial blowup) or perform some kind of integer conversion. I assume that the same issue would appear on the way back. The special case to support flag enums in this way might add a computational cost to the default enums (which I would not want). I don't have any suggestions, unfortunately.

oremanj:
It seems to me that it would work fine to fall back from the map lookups to a slightly slower alternate, either `enum_instance.value` for Py->C++ conversions or `EnumClass(number)` for C++->Py. This fallback could be done only for flag enums in order to avoid slowing down failing conversions of regular enums during overload resolution. `Flag` generates the objects representing flag combinations as they're requested, but they do still get "singletonized" / cached in the member2value_map, so we could choose to update our maps as well; then the slow path only needs to be taken once per combination. There are some other possible optimizations as well, such as caching the mask of which flags are defined, and preemptively rejecting lookups outside of that mask.

mwittgen:
Fixed the 3.11+ compilation failures with stable ABI enabled.

wjakob:
It would be helpful to have feedback from people who are invested into this -- @skallweitNV, @keithlostracco.

Does it look okay to you as well @oremanj? From what I can see, the modified PR now adopts your suggestion.

skallweitNV:
Question: Does it make sense to base nanobind's enum types on `Enum` and `Flag` in the first place? Wouldn't `IntEnum` and `IntFlag` (for enums with `nb::is_arithmetic`) be the more appropriate types? After all, we are binding C++ enums which are always of some integer type.

wjakob:
For non-arithmetic enums (e.g. C++11-style enum classes), `Enum` is the closest match in Python-land. For arithmetic ones, we use `IntEnum`. That's how the implementation works now, and I think that part makes sense.

Is your point mainly about flags?

keithlostracco:
This appears to be exactly the functionality I was looking for.

src/nb_enum.cpp
```diff
@@ -43,7 +43,7 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
                 PyUnicode_FromFormat("%U.%U", scope_qualname.ptr(), name.ptr()));
     }
 
-    const char *factory_name = is_arithmetic ? "IntEnum" : "Enum";
+    const char *factory_name = (is_arithmetic || is_flag_enum)  ? (is_flag_enum ? "Flag" : "IntEnum") : "Enum";
```
oremanj, line 46: This could be more straightforwardly written as `is_flag_enum ? "Flag" : is_arithmetic ? "IntEnum" : "Enum"`. But I think you might want the combination of flag+arithmetic to give you IntFlag as well.

src/nb_enum.cpp
```diff
@@ -56,7 +56,11 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
 
     if (is_arithmetic)
         result.attr("__str__") = enum_mod.attr("Enum").attr("__str__");
-
+    if (is_flag_enum)
+        result.attr("__str__") = enum_mod.attr("Flag").attr("__str__");
+#if PY_VERSION_HEX >= 0x030B0000
+    result.attr("_boundary_") =  enum_mod.attr("FlagBoundary").attr("KEEP");
```
oremanj, line 62: This choice seems not in keeping with nanobind's general defaults of strictness around enums. I think the default STRICT policy is more appropriate. People can use is_arithmetic (producing an IntFlag) if they want more flexibility. For IntFlag, KEEP is the default.

src/nb_enum.cpp
```diff
@@ -150,6 +154,28 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
     if (!t)
         return false;
 
+    if ((t->flags & (uint32_t) type_flags::is_flag_enum) !=0) {
+        auto base = PyObject_GetAttrString((PyObject *)o->ob_type, "__base__");
```
oremanj, line 158: These Python API calls return new object references which you must drop (`Py_DECREF`) when you're done using them. Also, this is way too much effort to spend on every individual enum-value conversion from Python to C++. You should do this only if the map lookup in `enum_tbl.rev` fails. If this slower-path conversion succeeds, then you can add to the `enum_tbl.rev` map to save time when converting the same flags value in the future.

src/nb_enum.cpp
```diff
@@ -150,6 +154,28 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
...
+        auto basename = PyObject_GetAttrString(base, "__name__");
+        Py_ssize_t size;
+        const char* data = PyUnicode_AsUTF8AndSize(basename, &size);
+        std::string s(data, size);
```
oremanj, line 162: This is not a sufficient reason to use `std::string`; nanobind generally tries to keep a low footprint of standard-library includes, to avoid code bloat. Check the size, then use `memcmp()`.

Also, I don't think this logic is correct. You should instead check whether `Py_TYPE(o)` -- the Python type of the incoming might-be-an-enumeration-value -- matches `t->type_py` -- the Python type of the nanobind enumeration you're trying to convert to. What you have here is both much slower than that, and is foolable:
```
class Flag:
    pass

class Dummy(Flag):
    value = 42

some_nanobind_function(Dummy())  # result is treated as flags of 42!

File: src/nb_enum.cpp [hunk 10]
```diff
@@ -150,6 +154,28 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
     if (!t)
         return false;
 
+    if ((t->flags & (uint32_t) type_flags::is_flag_enum) !=0) {
```
oremanj, line 157: Please follow the code style of the rest of this file. Spaces on both sides of binary operators, space after `if` (several instances of that further below), etc.

src/nb_enum.cpp
```diff
@@ -186,17 +211,19 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
...
 
 PyObject *enum_from_cpp(const std::type_info *tp, int64_t key) noexcept {
     type_data *t = nb_type_c2p(internals, tp);
     if (!t)
         return nullptr;
-
     enum_map *fwd = (enum_map *) t->enum_tbl.fwd;
-
+    if(t->flags & (uint32_t) type_flags::is_flag_enum) {
+        PyObject *value = PyLong_FromLongLong(key);
```
oremanj, line 223: This is definitely wrong and the fact that it wasn't caught by a unit test means you need better test coverage. This will turn any flag-enum value into a plain Python int, i.e., it loses its flag-enum membership. That kind of defeats the point of a flag enum, and is extra confusing because you're not allowed to pass that thing back into a C++ function that expects an instance of the flag-enum type.

oremanj (commented):
Approach in general seems fine. I would like to see some more thought put into the performance of Python/C++ interconversions, and I have some documentation and style suggestions as well. (Ultimately it's Wenzel's opinion that matters here, not mine, so if he disagrees with something I suggested, his opinion wins.)

docs/api_core.rst
```diff
@@ -1926,6 +1926,14 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
+   .. cpp:struct:: is_flag_enum
```
wjakob, line 1929: Shall we give it a shorter name? `is_flag`? (for symmetry with the arithmetic flag, which is called `is_arithmetic`)

docs/api_core.rst
```diff
@@ -1926,6 +1926,14 @@ The following annotations can be specified using the variable-length
...
+
+   Indicate that the enumeration may be used with bitwise
+   operations.  This enables the bitwise operators ``| & ^ ~``
+   with two enumeration as operands.
+   The result will an enumeration of the same type.
+   So ``Shape(2) | Shape(1) -> Shepe(3)`.
```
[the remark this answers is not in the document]
wjakob: Typo is still there? ("Shepe")

docs/changelog.rst
```diff
@@ -138,9 +143,11 @@ according to `SemVer <http://semver.org>`__. The following changes are
 noteworthy:
 
 * The :cpp:class:`nb::enum_\<T\>() <enum_>` binding declaration is now a
-  wrapper that creates either a ``enum.Enum`` or ``enum.IntEnum``-derived type.
+  wrapper that creates either a ``enum.Enum``, ``enum.IntEnum`` or ``enum.Flag``-derived type.
```
wjakob, line 146: I think it would make sense to support a matrix of 4 base classes: `nb::is_arithmetic` switches between `enum.*` and `enum.Int*` while `nb::is_flag` switches between `Enum` and `Flag`.

src/nb_enum.cpp
```diff
@@ -1,5 +1,5 @@
 #include "nb_internals.h"
-
+#include <string>
```
wjakob, line 2: Can we avoid depending on `std::string`? I've tried hard not to have anything depend on it in the core library.

src/nb_enum.cpp
wjakob: +1

src/nb_enum.cpp
wjakob: +1

src/nb_enum.cpp
wjakob: +1

wjakob: See also my point above. We likely want the full matrix of combinations.

src/nb_enum.cpp
wjakob: I'm actually wondering if it makes sense to cache these. With flags, it would seem one can generate 2^n cache entries for n bits, which has the potential of being undesirably large. Maybe flags enums are OK to go on a slow path? Or we somehow restrict the size of the cache?

wjakob (commented):
A few comments

src/nb_enum.cpp
oremanj: AFAICT, they are already being cached in members of the enum type. This "singleton-izes" so that one can test for flag combinations with `is`.
```
>>> import enum
>>> class Test(enum.Flag):
...   A, B, C, D, E, F, G, H = 1, 2, 4, 8, 16, 32, 64, 128
...
>>> for i in range(256): Test(i)
...
<Test: 0>
<Test.A: 1>
<Test.B: 2>
[etc ...]
>>> len(Test._value2member_map_)
256
```
The existing cache takes, for each referenced combination, one dict entry (like 32 bytes I think?) plus the footprint of the enum value object itself (sys.getsizeof says 48 bytes, but that doesn't include the 4-entry instance dictionary, name string, etc -- it's probably at least 128 bytes total). We're adding one robin_map entry (24 bytes?) in each direction (C++->Py and Py->C++) for each combination that is passed across the language boundary. So the additional cost of caching is something like 30% above what we're already paying just for using `enum.Flag`. That seems reasonable to me.

keithlostracco:
I test the version this morning before the last 4 commits and it worked with an or in the default arguments but with the new commits (June 18th) I get bad cast on import the extension built with nanobind in Python. Pretty sure this code is causing the bad cast:
```cpp
        .def(nb::init<CompFlagBits, uint8_t, const std::string&>(),
	        "flags"_a = CompFlagBits::InternalTimeAuto | CompFlagBits::CudaStreamDefault,
	        "device"_a = 0u, 
	        "td_path"_a = "",
	        nb::rv_policy::take_ownership)
```

skallweitNV:
> For non-arithmetic enums (e.g. C++11-style enum classes), `Enum` is the closest match in Python-land. For arithmetic ones, we use `IntEnum`. That's how the implementation works now, and I think that part makes sense.

True, for class enums the Enum is the closest match. 

> Is your point mainly about flags?

Yes, I need support for flags. Supporting the full matrix for is_arithmetic and is_flag seems like the perfect solution.

docs/api_core.rst
```diff
@@ -1926,6 +1926,14 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
+   .. cpp:struct:: flag_enum
+
+   Indicate that the enumeration may be used with bitwise
+   operations.  This enables the bitwise operators ``| & ^ ~``
+   with two with two enumerators as operands.
+   The result will have the same enumeration type as the operands.
+   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
```
nicholasjng, line 1935:
```suggestion
   with two enumerators as operands.
   The result will have the same enumeration type as the operands.
   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`.
```

docs/classes.rst
```diff
@@ -295,8 +295,13 @@ C++11-style strongly typed enumerations.
 
    When the annotation :cpp:class:`nb::is_arithmetic() <is_arithmetic>` is
    passed to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type
-   will support arithmetic and bit-level operations like comparisons, and, or,
-   xor, negation, etc.
+   will support arithmetic and bit-level operations (and, or,
+   xor, negation).  The operands of these operations may be either enumerators
+   When the annotation :cpp:class:`nb::flag_enum() <flag_enum>` is passed to
+   to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type will be an
```
nicholasjng, line 301:
```suggestion
   xor, negation).  The operands of these operations may be either enumerators
   When the annotation :cpp:class:`nb::flag_enum() <flag_enum>` is passed to
   :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type will be an
```

Also, judging from the tests, the operands must be both enumerators?

src/nb_enum.cpp
```diff
@@ -56,7 +56,8 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
 
     if (is_arithmetic)
         result.attr("__str__") = enum_mod.attr("Enum").attr("__str__");
-
+    if (is_flag)
+        result.attr("__str__") = enum_mod.attr("Flag").attr("__str__");
```
nicholasjng, line 60: Should this be `IntFlag.__str__`, since we use that factory above (L46)?

wjakob:
What's the status of this PR? Is it ready to be merged? I see that there are still some failing tests related to stubs.

src/nb_enum.cpp
```diff
@@ -43,7 +43,7 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
                 PyUnicode_FromFormat("%U.%U", scope_qualname.ptr(), name.ptr()));
     }
 
-    const char *factory_name = is_arithmetic ? "IntEnum" : "Enum";
+    const char *factory_name = (is_arithmetic || is_flag) ? (is_flag ? "IntFlag" : "IntEnum") : "Enum";
```
oremanj, line 46: This is not responsive to the previous discussion of which enum metaclass to use. `is_flag` without `is_arithmetic` should be `Flag`.

oremanj:
@wjakob I had not yet done another round of review because it seemed like there were review comments from the previous round that were unresolved - so I figured the author was still working on it. I flagged one thing I noticed.

mwittgen:
I was stuck on the python 3.10 failure. Will get back to this ticket.

wjakob:
It seems to me that this is a feature that a number of people are waiting for. @mwittgen please let us know if you'd like someone else to continue developing the PR so that it can be merged.

mwittgen:
Yes, please.

nicholasjng:
I have some time in the upcoming days, and would like to advance this further.

wjakob:
That sounds good. Could you create a new PR in this case @nicholasjng?

wjakob:
Closing this PR, development continues in PR #688.

PR #608, opened 2024-06-05
Add exception to assert introduced in #549 when the type caster of `U*` can produce the actual underlying object `U`

francesco-ballarin:
Follow up to discussion #605.

I confirmed with my downstream project that the patch does indeed fix the regression introduced in nanobind 2.0.
Please note that, even though the commit has my name and email address, the actual patch was provided by @oremanj.

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -108,8 +108,9 @@ using precise_cast_t =
 /// where the template argument T is the type you plan to extract.
 template <typename T>
 NB_INLINE uint8_t flags_for_local_caster(uint8_t flags) noexcept {
+    using Caster = make_caster<T>;
     constexpr bool is_ref = std::is_pointer_v<T> || std::is_reference_v<T>;
-    if constexpr (is_base_caster_v<make_caster<T>>) {
+    if constexpr (is_base_caster_v<Caster>) {
         if constexpr (is_ref) {
             /* References/pointers to a type produced by implicit conversions
                refer to storage owned by the cleanup_list. In a nb::cast() call,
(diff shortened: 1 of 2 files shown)


include/nanobind/nb_cast.h
```diff
@@ -123,7 +124,9 @@ NB_INLINE uint8_t flags_for_local_caster(uint8_t flags) noexcept {
            into storage owned by the caster, which won't live long enough.
            Exception: the 'char' caster produces a result that points to
            storage owned by the incoming Python 'str' object, so it's OK. */
-        static_assert(!is_ref || std::is_same_v<T, const char*>,
+        static_assert(!is_ref || std::is_same_v<T, const char*> ||
+                      std::conjunction_v<std::is_pointer<T>,
```
wjakob, line 128: Can you replace this with ordinary constexpr boolean expressions? (&&). Conjunction_v is unnecessarily heavy.

wjakob:
You need is_pointer_v, etc.

wjakob:
What would be the best way to explain this change in the changelog @oremanj ?

francesco-ballarin:
> You need is_pointer_v, etc.

Hopefully after four attempts I've got it right :(

oremanj:
I originally used conjunction because I was worried about potentially forming an invalid expression if T was not a pointer type, but on second thought I don’t think it helps or matters.

Possible changelog entry: “nanobind no longer prevents casting to a C++ container of pointers `T*` where `T` is a type with a user-defined type caster if the caster seems to operate by extracting a `T*` from the Python object rather than a `T`.” with a link to the issue that prompted this PR for more context.

I’m mostly offline until Monday and will respond to this in more depth at that time if needed.

francesco-ballarin:
Thanks @oremanj , I've added what you suggested to the changelog.

PR #618, opened 2024-06-13
Cmake qol

joaander:


joaander:
Checking with a diff, the CMake solution installs a few extra files not present in the `setup.py` solution. This can be fixed with a few small changes to the CMake code:
```
diff -r .venv2/lib/python3.12/site-packages/nanobind .venv/lib/python3.12/site-packages/nanobind
Only in .venv/lib/python3.12/site-packages/nanobind/cmake: collect-symbols-pypy.py
Only in .venv/lib/python3.12/site-packages/nanobind/cmake: collect-symbols.py
Only in .venv/lib/python3.12/site-packages/nanobind/cmake: nanobindConfigVersion.cmake
Only in .venv/lib/python3.12/site-packages/nanobind/ext/robin_map: CMakeLists.txt
Only in .venv/lib/python3.12/site-packages/nanobind/ext/robin_map/include/tsl: robin_set.h
```

pyproject.toml
```diff
@@ -1,7 +1,13 @@
+[project]
+name = "nanobind"
+version = "2.0.1"
+
 [build-system]
-requires = ["setuptools>=42", "wheel", "scikit-build", "cmake>=3.17", "ninja", "typing_extensions"]
-build-backend = "setuptools.build_meta"
+requires = ["scikit-build-core", "cmake>=3.17", "ninja"]
```
henryiii, line 6:
```suggestion
requires = ["scikit-build-core"]
```

pyproject.toml
```diff
@@ -1,7 +1,13 @@
+[project]
+name = "nanobind"
+version = "2.0.1"
```
henryiii, line 3: The other info needs to be here. "hatch new --init" cam=n do the conversion for you.

pyproject.toml
```diff
@@ -1,7 +1,13 @@
...
+
 [build-system]
-requires = ["setuptools>=42", "wheel", "scikit-build", "cmake>=3.17", "ninja", "typing_extensions"]
-build-backend = "setuptools.build_meta"
+requires = ["scikit-build-core", "cmake>=3.17", "ninja"]
+build-backend = "scikit_build_core.build"
 
+[tool.scikit-build]
+cmake.define.NB_TEST = "OFF"
```
henryiii, line 10: You need to tell scikit-build-core to make a pure wheel.

henryiii:
> sdist solution

What's wrong with SDists with this? It should Just Work™.

henryiii:
If you'd like to add me to your fork I can of the metadata conversion using hatch.

joaander:
> > sdist solution
> 
> What's wrong with SDists with this? It should Just Work™.

That's what I like to hear.

> If you'd like to add me to your fork I can of the metadata conversion using hatch.

I invited you, go ahead and help out where you can. I appreciate it.

henryiii:
Capturing the version isn't easy with the current regex plugin, since it's spread out. I can make it easier, though, in a future release, by adding a recombine option. Or a "combined" version could be placed in the header, with a compile time assert to verify it matches.

henryiii:
How is the `PROJECT_VERSION` being set? Not seeing it in the CMakeLists.txt.

joaander:
> How is the `PROJECT_VERSION` being set? Not seeing it in the CMakeLists.txt.

I added it to the `PROJECT` command. I assumed that the project used `bumpversion` to update for release, but this is not the case: https://github.com/wjakob/nanobind/blob/master/docs/release.rst

https://github.com/joaander/nanobind/blob/91785cf770d5f6acfc6f7b5c9a097ed34ea1bdb1/CMakeLists.txt#L2

We could rewrite this to match the original `setup.py` behavior and read the version from `nanobind.h`.

CMakeLists.txt
```diff
@@ -81,17 +80,27 @@ if(NB_CREATE_INSTALL_RULES AND NOT CMAKE_SKIP_INSTALL_RULES)
...
-      DIRECTORY ext/robin_map/include/
-      DESTINATION "${NB_INSTALL_EXT_DIR}/robin_map/include"
-    )
-    install(
-      FILES ext/robin_map/CMakeLists.txt
-      DESTINATION "${NB_INSTALL_EXT_DIR}/robin_map"
+      FILES ext/robin_map/include/tsl/robin_map.h
+            ext/robin_map/include/tsl/robin_hash.h
+            ext/robin_map/include/tsl/robin_growth_policy.h
+      DESTINATION "${NB_INSTALL_EXT_DIR}/robin_map/include/tsl"
```
joaander, line 86: This matches the behavior of the previous `setup.py`. It excludes `robin_set.h`. Is this intentional? Or should it install `*.h` from `robin_map`?

joaander:
I implemented the regex in cmake to pull the version from `nanobind.h`.

Aside from the duplicate `version` in `pyproject.toml` (which @henryiii) discusses in https://github.com/wjakob/nanobind/pull/618#issuecomment-2164351074, this branch now completely re-implements `setup.py` with `CMakeLists.txt` and `scikit-build-core` and adds the `nanobind-config-version.cmake`.

I tested
```
find_package(nanobind 2.0 CONFIG REQUIRED)
```
It enforces the version requirement and sets `nanobind_VERSION` as expected.

wojdyr:
my two cents: if the move to scikit-build-core is not decided, it'd still be useful to apply the part of this PR  that installs `nanobindVersionConfig.cmake`, so that users can call `find_package(nanobind …)` with version number.

joaander:
Thanks! Glad I could help.

PR #620, opened 2024-06-13
Use cmake-style path in nanobind_add_stub

metab0t:
The path of Python executable should be converted to cmake-style path to avoid invalid escape sequence.

Otherwise, `NB_STUBGEN_CMD` may contain string like `D:\mambaforge\python.exe` on Windows and will make this line writes incorrect string:
https://github.com/wjakob/nanobind/blob/c5ae2a36a704c1c62da99a81fad919261a3e9848/cmake/nanobind-config.cmake#L446

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -421,9 +421,9 @@ function (nanobind_add_stub name)
     list(APPEND NB_STUBGEN_OUTPUTS "${ARG_OUTPUT}")
   endif()
 
-  file(TO_CMAKE_PATH ${Python_EXECUTABLE} Python_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


cmake/nanobind-config.cmake
```diff
@@ -421,6 +421,8 @@ function (nanobind_add_stub name)
     list(APPEND NB_STUBGEN_OUTPUTS "${ARG_OUTPUT}")
   endif()
 
+  file(TO_CMAKE_PATH ${Python_EXECUTABLE} Python_EXECUTABLE)
```
wjakob, line 424: Would it not be better to give this an internal name, e.g. `NB_PYTHON_EXECUTABLE`?

metab0t: Yeah, it is better. Will change it.

wjakob:
Thanks!

PR #627, opened 2024-07-01
[cmake] include nanobind-config.cmake instead of calling find_package()

KerstinKeller:
find_package() is an abstraction to be used to localize a package. When building this project, the same can be achieved by just including the  `cmake/nanobind-config.cmake` file directly.

Especially the current implementation breaks in cases of using CMake dependency providers, where the dependency provider executes an `add_subdirectory` call. In this specific case the find_package leads to an infinite recursion of `find_package(nanobind)` calls.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,10 +139,7 @@ endif()
 # ---------------------------------------------------------------------------
 # Include nanobind cmake functionality
 # ---------------------------------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


wjakob:
This seems fine to me. I don't consider myself an expert on CMake, especially packing of CMake tooling. If this breaks someone's workflow, I may have to revert the change.

PR #631, opened 2024-07-03
Fix doc typo in `nanobind_add_stubgen`

breathe:
I was trying to use the PATTERN_FILE mechanism to rewrite the type's inferred for c++ `Eigen::` from `numpy.typing.ArrayLike` (which is overly broad) to `numpy.typing.NDArray`.  

While [testing to see if that mechanism](https://github.com/wjakob/nanobind/discussions/630) could work for what I need -- I noticed this documentation bug.

It seems the PATTERN_FILE mechanism won't work for what I need -- but thought I'd file this doc fix anyway.


wjakob:
Thanks!

PR #632, opened 2024-07-05
Handle nested class names in stubgen

### LINKED ISSUE #633: [BUG]: Stubgen does not produce correct import statements for nested class names
### Problem description

When a type signature in a binding references a nested type (e.g. a class defined inside another class), `stubgen` generates an `import` statement in the .pyi file which is not correct. The generated `import` statement includes the outer type.

### Reproducible example code

```text
See https://github.com/wjakob/nanobind/pull/632 for a failing unit test and a proposed fix.
```

ryancahoon-zoox:
Correct `stubgen`'s handling of nested types (e.g. a class defined inside another class).

Previously, `stubgen` generated `import` statements in the .pyi file which includes the outer type.

I fix this by testing that the module name in the generated `import` is a module by using `importlib.util.find_spec`. This is the approach [recommended by the python docs](https://docs.python.org/3/library/importlib.html#checking-if-a-module-can-be-imported): "If you need to find out if a module can be imported without actually doing the import, then you should use `importlib.util.find_spec()`."

If `stubgen` is not able to access the referenced module, the new functionality does have a graceful fall-back behavior: if none of the possible partitions of module_name and class_name result in a successful import, it will default to the previous behavior of assuming that just the last segment of the type name is the class name (and the rest is the module name).

This may add some performance overhead from needing to call `importlib.util.find_spec` on each referenced type, but I don't know of another way to handle this nested-type case, unless we rely only on some sort of naming convention to distinguish type names from module names.

A unit test for the nested-type case is included.

Closes #633

### DIFF
--- a/tests/test_typing.cpp
+++ b/tests/test_typing.cpp
@@ -4,6 +4,29 @@
 namespace nb = nanobind;
 using namespace nb::literals;
 
+class NestedClass {};
+
+namespace nanobind {
+namespace detail {
+template <>
+struct type_caster<NestedClass> {
+    NB_TYPE_CASTER(NestedClass, const_name("py_stub_test.AClass.NestedClass"));
+
+    bool from_python(handle /*src*/, uint8_t /*flags*/, cleanup_list*) noexcept {
+        return true;
+    }
+
+    static handle from_cpp(const NestedClass&, rv_policy, cleanup_list*) noexcept {
+        nanobind::object py_class =
+            nanobind::module_::import_("py_stub_test").attr("AClass").attr("NestedClass");
+        nanobind::object py_proto = py_class();
+        py_proto.inc_ref();
+        return py_proto;
+    }
+};
+}
+}
+
 // Declarations of various advanced constructions to test the stub generator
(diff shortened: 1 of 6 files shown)


tests/test_typing.cpp
```diff
@@ -4,6 +4,29 @@
...
+
+    bool from_python(handle /*src*/, uint8_t /*flags*/, cleanup_list*) noexcept {
+        return true;
+    }
+
+    static handle from_cpp(const NestedClass&, rv_policy, cleanup_list*) noexcept {
+        nanobind::object py_class =
+            nanobind::module_::import_("py_stub_test").attr("AClass").attr("NestedClass");
+        nanobind::object py_proto = py_class();
+        py_proto.inc_ref();
```
wjakob, line 23: These two lines can be shortened to: `return py_class().release()`

wjakob:
Looks good to me, just a minor comment.

tests/test_typing.cpp
ryancahoon-zoox: Done

PR #634, opened 2024-07-06
remove two semicolons (-Wextra-semi ritual purification)

wojdyr:
I read the comment about "ritual purification" in nb_attr.h, but couldn't resist.
(I don't plan more such PRs)


wjakob:
It's ok ;)

PR #635, opened 2024-07-06
Fix deprecated enum-enum conversion warnings in C++20

calcmogul:
Merge requests to Eigen have just been casting to int to fix the warning.


wjakob:
Thanks!

PR #637, opened 2024-07-11
Handle print leaks when `inst_c2p` has collisions

rtabbara:
Populating the internals `inst_c2p` map handles the case where there are C++ instance pointer key collisions. This can occur for example when there are several distinct Python objects which reference the same memory address, such as a struct and its first member. In this instance the value within the map becomes a linked-list of type `nb_inst_seq*`.

In the case where there are reported nanobind leaks, there was missing logic to handle the possibility of the values in our `inst_c2p` map being a linked-list and hence this PR adds this additional check.


wjakob:
Thank you Rami!

PR #651, opened 2024-07-24
Fix stubgen crash when module doesn't set __spec__

calcmogul:
Per https://docs.python.org/3/library/importlib.html#importlib.util.find_spec, find_spec() can throw ValueError if `__spec__` isn't set.


wjakob:
LGTM, Thanks!

Issue #653, opened 2024-07-26
test37_noninteger_stride fails with numpy<1.23


wojdyr:
### Problem description

This test could be skipped for numpy<1.23. It fails with:
```
    @needs_numpy
    def test37_noninteger_stride():
        a = np.array([[1, 2, 3, 4, 0, 0], [5, 6, 7, 8, 0, 0]], dtype=np.float32)
        s = a[:, 0:4]  # slice
        t.pass_float32(s)
        assert t.get_stride(s, 0) == 6;
        assert t.get_stride(s, 1) == 1;
>       v = s.view(np.complex64)
E       ValueError: To change to a dtype of a different size, the array must be C-contiguous

test_ndarray.py:689: ValueError
```
It was changed in https://github.com/numpy/numpy/commit/1a009e0d061df490fa91df8684768ed3217704db

hpkfft:
Hi @wojdyr,

Thank you for filing this issue and clearly documenting when numpy added the functionality on which this test relies.
If you have time, please check whether PR #659 fixes this issue for you.   (The fix is simple--it may be easiest to just cut-and-paste the change into your local copy of `test_ndarray.py`.)

Regards,
Paul

wojdyr:
Hi Paul, thanks, your fix works.

When I was opening this issue I was actually wondering if I should open a PR instead. But since this test can be skipped in more than one way (one could, alternatively, check `np.__version__`) and I couldn't decide which way is better, I ended up just opening the issue. Thanks for picking it up.

wjakob:
Closed via #659

PR #659, opened 2024-07-30
Skip ndarray noninteger stride test for numpy<1.23.

hpkfft:
The ndarray test `test37_noninteger_stride` uses numpy to create an array with padding between rows and then uses numpy `view()`  to convert to a complex dtype.   This capability was added in numpy 1.23.  Previously, for this conversion to work, the entire array had to be C-contiguous.

This commit fixes issue #653.


wjakob:
Thanks!

PR #661, opened 2024-08-03
type_caster<char>: allow conversion of "\0" to C++ char

### LINKED ISSUE #657: [BUG]: casting to char doesn't allow \0
### Problem description

A C++ function takes a char arg. Automatic type casting works except when the character is \0 (which is a valid input for this C++ function).

nanobind 2.0.1 from git. This problem is not present in pybind11.

It looks to me that to tell apart `""` from `"\0"`, the caster would need to use the second arg of PyUnicode_AsUTF8AndSize.

### Reproducible example code

```text
import test_functions_ext as t
  
t.test_cast_char("c")     # OK
t.test_cast_char("\x41")  # OK -> 'A'
t.test_cast_char("\x00")  # TypeError
```

wojdyr:
Fixes #657

Stores the size from `PyUnicode_AsUTF8AndSize()` and checks it in `can_cast<char>`.

I stored the size rather than a boolean is-length-equal-one flag, because the total struct size is the same anyway, and if we'd want to convert 2-byte UTF8 characters (U+0080 - U+00FF) to C++ chars (128-255) in the future, we'll need to check if size==2.

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -275,12 +275,13 @@ template <> struct type_caster<bool> {
 template <> struct type_caster<char> {
     using Value = const char *;
     Value value;
+    Py_ssize_t size;
     static constexpr auto Name = const_name("str");
     template <typename T_>
     using Cast = std::conditional_t<is_pointer_v<T_>, const char *, char>;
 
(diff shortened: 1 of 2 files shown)


wjakob:
LGTM, thanks!

PR #665, opened 2024-08-05
fix typo in documentation

mgeplf:


wjakob:
Thanks!

wojdyr:
Btw do you know codespell?

Doing just `pip install codespell` and then `codespell` without configuration
finds a number of typos in any sizeable project, usually with not too many false positives.

PR #667, opened 2024-08-06
nb_cast: Add support for `parent` arg to allow for `reference_internal` rv policy

rtabbara:
Also small stubgen fix. Need to use != operator for string comparison

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -606,6 +606,20 @@ object cast(T &&value, rv_policy policy = rv_policy::automatic_reference) {
     return steal(h);
 }
 
+template <typename T>
+object cast(T &&value, rv_policy policy, handle parent) {
+    detail::cleanup_list cleanup(parent.ptr());
+    handle h = detail::make_caster<T>::from_cpp((detail::forward_t<T>) value,
+                                                policy, &cleanup);
+
+    cleanup.release();
+
+    if (!h.is_valid())
+        detail::raise_cast_error();
+
+    return steal(h);
+}
+
 template <typename T> object find(const T &value) noexcept {
     return steal(detail::make_caster<T>::from_cpp(value, rv_policy::none, nullptr));
 }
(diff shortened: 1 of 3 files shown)


wjakob:
Thanks Rami! 

Two requests:
1. Could you separate these changes into two commits for the two separate components.
2. Given the fact that the change to `nb::cast` now requires a cleanup list, I would say it's better to have two separate versions of the `nb::cast` interface (one with the `parent` argument, the other without). Also, the cleanup list requires you to call a cleanup routine explicitly at the end. The destructor does not do it. (An questionable design decision related to a small impact on binary size, and this really being more of an implementation detail that users aren't ever expected to call ;))

rtabbara:
Sure, no problem at all! I've updated the PR based off of your requests.

include/nanobind/nb_cast.h
```diff
@@ -606,6 +606,19 @@ object cast(T &&value, rv_policy policy = rv_policy::automatic_reference) {
...
 
+template <typename T>
+object cast(T &&value, handle parent) {
+    detail::cleanup_list cleanup(parent.ptr());
+    handle h = detail::make_caster<T>::from_cpp((detail::forward_t<T>) value,
+                                                rv_policy::reference_internal, &cleanup);
+    if (!h.is_valid())
+        detail::raise_cast_error();
+
+    cleanup.release();
```
wjakob, line 617: I think you will need to do the `.release()` call one line above already. Otherwise we might raise an exception with leaks.

wjakob: Can you also add this additional function to the `docs/api_core.rst` file?

wjakob:
Hi Rami,

sorry, one more thing. Could you still keep the rv_policy as second argument of the added function instead of fixing it to reference_internal? This way, the API behaves more like single function with defaults (which you had before), and the usage is also more explicit on the caller’s side.

rtabbara:
Sure, I think I've fixed it up now.

wjakob:
Thanks!

PR #674, opened 2024-08-08
Update documentation with new nanobind-bazel v2.1.0 functionality

nicholasjng:
Includes stubgen with the `nanobind_stubgen()` macro, shared libraries with `nanobind_shared_library()`, and contains a small prose section on how to declare (and run) stubgen targets alongside nanobind extensions in Bazel.

Also, update all mentions of a BCR version number to v2.1.0 (current latest) to increase adoption of the latest rule version.

-------------------------------

Hi Wenzel, this is the documentation on additions made in nanobind-bazel v2.1.0, which I released to the BCR two weeks ago. Sorry for the delay, I've been in a bit of a productivity hole lately. Happy to hear your thoughts!

### DIFF
--- a/docs/api_bazel.rst
+++ b/docs/api_bazel.rst
@@ -35,14 +35,46 @@ The main tool to build nanobind extensions is the ``nanobind_extension`` rule.
 
     It corresponds directly to the builtin
     `cc_binary <https://bazel.build/reference/be/c-cpp#cc_binary>`__ rule,
-    with all keyword arguments being directly forwarded to a ``cc_binary`` 
+    with all keyword arguments being directly forwarded to a ``cc_binary``
     target.
 
     The ``domain`` argument can be used to build the target extension under a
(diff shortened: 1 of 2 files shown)


wjakob:
This looks great -- thank you for integrating stubgen support!

wjakob:
Hi @nicholasjng,

I realized that the `api_bazel.rst` file uses the CMake domain. For example, one line declares

```rst
cmake:command:: nanobind_extension
```

This is problematic because it conflicts with a CMake command of the same name declared in `api_cmake.rst`, which breaks crossreferencing. 

Could I ask you to look into this and declare the Bazel commands in a different Sphinx domain?

nicholasjng:
From my first scan of PyPI, I don't think there is a Bazel domain as readily available as CMake is. I will work on a fix.

nicholasjng:
Just thinking simply, would the `py` domain work for you? I think Starlark is a true subset of Python, which would presumably make highlighting and cross-referencing work out of the box.

PR #675, opened 2024-08-08
generalize type_caster<std::optional<T>>

wojdyr:
Add `optional_caster<>` that works like in pybind11.
It can be used for type casting optional types other than std::optional:
boost::optional, absl::optional, nonstd::optional, llvm::Optional, etc.

I guess these types are mostly obsolete in C++17, but there still might be a reason to use other such types. In one project I use a custom OptionalInt that has 4 bytes instead of 8 (at the cost of not using INT_MIN).

### DIFF
--- a/include/nanobind/nb_cast.h
+++ b/include/nanobind/nb_cast.h
@@ -238,19 +238,21 @@ template <> struct type_caster<void> {
     }
 };
 
-template <> struct type_caster<std::nullptr_t> {
+template <typename T> struct none_caster {
     bool from_python(handle src, uint8_t, cleanup_list *) noexcept {
         if (src.is_none())
             return true;
         return false;
     }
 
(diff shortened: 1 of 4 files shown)


wjakob:
This looks good, but how about completely separating from the STL specifics by putting the optional caster into the `detail` subfolder, just like is done for the list casters?

wojdyr:
Makes sense, I moved it to detail/nb_optional.h, and rebased on master.

wojdyr:
BTW, I see that type_casters for `std::nullptr_t`, `std::nullopt_t` and `std::monostate` are the same caster.
Would `none_caster` be a good name for a base struct for these three?

wjakob:
Sure, that would be a nice simplification. Do you wish to include this in the current PR?

wojdyr:
added

wjakob:
Thank you!

PR #676, opened 2024-08-09
Fix handling of keyword arguments with nb::new_()

### LINKED ISSUE #668: [BUG]: nb::new_ + nb::arg triggers assertion about the wrong argument count
### Problem description

I double checked it on [the example from docs][1]. It gives:

    nanobind/include/nanobind/nb_func.h: In instantiation of ‘PyObject* nanobind::detail::func_create(Func&&, Return (*)(Args ...), std::index_sequence<Is2 ...>, const Extra& ...) [with bool ReturnRef = false; bool CheckGuard = true; Func = nanobind::new_<std::unique_ptr<Pet> (*)(std::__cxx11::basic_string<char>, int), std::unique_ptr<Pet>(std::__cxx11::basic_string<char>, int)>::execute<nanobind::class_<Pet>, nanobind::arg, nanobind::arg>(nanobind::class_<Pet>&, const nanobind::arg&, const nanobind::arg&)::<lambda(nanobind::handle, std::__cxx11::basic_string<char>, int)>; Return = std::unique_ptr<Pet>; Args = {nanobind::handle, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int}; long unsigned int ...Is = {0, 1, 2}; Extra = {nanobind::scope, nanobind::name, nanobind::arg, nanobind::arg}; PyObject = _object; std::index_sequence<Is2 ...> = std::integer_sequence<long unsigned int, 0, 1, 2>]’:
    …
    nanobind/include/nanobind/nb_func.h:91:40: error: static assertion failed: The number of nb::arg annotations must match the argument count!
       91 |     static_assert(!has_arg_annotations || nargs_provided + is_method_det == nargs,

[1]: https://nanobind.readthedocs.io/en/latest/classes.html#customizing-python-object-creation

### Reproducible example code

```c++
class Pet {
  private:
    std::string name;
    int age;
    Pet(const std::string &name, int age) : name(name), age(age) { }
  public:
    static std::unique_ptr<Pet> make(std::string name, int age) {
        return std::unique_ptr<Pet>(new Pet(name, age));
    }
};

// ...

   nb::class_<Pet>(m, "Pet")
        .def(nb::new_(&Pet::make), "name"_a, "age"_a);
```

oremanj:
Fixes #668. Note that `__new__` is not a classmethod, because you have to pass the initial `cls` argument when you call it, so I didn't think this merited an additional function attribute. Instead I'm treating it as simply an interface transformation between `__new__` which takes the class parameter, and the argument of `new_()` which doesn't.


wjakob:
Thanks!

PR #683, opened 2024-08-13
support None for const char* arg, with minimal overhead

wojdyr:
Since pybind11 also accepts None for `const char*` as an exception among built-in types, remove a paragraph about differences in porting.rst.

The overhead is minimal. Checking the size of  `test_functions_ext.cpython-311-x86_64-linux-gnu.so` built with -O3, it increased only by 32 bytes.

I'm aware that it's possible to use `std::optional<const char*>`, but to me it looks more awkward than `std::optional<int>`, perhaps because `const char*`, unlike `std::string`, looks already optional.

---
There is a third place in docs where None arg is covered:

https://github.com/wjakob/nanobind/blob/8e5fd142d9791f3f7ff66485de479578e64d9f37/docs/functions.rst?plain=1#L178-L181

Perhaps it's outdated, because having argument such as `int*` works fine, just doesn't accept None. Or I misunderstand it. Anyway, I didn't try to edit it.


include/nanobind/nb_cast.h
```diff
@@ -286,7 +286,8 @@ template <> struct type_caster<char> {
         value = PyUnicode_AsUTF8AndSize(src.ptr(), &size);
         if (!value) {
             PyErr_Clear();
-            return false;
+            // optimize for src being a string, check for None afterwards
+            return src.is_none();
```
hpkfft, line 290: Maybe you want:
```
    return src.is_none() && flags & (uint8_t) cast_flags::accepts_none
```
where `flags` is the name given to the second argument to `from_python()` above.
But I'm not the expert on such things....

wojdyr: I think it's not needed. `accepts_none` is checked in `nb_func_vectorcall_complex()` and None can't go through `nb_func_vectorcall_simple()`. I remember testing it, but I could be missing something.

wojdyr:
I added a check for flags as @hpkfft proposed above, in case `from_python` is called differently than by the function dispatch mechanism. And rebased.

PR #687, opened 2024-08-19
Fix the signature of tp_new

hoodmane:
[According to the docs](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_new), the signature of `tp_new` should be: `PyObject* tp_new(PyTypeObject *subtype, PyObject *args, PyObject *kwds)`

We don't care about the arguments, but [according to the C standard clause 6.3.2.3.8](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf#page=60) this is undefined behavior:

> A pointer to a function of one type may be converted to a pointer to a
> function of another type and back again; the result shall compare equal to
> the original pointer. If a converted pointer is used to call a function whose
> type is not compatible with the pointed-to type, the behavior is undefined.

In normal platforms the behavior is to work as expected, but in WebAssembly, this crashes.

Fixes https://github.com/pyodide/pyodide/issues/5015.

cc @henryiii @ryanking13

### DIFF
--- a/src/nb_internals.h
+++ b/src/nb_internals.h
@@ -285,7 +285,7 @@ extern char *type_name(const std::type_info *t);
 
 // Forward declarations
 extern PyObject *inst_new_ext(PyTypeObject *tp, void *value);
-extern PyObject *inst_new_int(PyTypeObject *tp);
+extern PyObject *inst_new_int(PyTypeObject *tp, PyObject *args, PyObject *kwds);
 extern PyTypeObject *nb_static_property_tp() noexcept;
 extern type_data *nb_type_c2p(nb_internals *internals,
                               const std::type_info *type);
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -65,7 +65,7 @@ static int inst_init(PyObject *self, PyObject *, PyObject *) {
 }
 
 /// Allocate memory for a nb_type instance with internal storage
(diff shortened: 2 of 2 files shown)


hoodmane:
I wonder if we could add Pyodide tests to nanobind CI.

henryiii:
We've got them in pybind11, so probably?

hoodmane:
The CI uses cmake directly and I was unable to figure out how to get it to stop saying: 
> Could NOT find Python (missing: Python_INCLUDE_DIRS Development.Module)

So probably someone with a bit more cmake experience would have to do this.

As an aside, I see that people on the internet suggest using `python -c "import sysconfig; print(sysconfig.get_path('include'))"` to get the include path, which doesn't work correctly in our Pyodide virtual environments. Probably we should fix that...

leandro-benedet-garcia:
@hoodmane I will give it a try in another PR

henryiii:
FYI, in pybind11 I used scikit-build-core to make a `tests/pyproject.toml` that build the tests as a wheel, then made sure the tests could run against an installed version of the tests binary.

wjakob:
Oops, that's not good. Thank you for finding and fixing this issue. I am open to including PyIodide in the test matrix—however, someone else would need to create a PR, I don't have the time to pursue this myself.

PR #688, opened 2024-08-19
Add support for Flag enums, continued

nicholasjng:
Continuation of #599, as discussed.

I'm interested in how to fix the tests - I added extra logic on which metaclass and ` __str__` method to use as per https://github.com/wjakob/nanobind/pull/599#discussion_r1689110819, but this breaks more tests than before.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1971,7 +1971,7 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
-   .. cpp:struct:: is_flag
+   .. cpp:struct:: flag_enum
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1975,7 +1975,7 @@ The following annotations can be specified using the variable-length
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
-   with two with two enumerators as operands.
+   with two enumerators as operands.
    The result will have the same enumeration type as the operands.
    So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
 
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1972,7 +1972,7 @@ The following annotations can be specified using the variable-length
    permissible.
    Changes the resulting Python enumeration type to be based on `enum.IntEnum`.
 
-   .. cpp:struct:: flag_enum
+   .. cpp:struct:: is_flag
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1975,9 +1975,9 @@ The following annotations can be specified using the variable-length
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
-   with two enumeration as operands.
-   The result will an enumeration of the same type.
-   So ``Shape(2) | Shape(1) -> Shepe(3)`.
+   with two with two enumerators as operands.
+   The result will have the same enumeration type as the operands.
+   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
 
 Function binding
 ----------------
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1971,6 +1971,14 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
+   .. cpp:struct:: is_flag_enum
+
+   Indicate that the enumeration may be used with bitwise
+   operations.  This enables the bitwise operators ``| & ^ ~``
+   with two enumeration as operands.
+   The result will an enumeration of the same type.
+   So ``Shape(2) | Shape(1) -> Shepe(3)`.
+
 Function binding
 ----------------
 
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1971,7 +1971,7 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
-   .. cpp:struct:: is_flag_enum
+   .. cpp:struct:: is_flag
 
    Indicate that the enumeration may be used with bitwise
    operations.  This enables the bitwise operators ``| & ^ ~``
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1534,7 +1534,7 @@ Casting
 
    Convert the C++ object ``value`` into a Python object. The return value
    policy `policy` is used to handle ownership-related questions when a new
-   Python object must be created. A valid `parent` object is required when 
+   Python object must be created. A valid `parent` object is required when
    specifying a `reference_internal` return value policy.
 
    The function raises a :cpp:type:`cast_error` when the conversion fails.
@@ -1970,6 +1970,7 @@ The following annotations can be specified using the variable-length
    ``Shape(2) * 1.5 == 3.0``.) It is unspecified whether operations on
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
+   Changes the resulting Python enumeration type to be based on `enum.IntEnum`.
 
    .. cpp:struct:: flag_enum
 
@@ -1977,7 +1978,10 @@ The following annotations can be specified using the variable-length
    operations.  This enables the bitwise operators ``| & ^ ~``
    with two enumerators as operands.
    The result will have the same enumeration type as the operands.
-   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
+   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`.
+   Changes the resulting Python enumeration type to be based on either
+   `enum.IntFlag` or `enum.Flag`, depending on whether or not the enumeration
+   is marked to support arithmetic operations.
 
 Function binding
 ----------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,8 +19,8 @@ Version 2.2.0 (TBA)
 -------------------
 
 * nanobind now allows a :cpp:class:`nb::enum_\<T\>() <enum_>` to specify the
-  new class binding annotation :cpp:class:`nb::flag_enum() <flag_enum>`,
-  which produces an enumeration type derived from `enum.IntFlag`. Instances of such
+  new class binding annotation :cpp:class:`nb::is_flag() <is_flag>`,
+  which produces an enumeration type derived from `enum.Flag`. Instances of such
   an enumeration type represent a bitwise combination of the defined enumerators,
   and they may be combined using bitwise operators ``& | ^ ~``.
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,8 +18,11 @@ below inherit that of the preceding release.
 Version 2.2.0 (TBA)
 -------------------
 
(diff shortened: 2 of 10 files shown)


wjakob:
I would like that arithmetic continue to print the associated member. The actual numbers in an enumeration are often an irrelevant implementation detail, and it's the identifier that matters. (This is really useful when getting dumps of function signatures in error messages or generating stubs)

nicholasjng:
Fair. I added `Enum.__str__` as the canonical `__str__` method for all generated enum classes.

That leaves the "invalid value" error, which is a consequence of the `STRICT` boundary together with `Flag._flag_mask_` being set to 0 for generated nanobind enums based on `enum.Flag` -  it should likely be `2 ** num_members - 1`, but I don't know how to obtain the size of a C++ enum. I will investigate.

wjakob:
> That leaves the "invalid value" error, which is a consequence of the STRICT boundary together with Flag._flag_mask_ being set to 0 for generated nanobind enums based on enum.Flag - it should likely be 2 ** num_members - 1, but I don't know how to obtain the size of a C++ enum. I will investigate.

How about just `AND`-ing together all the fields as they are registered?

By the way, to understand how to add enum entries after the fact, I found the following project helpful: https://github.com/ethanfurman/aenum

It's written by the creator of the `enum` package and includes a function to extend an enum, in a way that works across various Python versions and `enum` implementations. I used as inspiration when writing the new C++-based wrapper of the builtin Python enums.

wjakob:
Is this ready for a review and potential merge?

nicholasjng:
yes! I could not request a review on my own, it would be great if you could take a look (and potentially others from the original PR) when you have time. Thanks!

docs/api_core.rst
```diff
@@ -1971,6 +1971,14 @@ The following annotations can be specified using the variable-length
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
 
+   .. cpp:struct:: flag_enum
+
+   Indicate that the enumeration may be used with bitwise
+   operations.  This enables the bitwise operators ``| & ^ ~``
+   with two enumerators as operands.
+   The result will have the same enumeration type as the operands.
+   So ``Shape(2) | Shape(1)`` is equivalent to ``Shape(3)`
```
wjakob, line 1980: In both `flag_enum` and `is_arithmetic`, it may be worth mentioning that these flags change the base type of the enumeration (`enum.IntEnum`, `enum.IntFlag`, etc). Many people will have worked with these already and then sort of know what to expect.

docs/classes.rst
```diff
@@ -295,8 +295,13 @@ C++11-style strongly typed enumerations.
...
    passed to :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type
-   will support arithmetic and bit-level operations like comparisons, and, or,
-   xor, negation, etc.
+   will support arithmetic and bit-level operations (and, or,
+   xor, negation).  The operands of these operations may be either enumerators
+   When the annotation :cpp:class:`nb::flag_enum() <flag_enum>` is passed to
+   :cpp:class:`nb::enum_\<T\> <enum_>`, the resulting Python type will be an
+   `enum.IntFlag`, meaning its enumerators can be combined using bitwise operators
+   in a type-safe way: the result will have the same enumeration type as the operands,
+   and only enumerators of the same type can be combined.
```
wjakob, line 304: I don't think this text is up-to-date wrt. the code. We can have combinations of `flag_enum` and `is_arithmetic`, which influences the base type and valid set of arithmetic operators for a total of 4 possibilities.

src/nb_enum.cpp
```diff
@@ -113,6 +119,23 @@ void enum_append(PyObject *tp_, const char *name_, int64_t value_,
         fail("refusing to add duplicate key \"%s\" to enumeration \"%s\"!",
              name_, type_name(tp).c_str());
 
+    # if PY_VERSION_HEX >= 0x030B0000
+    // In Python 3.11+, update the flag and bit masks by hand,
+    // since enum._proto_member.__set_name__ is not called in this code path.
+    if (t->flags & (uint32_t) enum_flags::flag_enum) {
+        int64_t flag_mask = cast<int64_t>(tp.attr("_flag_mask_"));
```
wjakob, line 126: This cast will fail when `_flag_mask_` does not fit into a signed 64 bit integer (e.g. when all bits are set). That seems to me to be a fairly common situation and should probably be covered in the tests. It might be that we need special cases for signed and unsigned integers.

Alternatively, you might be able to write this code in terms of Python `int` objects without ever casting to C++.

src/nb_enum.cpp
```diff
@@ -150,6 +173,21 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
     if (!t)
         return false;
 
+    if ((t->flags & (uint32_t) enum_flags::flag_enum) != 0 && Py_TYPE(o) == t->type_py) {
```
wjakob, line 176: What situation would cause the first part of the condition to be `true`, while the second part is `false`?

src/nb_enum.cpp
```diff
@@ -150,6 +173,21 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
...
+        auto pValue = PyObject_GetAttrString(o, "value");
```
wjakob, line 177: `auto` is great for complicated C++ types like (`std::map<.., ..>::const_iterator`, etc.). Here, I greatly prefer to be explicit and write `PyObject *`.

src/nb_enum.cpp
```diff
@@ -150,6 +173,21 @@ bool enum_from_python(const std::type_info *tp, PyObject *o, int64_t *out, uint8
...
+        if (pValue == nullptr) {
+            PyErr_Clear();
+            return false;
+        }
+        long long value = PyLong_AsLongLong(pValue);
```
wjakob, line 182: See above for the issue with signed 64 bit integers and all bits being set.

src/nb_enum.cpp
wjakob: More thoughts on this: Can flag enums actually be signed? If not, perhaps `uint64_t` is the right type to use.

wjakob (commented):
Thank you -- here is a first round of feedback.

src/nb_enum.cpp
nicholasjng: Can I actually cast this completely safely from overflows? I checked the Python docs, and each Py->C++ conversion (like PyLong_AsLongLong) seems to carry some overflow management.

Can I get away here by doing inplace computations like `aenum`, maybe like calling `tp.attr("_flag_mask_).attr("__iadd__")()`? I guess you meant that by avoiding the casts?

wjakob: You can cast to 64 bit signed and unsigned integers based on the signedness of the type, and that will have you covered for integers up to 64 bit (which is in any case, the most that `enum` portably supports in C++).

You will see that the existing enum code follows this pattern in some places. An alternative to this is that you can also do arithmetic directly using the wrapper types. For example, `nb::handle` et al. support `operator|`, etc.

docs/api_core.rst
```diff
@@ -1970,6 +1970,18 @@ The following annotations can be specified using the variable-length
    ``Shape(2) * 1.5 == 3.0``.) It is unspecified whether operations on
    mixed enum types (such as ``Shape.Circle + Color.Red``) are
    permissible.
+   Changes the resulting Python enumeration type to be based on `enum.IntEnum`.
+
+   .. cpp:struct:: flag_enum
```
nicholasjng, line 1975: Should this be changed back to `is_flag` to be more in line with the other flags? That came up already in the previous PR.

src/nb_enum.cpp
wjakob: Currently, the in-place operators only work when the left hand side is a `nb::object ` subclass.

`tp.attr("_flag_mask_") |= val` will return a Python object with the result but cannot modify the left hand side.
This is a stupid limitation and something that could likely be fixed by adding the overloads to `nb_accessor.h`.

docs/api_core.rst
wjakob: Yeah, I prefer that name.

docs/changelog.rst
```diff
@@ -18,6 +18,12 @@ below inherit that of the preceding release.
 Version 2.2.0 (TBA)
 -------------------
 
+* nanobind now allows a :cpp:class:`nb::enum_\<T\>() <enum_>` to specify the
```
wjakob, line 21: The changelog entry is out of date (Enum vs IntEnum, etc.)

src/nb_enum.cpp
```diff
@@ -113,6 +119,21 @@ void enum_append(PyObject *tp_, const char *name_, int64_t value_,
...
              name_, type_name(tp).c_str());
 
+    # if PY_VERSION_HEX >= 0x030B0000
+    // In Python 3.11+, update the flag and bit masks by hand,
+    // since enum._proto_member.__set_name__ is not called in this code path.
+    if (t->flags & (uint32_t) enum_flags::flag_enum) {
+        setattr(tp, "_flag_mask_", tp.attr("_flag_mask_") | val);
+
+        bool is_single_bit = (value_ != 0) && (value_ & (value_ - 1)) == 0;
+        if (is_single_bit) {
```
wjakob, line 129: the nanobind codebase skips the braces for a simple 1-liner `if` condition body.

tests/test_enum.py
```diff
@@ -136,6 +136,19 @@ def test08_enum_comparisons():
     assert not (t.Enum.B == t.SEnum.B) and not (t.SEnum.B == t.Enum.B)
     assert t.Enum.B != t.SEnum.C and t.SEnum.C != t.Enum.B
 
+def test06_enum_flag():
+    assert (t.Flag(1) | t.Flag(2)).value == 3
```
wjakob, line 140: Tests for string conversion are missing.

src/nb_enum.cpp
wjakob: It's not uncommon to have an enumeration like

```cpp
enum class Foo : uint64_t {
     A = 1 << 0,
     B = 1 << 1,
     // ...
     All = (uint64_t) -1
};
```
I'm worried that this won't pass the tests because it cannot be converted to a `long long`. In fact, I think it would be good to add an explicit testcase for a flag enum since an all-bits-1 enumerant is such a standard pattern there.

wjakob (commented):
A few more comments.

wjakob:
> Just as a question of understanding before I start addressing the newest comments: Will the proposed enum actually require 64 entries (for uint64_t), or is it more about the All value being at the end of uint64_t range?

I'm thinking of enumerations that include one member representing a mask that has all bits set, not necessarily an enum with 64 separate flag entries.

nicholasjng:
I added tests for `__str__/__repr__` as well as an unsigned uint64_t flag test case. The only failing test on my local machine is `t.from_enum(t.UnsignedFlag.C) == 0xffffffffffffffff`, presumably because of the overflow UB.

Is the coverage okay now? I went off the existing enum test cases (test 01), let me know if you would like to see more.

tests/test_enum.py
nicholasjng: Added above in [599b191](https://github.com/wjakob/nanobind/pull/688/commits/599b191e84be663ecf4ed8598344243ebac176ac).

wjakob:
> That's a confusing error message, too. It says "incompatible function arguments", but the invoker type is in the list of overloads (no. 4)?

Indeed, that's something not specific to enums. An even more confusing case are nd-arrays with an incompatible shape or layout. There is a tradeoff here: nanobind could have a complex system to communicate why an overload is not selected, but it would come at a cost that I think is incompatible with the design goals of this library.

src/nb_enum.cpp
wjakob: Hi @hpkfft--I think that would be the case if we're doing arithmetic causing overflow with a type, but is that the case? AFAIK we're just casting `long long` / `unsigned long long` into `int64_t` and back.

nicholasjng:
No, I think I'm done for the moment. Thanks for the thorough reviews!

wjakob:
Thank you very much!

PR #691, opened 2024-08-20
Have type_caster for Eigen::Map create its own temp array

### LINKED ISSUE #682: [BUG]: Handling of containers (e.g. std::vector) of Eigen::Ref<const T>
### Problem description

Hello,

I would like to report a bug which happens under specific circumstances:
1. from-Python conversions to types like `std::vector<Eigen::Ref<const MatrixXd>>` (with the default, fixed inner stride),
2. when the Numpy arrays are (non-contiguous) slices of the same array with the same dimensions (see example code),
3. when using versions of Eigen predating [this PR on gitlab](https://gitlab.com/libeigen/eigen/-/issues/2668) (which added the much needed move constructors).

Following @WKarel's work on PR #215, points 1-2 will lead to the creation of a temporary owned object in the `Ref::m_object` protected member. This happens on this line, because `DMap` has dynamic strides, and the `Eigen::Ref` has fixed inner stride hence the temporary is created:
https://github.com/wjakob/nanobind/blob/8e5fd142d9791f3f7ff66485de479578e64d9f37/include/nanobind/eigen/dense.h#L467

I know of two mitigations for now:
- using a version of Eigen that has the move constructors for `Eigen::Ref`
- use `nb::DRef` which was built to avoid copies

### Reproducible example code

```cpp
using MatXdR = Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
using MatRefXdR = Eigen::Ref<const MatXdR>;
using Vec_MatRefXdR = std::vector<MatRefXdR>;
m.def("printRefs", [](const Vec_MatRefXdR& ms) {
     std::cout << "Got matrix\n" << ms[0] << std::endl;
     std::cout << "Got matrix\n" << ms[1] << std::endl;
      // passthrough
      std::vector<MatXdR> out(ms.size());
      for (size_t i = 0; i < ms.size(); i++) {
           out[i] = ms[i];
      }
      return out;
});
```

```python
# in test_eigen.py
@needs_numpy_and_eigen
def test_vec_ref():
    print()
    np.set_printoptions(precision=4, linewidth=200)
    a = np.random.randn(8, 4).T  # .T ensures "wrong" order for slices
    s1 = a[:2]
    s2 = a[2:]
    print(s1)
    print(s2)
    r1, r2 = t.printRefs([s1, s2])
    print(r1)
    print(r2)
    assert np.allclose(s1, r1)
    assert np.allclose(s2, r2)
```

The code prints the following result (and the asserts fail)
```
[[-0.7264  0.4645 -1.3106  1.3194 -1.5611 -1.6445 -0.6314  0.6161]
 [ 0.6566 -0.5272  1.5168  1.2926  0.6595  0.6201 -0.8453  1.0318]]
[[-0.6287  0.4879 -0.8519 -0.2426 -0.0305  0.5116 -1.189  -1.3493]
 [-0.4122 -1.2882 -1.1801  1.0621  0.8644  1.1376 -0.7016  0.2041]]
Got matrix
 -0.628672    0.48795  -0.851887  -0.242632 -0.0305066   0.511559     -1.189   -1.34933
 -0.412173   -1.28822   -1.18013    1.06208   0.864423    1.13762  -0.701575   0.204068
Got matrix
 -0.628672    0.48795  -0.851887  -0.242632 -0.0305066   0.511559     -1.189   -1.34933
 -0.412173   -1.28822   -1.18013    1.06208   0.864423    1.13762  -0.701575   0.204068
[[-0.6287  0.4879 -0.8519 -0.2426 -0.0305  0.5116 -1.189  -1.3493]
 [-0.4122 -1.2882 -1.1801  1.0621  0.8644  1.1376 -0.7016  0.2041]]
[[-0.6287  0.4879 -0.8519 -0.2426 -0.0305  0.5116 -1.189  -1.3493]
 [-0.4122 -1.2882 -1.1801  1.0621  0.8644  1.1376 -0.7016  0.2041]]
```

ManifoldFR:
This PR:
- Has `type_caster<Eigen::Map<...>>` create its own copy of the input `NDArray`
  - with appropriate strides,
  - pointing to data with same lifetime as the function (using the `cleanup_list *cleanup` arg in `from_python`).
  - This will only happen when the inner stride is 1 (maybe we can handle other strides later?), otherwise the caster will return false (`Eigen::Ref` will then dispatch to the `Map` type caster with dynamic strides and rely on its constructor creating an internal copy `Ref::m_object`, which was the previous behaviour)
- adds a test to `test_eigen.cpp|test_eigen.py` which tests a function with argument `std::vector<Eigen::Ref<const VectorXd>>`, which creates a temp copy when an array with the wrong order is given.

Fixes #682

### DIFF
--- a/include/nanobind/eigen/dense.h
+++ b/include/nanobind/eigen/dense.h
@@ -253,17 +253,36 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
     template <typename T_> using Cast = Map;
     template <typename T_> static constexpr bool can_cast() { return true; }
 
+    bool need_copy;
     NDArrayCaster caster;
+    NDArray arr_copy;  // use this if necessary
 
     bool from_python(handle src, uint8_t flags, cleanup_list *cleanup) noexcept {
         // Disable implicit conversions
         return from_python_(src, flags & ~(uint8_t)cast_flags::convert, cleanup);
     }
 
+    /// given inner stride + compile-time info on number of dimensions, shape and row- or col-major,
+    /// get appropriate strides for the temp copy array
+    static void get_auto_strides(const size_t *shape, int64_t *strides, bool inner_stride = 1) {
+        if constexpr (ndim_v<T> == 1) {
+            strides[0] = inner_stride;
+        } else if (!T::IsRowMajor) {
+            // col major: (1, shape[0])
+            strides[0] = inner_stride;
+            strides[1] = shape[0];
+        } else {
+            // row major: (shape[1], 1)
+            strides[0] = shape[1];
+            strides[1] = inner_stride;
+        }
+    }
+
     bool from_python_(handle src, uint8_t flags, cleanup_list* cleanup) noexcept {
         if (!caster.from_python(src, flags, cleanup))
             return false;
 
+        need_copy = false;
         // Check for memory layout compatibility of non-contiguous 'Map' types
         if constexpr (!is_contiguous_v<Map>)  {
             // Dynamic inner strides support any input, check the fixed case
@@ -275,7 +294,9 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
                         is_actual = caster.value.stride(
                             (ndim_v<T> != 1 && T::IsRowMajor) ? 1 : 0);
 
-                if (is_expected != is_actual)
+                need_copy |= is_expected != is_actual;
+                // do not handle non-contiguous inner dimension
+                if (need_copy && (is_expected != 1))
                     return false;
             }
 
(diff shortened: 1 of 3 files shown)


wjakob:
This is a rather intrusive change given that it's only needed to work around a defect in Eigen that has since been fixed. I'm inclined to rather detect the conditions where a problem could occur at compile time, and to then inform the user that they need to update their Eigen version.

Another opinion would be useful -- do you have thoughts on this @WKarel?

WKarel:
https://github.com/wjakob/nanobind/issues/682#issuecomment-2301510278

PR #696, opened 2024-08-21
Add leak/implicit_cast warnings getters

rtabbara:
Can be useful to check the state of these internal nanobind properties if for instance you only want to temporarily disable the warnings in your project and then restore the original state.


wjakob:
Thanks!

PR #698, opened 2024-08-21
Document Meson WrapDB package for nanobind

WillAyd:


docs/api_cmake.rst
```diff
@@ -14,6 +14,11 @@ If you prefer another build system, then you have the following options:
   interface <https://github.com/nicholasjng/nanobind-Bazel>`__ to nanobind.
   Please report Bazel-specific issues there.
 
+- `Will Ayd <https://github.com/WillAyd/>`__ has created a `Meson WrapDB
+  package https://mesonbuild.com/Wrapdb-projects.html>`__ for nanobind. Please
```
henryiii, line 18: Is this a package that lives somewhere? Or is the code in this repo?

WillAyd: The Meson configuration is located in the subsequently linked https://github.com/mesonbuild/wrapdb repository.

You can reference the original PR if you wanted to know more details about how it is laid out https://github.com/mesonbuild/wrapdb/pull/1556

WillAyd: By the way, thanks for scikit-build-core - that's also a great package!

henryiii: Ahh, so I think that it must live there, and adding stuff here won't help it, since needs to be pulled in by meson. And the version number there has to be bumped after releases here. :/

Okay, looks good to me.

WillAyd: Yea that's how things are set up now. You could alternatively decide to host the meson configuration files directly in this repository instead of in the WrapDB, but for now without claiming first class support for Meson it makes sense to have that separate

wjakob:
Looks great, thank you!

PR #701, opened 2024-08-26
Up to 2x faster instance construction using type vector calls

wjakob:
nanobind has always used [PEP 590 vector calls](https://www.python.org/dev/peps/pep-0590) to efficiently dispatch calls to function and method bindings, but it lacked the ability to do so for instance constructors (e.g., ``MyType(arg1, arg2, ...)``).

Version 2.2.0 adds this missing part, which accelerates object construction by up to a factor of 2×. The difference is especially pronounced when passing keyword arguments to constructors. Note that this feature is only supported on Python 3.9+ and when building non-stable ABI extensions. If CPython PR [#123332](https://github.com/python/cpython/pull/123332) is accepted, this fast path might also become available in the stable ABI on Python 3.14+.

### DIFF
--- a/include/nanobind/nb_class.h
+++ b/include/nanobind/nb_class.h
@@ -62,10 +62,12 @@ enum class type_flags : uint32_t {
     is_arithmetic            = (1 << 16),
 
     /// Is the number type underlying the enumeration signed?
-    is_signed                = (1 << 17)
+    is_signed                = (1 << 17),
 
-    // One more flag bits available (18) without needing
-    // a larger reorganization
+    /// Does the type implement a custom __new__ operator?
+    has_new                  = (1 << 18)
+
+    // No more bits bits available without needing a larger reorganization
 };
 
 /// Flags about a type that are only relevant when it is being created.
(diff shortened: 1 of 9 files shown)


include/nanobind/nb_class.h
```diff
@@ -62,10 +62,12 @@ enum class type_flags : uint32_t {
...
     /// Is the number type underlying the enumeration signed?
-    is_signed                = (1 << 17)
+    is_signed                = (1 << 17),
 
-    // One more flag bits available (18) without needing
-    // a larger reorganization
+    /// Does the type implement a custom __new__ operator?
+    has_new                  = (1 << 18)
+
+    // No more bits bits available without needing a larger reorganization
```
nicholasjng, line 70: This will stall the flag enums PR, since we're now out of bits for `is_flag`, right?

wjakob: Addressed by 72eb1b236326d77d0ff90b39b69312f2239f1527.

nicholasjng: Thank you!

PR #703, opened 2024-08-29
Fix importing `TypeAlias` in stubgen

yosh-matsuda:
Fix the import problem in the stub file for old Python versions as `TypeAlias` has been available since 3.10.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -479,7 +479,10 @@ class StubGen:
 
             if same_module:
                 # This is an alias of a type in the same module or same top-level module
-                alias_tp = self.import_object("typing", "TypeAlias")
+                if sys.version_info >= (3, 10, 0):
+                    alias_tp = self.import_object("typing", "TypeAlias")
+                else:
+                    alias_tp = self.import_object("typing_extensions", "TypeAlias")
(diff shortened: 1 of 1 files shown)


wjakob:
The python typing specs says that it's fine to use constructs from newer Python versions, as the static type checkers in any case don't run on the same python version has the host program. So I'm not sure if this is the right way to go.

yosh-matsuda:
I didn't know that.

Actually, mypy complains about this like

```
$ mypy --python-version 3.9 test.py                                                                                                                                                                                                                                                                                                                                                                    nb_module/__init__.pyi:6: error: Module "typing" has no attribute "TypeAlias"  [attr-defined]
nb_module/__init__.pyi:6: note: Use `from typing_extensions import TypeAlias` instead
nb_module/__init__.pyi:6: note: See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module
...
```

(so I thought it should be fixed) but pyright doesn't.

To avoid annoyance (for mypy users), it may be better to match the version of the module and the description of the stub file. Stub files are generated and distributed for each Python version.

wjakob:
Fair enough, let's fix it if MyPy is unhappy about this.

PR #704, opened 2024-08-29
Fix extraction of class name with new parameter syntax

yosh-matsuda:
I found that a compilation error occurs when the [new parameter syntax for generics](https://docs.python.org/3.12/reference/compound_stmts.html#type-params) is used as a signature. 
 
This fix adds the searching for `[` to the `extract_name` function.

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -810,11 +810,16 @@ NB_NOINLINE char *extract_name(const char *cmd, const char *prefix, const char *
           cmd, s, prefix);
     p += prefix_len;
 
-    // Find the opening parenthesis
+    // Find the opening parenthesis or bracket
     const char *p2 = strchr(p, '(');
+    const char *p3 = strchr(p, '[');
+    if (p2 == nullptr)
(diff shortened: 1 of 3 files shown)


src/nb_type.cpp
```diff
@@ -810,11 +810,20 @@ NB_NOINLINE char *extract_name(const char *cmd, const char *prefix, const char *
           cmd, s, prefix);
     p += prefix_len;
 
-    // Find the opening parenthesis
+    // Find the opening parenthesis or bracket
+#if PY_VERSION_HEX >= 0x030C0000
+    const char *p2 = strchr(p, '(');
+    const char *p3 = strchr(p, '[');
+    p2 = p2 == nullptr ? p3 : (p3 == nullptr ? p2 : (p2 < p3 ? p2 : p3));
```
wjakob, line 817: I think it's fine to have just one version of this code.

Can you write the ternary expression using an `if` statement? It's a little hard to read.

wjakob:
Perfect, thank you!

PR #705, opened 2024-08-29
Print `...` for ellipsis type in type parameters

yosh-matsuda:
I want to print `...` as a parameter of `nb::typed`, e.g. `tuple[int, ...]`, but using `nb::ellipsis` shows its type, `EllipsisType`.

In this PR change, I have changed the type string to `...` only when `nb::ellipsis` is in `nb::typed`. 

Is this acceptable?


include/nanobind/nb_cast.h
```diff
@@ -354,12 +354,21 @@ NB_TYPED_NAME_PYTHON38(type_object, NB_TYPING_TYPE)
 #endif
 
 template <typename T, typename... Ts> struct type_caster<typed<T, Ts...>> {
+private:
+    template<typename U> static constexpr auto param_name() {
```
wjakob, line 358: No need to create a `constexpr` function. `nb_descr.h` already defines

```cpp
// Ternary description (like std::conditional)
template <bool B, size_t N1, size_t N2>
constexpr auto const_name(char const(&text1)[N1], char const(&text2)[N2]) {
    (void) text1; (void) text2;

    if constexpr(B)
        return const_name(text1);
    else
        return const_name(text2);
}
```

wjakob:
Needs a test.

yosh-matsuda:
Thank you for reviewing. It is fixed, and the test is added.

wjakob:
Great, thank you!

PR #715, opened 2024-09-12
Avoid undefined behavior when subtracting tagged pointers

thurstond:
When compiling with HardwareAddressSanitizer (https://clang.llvm.org/docs/HardwareAssistedAddressSanitizerDesign.html), the top byte of pointers are tagged, hence subtracting two pointers may overflow. Overflow is undefined behavior for signed integers, resulting in errors such as:
```
    third_party/nanobind/src/nb_type.cpp:123:50: runtime error: signed integer overflow: -6701078944169102656 - 9079534012674239392 cannot be represented in type 'intptr_t' (aka 'long')
```

This patch fixes the issue by using unsigned integers, for which overflow is well-defined.

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -124,7 +124,10 @@ PyObject *inst_new_ext(PyTypeObject *tp, void *value) {
     }
 
     // Compute offset to instance value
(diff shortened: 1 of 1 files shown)


wjakob:
Thank you!

PR #718, opened 2024-09-15
Fix py_stub_test dependency in out-of-source builds

mgorny:
Fix the dependency between py_stub stubgen invocation and `py_stub_test.py` source file when building out of source. Previously, the dependency would be on the original file in the source directory rather than the one copied to build directory, which could lead to stubgen being invoked prior to it being copied and therefore failing to import it.


wjakob:
Good catch, thanks!

mgorny:
Thanks!

PR #719, opened 2024-09-16
CI: test old GCC and Clang compilers

wojdyr:
They are prone to failing when the code is changing and they happen to be available on the ubuntu-20.04 worker.
Clang 7 and 8 are currently commented out because they fail (#713).


wjakob:
Thanks!

Issue #487, opened 2024-03-21
[BUG]: The stride of an ndarray can be wrong for numpy views


hpkfft:
### Problem description

Numpy allows a new view of an array with the same data, which can cause a reinterpretation of the bytes in memory.
In particular, an array with 2N real values per row can be reinterpreted as an array with N complex values per row.
[This is useful for FFTs of real data, as the result is complex-valued, and since the Fourier transform is reversible, there is not more "data content" in the output than in the input.  So, numpy's views do the right thing.]

The issue arises when the stride from one row to the next is an odd number of real-valued elements.   Numpy measures strides in bytes, so a view does not change the stride and it just works.   In nanobind, strides are measured in units of  `itemsize`, so an nb::ndarray cannot correctly represent a stride that is not an integral multiple of the size of the `dtype`.

### Reproducible example code

```text
C++

    m.def("inspect", [](const nb::ndarray<nb::ro>& a) {
            printf("Array on nb::device::%s%s (id %u) at address %p\n",
                   (a.device_type() == nb::device::cpu::value)  ? "cpu"  : "",
                   (a.device_type() == nb::device::cuda::value) ? "cuda" : "",
                   a.device_id(), a.data());
            printf("    dtype : %s%s%s%s   itemsize : %2zu\n",
                   (a.dtype() == nb::dtype<float>())    ? "float32" : "",
                   (a.dtype() == nb::dtype<double>())   ? "float64" : "",
                   (a.dtype() == nb::dtype<std::complex<float>>())
                                             ? "complex<float32>" : "",
                   (a.dtype() == nb::dtype<std::complex<double>>())
                                             ? "complex<float64>" : "",
                   a.itemsize());
            printf("    ndim : %zu\n", a.ndim());
            for (size_t i = 0; i < a.ndim(); ++i) {
                printf("    shape(%zu) : %2zu    stride(%zu) : %2zd\n",
                              i,  a.shape(i),       i,  a.stride(i));
            }
        });
```

Python (interactive session):
```
>>> a = np.array([[1, 2, 3, 4, 5, 6, np.NAN],
...               [8, 0, 0, 0, 0, 0, np.NAN]], dtype=np.float32)
>>> a
array([[ 1.,  2.,  3.,  4.,  5.,  6., nan],
       [ 8.,  0.,  0.,  0.,  0.,  0., nan]], dtype=float32)
>>> m.inspect(a)
Array on nb::device::cpu (id 0) at address 0x1a5ddb0
[...]

PR #721, opened 2024-09-18
``nb::ndarray<...>`` implementation refactored

wjakob:
This PR refactors the ``nb::ndarray`` implementation to remove certain redundancies. In particular, there were duplicate code paths to process ``ndarray`` template parameters at compile time and at runtime, which are now merged.

Significant edits to the documentation are intended to make nd-array bindings more approachable to newcomers. Finally, the refactor was also an opportunity to realize two usability improvements:

1. The constructor to return new nd-arrays from C++ now considers all template arguments:

   - **Memory order**: ``c_contig``, ``f_contig``.
   - **Shape**: ``nb::shape<3, 4, 5>``, etc.
   - **Device type**: ``nb::device::cpu``, ``nb::device::cuda``, etc.
   - **Framework**: ``nb::numpy``, ``nb::pytorch``, etc.
   - **Data type**: ``uint64_t``, ``std::complex<double>``, etc.

   Previously, only the **framework** and **data type** annotations were taken into account when returning nd-arrays, while all of them were examined when *accepting* arrays during overload resolution. This inconsistency was a repeated source of confusion among users.

   To give an example, the following now works out of the box without the need to redundantly specify the shape and strides to the ``Array`` constructor below:

   ```cpp
   using Array = nb::ndarray<float, nb::numpy, nb::shape<4, 4>, nb::f_contig>;

   struct Matrix4f {
       float m[4][4];
       Array data() { return Array(m); }
    };

   nb::class_<Matrix4f>(m, "Matrix4f")
       .def("data", &Matrix4f::data, nb::rv_policy::reference_internal);
   ```

2. A new nd-array ``.cast()`` method forces the immediate creation of a Python object with the specified target framework and return value policy, while preserving the type signature in return values. This is useful to return temporaries (e.g. stack-allocated memory) from functions.

There are two minor but potentially breaking changes:

1. The ndarray type caster now interprets the ``rv_policy::automatic_reference`` return value policy analogously to the ``rv_policy::automatic``, which means that it references a memory region when the user specifies an ``owner``, and it otherwise copies. This makes it safe to use the ``nb::cast()`` and ``nb::ndarray::cast()`` functions that use this policy as a default.

2. The ``nb::any_contig`` memory order annotation, which previously did nothing, now accepts C- or F-contiguous arrays and rejects non-contiguous ones.

In both of these cases, the prior convention seems like it would cause bugs/breakage in practice. If nobody depends on this behavior, it should be OK to fix these without a major version bump.

A small change without compatibility implications: the `owner` argument has a default `{}` argument again. I think this is reasonably safe because the `automatic_*` return value policies in nanobind will copy the input array if there isn't an owner. This effectively reverts a change from commit 937a1df52428f04a0a240cc5827fb3c6769cf5c2.


wjakob:
I did a pass over the nd-array implementation to fix various things that bothered me and caused confusion amongst users. While it's a quite big change, I am hoping that this won't break anyone's code -- it would be great if you could try the `ndarray` branch and report back.

Tagging folks who are using the ndarray-feature according to the GitHub issue tracker and discussions: @WKarel, @ManifoldFR, @hpkfft, @yosh-matsuda, @garth-wells, @luigifcruz, @awni, @qnzhou, @skallweitNV, @oremanj, @wojdyr, @jmalkin, @Chronum94

awni:
Thanks for the heads up. I built MLX with this branch and ran our test suite and it cleared ✅

wojdyr:
Sounds great. I was one of the many confused people when I started using it. Another gotcha is, when porting from pybind11, that in pybind11 strides are in bytes and in nanobind – in the number of items.

I tested it with my bindings and haven't seen any problems.

wjakob:
> Another gotcha is, when porting from pybind11, that in pybind11 strides are in bytes and in nanobind – in the number of items.

This is a design decision of DLPack that ripples down to nanobind. I will mention it in the documentation.

ManifoldFR:
Looks great ! Great changes to the docs.
I built proxsuite with this branch and all our tests still pass, as expected 👍🏼

docs/ndarray.rst
```diff
@@ -79,374 +79,329 @@ Array constraints
...
+input. This produces the following error message:
 
-Note that the view performs no reference counting. You may not store it in a way
-that exceeds the lifetime of the original array.
+.. code-block:: pycon
 
-When using OpenMP to parallelize expensive array operations, pass the
-``firstprivate(view_1, view_2, ...)`` so that each worker thread can copy the
-view into its register file.
+   >>> my_module.process(nd-array=np.zeros(1))
```
oremanj: another search/replace that shouldn't've; also in the error message below

docs/ndarray.rst
```diff
@@ -79,374 +79,329 @@ Array constraints
...
+underlying buffer and increase the reference count until they go out of scope.
+You may call freely call :cpp:class:`nb::ndarray\<...\> <ndarray>` methods from
+multithreaded code even when the `GIL
+<https://wiki.python.org/moin/GlobalInterpreterLock>`__ is not held, for
+example to examine the layout of an array and access the underlying storage.
 
-   void fill(nb::ndarray<nb::c_contig, nb::device::cpu> arg) {
-       if (arg.dtype() == nb::dtype<float>() && arg.ndim() == 2) {
-           auto v = arg.view<float, nb::ndim<2>>(); // <-- new!
+There are two exceptions to this: creating *new* nd-array object from C++
```
oremanj, line 207:
```suggestion
There are two exceptions to this: creating a *new* nd-array object from C++
```

docs/ndarray.rst
```diff
@@ -79,374 +79,329 @@ Array constraints
...
+(discused :ref:`later <returning-nd-arrays>`) and casting it to Python via
+:cpp:func:`ndarray::cast` function both involve Python API calls that require
```
oremanj, line 209:
```suggestion
the :cpp:func:`ndarray::cast` function both involve Python API calls that require
```

docs/ndarray.rst
```diff
@@ -79,374 +79,329 @@ Array constraints
...
-C-style ordering. Both ``strides`` and ``shape`` will be copied by the
-constructor, hence the targets of these pointers don't need to remain valid
-following the call.
-
-An alternative form of the constructor takes ``std::initializer_list`` instead
-of shape/stride arrays for brace-initialization and infers ``ndim``:
+The previous example was rather simple because all the array configuration was fully known
+at compile time and specified via the :cpp:class:`nb::ndarray\<...\> <ndarray>`
+template parameters. In general, there are often dynamic aspects of the
+configuration that must be explicitly passed to the constructor. It signature
```
oremanj, line 301:
```suggestion
configuration that must be explicitly passed to the constructor. Its signature
```

docs/ndarray.rst
```diff
@@ -79,374 +79,329 @@ Array constraints
...
            handle owner,
-           st::initializer_list<int64_t> strides = { },
-           dlpack::dtype dtype = nb::dtype<Scalar>(),
-           int32_t device_type = nb::device::cpu::value,
-           int32_t device_id = 0) { .. }
+           std::initializer_list<int64_t> strides,
+           dlpack::dtype dtype,
+           int device_type,
+           int device_id,
+           char order) { .. }
```
oremanj, line 315:
```suggestion
           std::initializer_list<int64_t> strides = ...,
           dlpack::dtype dtype = ...,
           int device_type = ...,
           int device_id = ...,
           char order = ...) { .. }
```
Maybe indicate which ones have defaults?

docs/api_extra.rst
```diff
@@ -824,6 +881,17 @@ section <ndarrays>`.
...
+   .. cpp:function:: auto cast(rv_policy policy = rv_policy::automatic_reference, handle parent = {})
+
+      The expression ``array.cast(policy, handle)`` is almost equivalent to
+      :cpp:func:`nb::cast(array, policy, handle) <cast>`.
+
+      The main difference is that the return type of :cpp:func:`nb::cast
+      <cast>` is :cpp:class:`nb::object <object>`, which renders as a rather
+      non-descriptive ``object`` in Python bindings. The ``.cast()`` method
+      instead returns a wrapper type, whose signature matches that of the
+      original nd-array.
```
oremanj, line 893:
```suggestion
      instead returns a wrapper type that will be rendered in function signatures
      with the type of the original nd-array.
```
I didn't understand at first what it meant for the wrapper type to have a signature

docs/ndarray.rst
```diff
@@ -516,7 +529,7 @@ For example, the following snippet makes ``__fp16`` (half-precision type on
 .. code-block:: cpp
 
    namespace nanobind {
-       template <> struct ndarray_traits<__fp16> {
+       template <> struct nd-array_traits<__fp16> {
```
oremanj: another search/replace

docs/ndarray.rst
```diff
@@ -525,9 +538,104 @@ For example, the following snippet makes ``__fp16`` (half-precision type on
...
+       for (size_t i = 0; i < arg.shape(0); ++i)
+           for (size_t j = 0; j < arg.shape(1); ++j)
+               arg(i, j) = /* ... */;
+   }
+
+While functional, this code is not perfect. The problem is that to compute the
+address of an entry, ``operator()`` accesses the DLPack array descriptor. This
+indirection can break certain compiler optimizations.
+
+nanobind provides the method :cpp:func:`nd-array\<...\>::view() <ndarray::view>`
```
oremanj: another

include/nanobind/ndarray.h
```diff
@@ -105,261 +141,125 @@ template <size_t... S> struct ndim_shape<std::index_sequence<S...>> {
...
+    using DeviceType = unused;
+    static constexpr int32_t N = -1;
 };
 
-template <ssize_t... Is, typename... Ts> struct ndarray_info<shape<Is...>, Ts...> : ndarray_info<Ts...> {
-    using shape_type =
-        std::conditional_t<
-                std::is_void_v<typename ndarray_info<Ts...>::shape_type>,
-                shape<Is...>, typename ndarray_info<Ts...>::shape_type>;
+// Template infrastructure to collect ndarray annotations and fail if duplicates are found
```
oremanj, line 220: Does this still fail if duplicates are found? It doesn't appear so

tests/test_ndarray.py
```diff
@@ -850,3 +875,55 @@ def test_conv(x):
...
+    bd = a.data_ref()
+    for i in range(16):
+        ad[i%4, i//4] = i
+    del a, ad
+    for i in range(16):
+        assert bd[i%4, i//4] == i
+
+@needs_numpy
+def test46_test_matrix4f_copy():
+    assert t.Matrix4f.data_ref.__doc__.replace('data_ref', 'data') == t.Matrix4f.data.__doc__
```
oremanj, line 912:
```suggestion
    assert t.Matrix4f.data_ref.__doc__.replace('data_copy', 'data') == t.Matrix4f.data.__doc__
```

hpkfft:
Hi @wjakob, Thanks for inviting me to test and comment.  I successfully built and tested my FFT extension with this branch without needing any source code changes on my part.

Hi @wojdyr, Regarding strides measured in bytes, you may be interested in [dlpack strides](https://dmlc.github.io/dlpack/latest/c_api.html#c.DLTensor.strides) and in discussion #487

include/nanobind/ndarray.h
wjakob: No, I removed this part of the code. It seemed too much complexity for too little benefit.

docs/api_extra.rst
```diff
@@ -725,21 +748,52 @@ section <ndarrays>`.
...
       ``shape``.
 
+      Both ``strides`` and ``shape`` will be copied by the constructor, hence
+      the targets of these initializer lists don't need to remain valid
+      following the constructor call.
+
+      The Python *global interpreter lock* (GIL) must be held when calling this
+      function.
+
+   .. cpp:function:: ndarray(VoidPtr data)
```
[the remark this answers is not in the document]
wjakob: I really like the ``nb::ndarray<...>`` compile-time template interface. One can simply dump some constraints in any order. nanobind will enforce them in bindings, arrange them consistently in docstrings, and use defaults for missing attributes.

The ``nb::ndarray()`` run-time constructor is super-restrictive in comparison. You need to provide parameters in a certain order and can't skip irrelevant ones.  Ideally I would have a *single* template constructor that takes anything and detects what it is given, but this is tricky to fix now because several of these parameters have similar types. (e.g. integers or lists of integers).

wjakob:
Thank you for the feedback @oremanj!

docs/api_extra.rst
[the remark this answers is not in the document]
wjakob: I gave it another attempt in 3f835139e89ae0b3b8caaed3586f54937f36c819. Still not as convenient as I would ideally like it, but this is probably the best that can be done with backward compatibility in mind.

yosh-matsuda:
@wjakob Hi, this change works well in my project. Thank you!

PR #723, opened 2024-09-18
bind_map and bind_vector: add sig to __eq__ and __ne__

wojdyr:
Use the sig attribute to avoid mypy [override] error.

Following the advice from https://nanobind.readthedocs.io/en/latest/typing.html#functions


wjakob:
Thanks!

PR #727, opened 2024-09-19
Set singles mask only if already present on the enum

nicholasjng:
The singles mask attribute was added for the STRICT boundary default, but only starting at Python 3.11.4. This means that for any Python 3.11 before that, we need to avoid updating it (accessing it in the process), which would result in an attribute error.

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -126,7 +126,7 @@ void enum_append(PyObject *tp_, const char *name_, int64_t value_,
         setattr(tp, "_flag_mask_", tp.attr("_flag_mask_") | val);
 
         bool is_single_bit = (value_ != 0) && (value_ & (value_ - 1)) == 0;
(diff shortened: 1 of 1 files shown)


nicholasjng:
NB: I'm not really sure if this is the right fix, since that relies on the enum type being constructed correctly already. We could still update the mask unconditionally by doing something like `setattr(tp, "_singles_mask_", hasattr(tp, "_singles_mask_") ? tp.attr("_singles_mask_") | val : 0);`

nicholasjng:
@hpkfft could you build with this branch and see if that fixes the issue?

wjakob:
@hpkfft Also, can you pass the test suite with this Python version and fix to nanobind?

hpkfft:
Yes, all is well with your fix.
The testsuite passes: 351 passed, 88 skipped.   (I don't have eigen or cupy installed.)

PR #732, opened 2024-09-24
Fix enum flag from cpp conversion

skallweitNV:
- Use `__str__` from the underlying flag factory class (allows printing flags like `Flag.A|B|C`)
- Fix conversion from cpp -> python of enum flags that contain a combination of flags.


skallweitNV:
Hmm, looks like `__str__` has changed in Python >= 3.11. It now returns numeric values for the `Int` variants.

src/nb_enum.cpp
```diff
@@ -71,7 +71,7 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
     scope.attr(name) = result;
     result.attr("__doc__") = ed->docstr ? str(ed->docstr) : none();
 
-    result.attr("__str__") = enum_mod.attr("Enum").attr("__str__");
```
wjakob, line 74: Can you use `factor_name` (and otherwise `Enum`) only when this is a flag?

wjakob (commented):
Could you add a testcase that checks the behavior of the flag `__str__` routine?

src/nb_enum.cpp
skallweitNV: Done.

skallweitNV:
@wjakob this should be good to merge now.

PR #733, opened 2024-09-25
Fix indexing into args array

dvicini:
Minor fix to indexing into args array: 1) fix the use of negative indices and 2) use args_in instead of f->args.


wjakob:
Thank you very much for catching this, Delio! Before seeing your PR, I had amended my commit to turn it back into the original version you had posted on Slack.

dvicini:
Thanks, sounds good then. I will close this then.

PR #734, opened 2024-09-26
Fix unneeded semicolons triggering warnings with pedantic enabled

matthewdcong:
I know we're not looking to support a clean build with `-pedantic` enabled but this is a mostly cosmetic fix in the right direction.


wjakob:
yep, these are fine -- thanks!

PR #735, opened 2024-09-26
stubgen.py: handle method aliases similarly to function aliases

wojdyr:
Added support for method aliases (in the same class):

    A.attr("old_name") = A.attr("new_name")

will generate in class A in pyi:

    old_name = new_name


wjakob:
What if the method is a function or a method from another class? It seems to me like the name of the initializer must have the ``__qualname__`` if it is different from where it is being inserted.

Also, could there be issues when the method references something that hasn't been declared yet? What if the stubs say

```
class A:
    f = A.g
    def g(self): ...
```

This would be an error in Python. Stubs are sometimes a bit more forgiving, but I'm wondering if it is the case here.

wojdyr:
I changed it to `__qualname__` when it refers a different class.
Or should it be always `__qualname__`?

As for the declaration order, no idea. If it's wrong, the user can just change the order in cpp.

wjakob:
Sorry about the huge delay. I looked at this again today and was still not 100% convinced. In particular, I did not understand why this code is specific to methods. Top-level nanobind functions could also be aliases and should then be arguably handled the same way. Do you think there is a way of implementing this alias detection logic so that it is part of the `nb_func` handling code?

wojdyr:
Functions are sent to `put_function` and aliases are handled there.
Methods are sent directly to `put_nb_func`.
I just changed this PR to send methods also to `put_function`, and this fixes the aliasing case that I'm interested in.
I tested it on my project and on nanobind tests – pyi files are otherwise unchanged. But it may have side-effects that I'm not aware of.
So now it's only one-line change in stubgen.py.

wjakob:
This looks good to me. For consistency and resilience to future changes, I think that the following change should additionally be applied:

```diff --git a/src/stubgen.py b/src/stubgen.py
index 29400bf..217e8ba 100755
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -395,6 +395,12 @@ class StubGen:
             self.write_ln(f"{name} = {fn_name}\n")
             return

+        # Special handling for nanobind functions with overloads
+        if type(fn).__module__ == "nanobind":
+            fn = cast(NbFunction, fn)
+            self.put_nb_func(fn, name)
+            return
+
         if isinstance(fn, staticmethod):
             self.write_ln("@staticmethod")
             fn = fn.__func__
@@ -402,12 +408,6 @@ class StubGen:
             self.write_ln("@classmethod")
             fn = fn.__func__

-        # Special handling for nanobind functions with overloads
-        if type(fn).__module__ == "nanobind":
-            fn = cast(NbFunction, fn)
-            self.put_nb_func(fn, name)
-            return
-
         if name is None:
             name = fn.__name__
             assert name
```

This is because `put_nb_func` also adds the `@staticmethod` prefix.

wjakob:
I will merge this and apply that change directly.

PR #745, opened 2024-10-03
Switch to integer data type in Eigen test for default arguments

dfm:
The `test10_eigen_scalar_default` test in `test_eigen.py` relies unnecessarily on exact equality of floating point numbers. In my environment, with a recent commit of Eigen, the addition in the test is causing the equality check to fail. There might be other upstream issues to consider here, but the specific test failure here seems to be a false positive! This PR switches that test to use integer zeros to be less sensitive to numerics. Let me know what you think - thanks!


dfm:
Closing because there's probably some other issue with my environment. Sorry for the noise!

Issue #746, opened 2024-10-03
[BUG]: Eigen integration silently fails with recent versions of Eigen


dfm:
### Problem description

When I build the following simple nanobind module with the current `master` branch of Eigen:

```c++
#include <nanobind/nanobind.h>
#include <nanobind/eigen/dense.h>

namespace nb = nanobind;

NB_MODULE(nanobind_example_ext, m) {
    using Matrix = Eigen::Matrix<int, 1, 1>;
    m.def("add", [](Matrix a, Matrix b) { return a + b; });
}
```

The output of calling:

```python
import numpy as np
nanobind_example.add(np.int32([0]), np.int32([0]))
```

is `array([17726950], dtype=int32)` instead of the expected `array([0], dtype=int32)`.

This can be fixed by adding a return type:

```diff
-     m.def("add", [](Matrix a, Matrix b) { return a + b; });
+     m.def("add", [](Matrix a, Matrix b) -> Matrix { return a + b; });
```

and the example above works properly using the last stable release of Eigen, so I expect it has something to do with a change in the type produced by the addition. Unfortunately I don't know enough about the nanobind (or Eigen!) internals to know where to start debugging this, but I wanted to check here if this was an expected sharp edge, or if this is something that should work.

I note that I originally came across this issue when I discovered a test failure here:

https://github.com/wjakob/nanobind/blob/f3e2796d7bd12f77856eef53772644115a5e8788/tests/test_eigen.py#L303-L305

which has this same syntax.

### Reproducible example code

```c++
#include <nanobind/nanobind.h>
#include <nanobind/eigen/dense.h>

namespace nb = nanobind;

NB_MODULE(nanobind_example_ext, m) {
    using Matrix = Eigen::Matrix<int, 1, 1>;
    m.def("add", [](Matrix a, Matrix b) { return a + b; });
}
```

```python
import numpy as np
assert nanobind_example.add(np.int32([0]), np.int32([0])) == 0
```

wjakob:
I can't reproduce this issue on my machine. What version of Eigen are you using? What OS is this on?

wjakob:
[...]

PR #749, opened 2024-10-04
Add a CI workflow for building against Eigen from source

dfm:
This isn't really meant to be merged, but it's a follow up from https://github.com/wjakob/nanobind/issues/746 demonstrating the test failure with the current master branch of Eigen. This workflow seems to fail when run on macOS (but not on Ubuntu), and (as @hawkinsp suggested) I expect the issue is related to using clang instead of gcc.

One way to "fix" this test failure is to update this function:

https://github.com/wjakob/nanobind/blob/bff96e268749e788b69153221c1546b8904a9b53/tests/test_eigen.cpp#L172-L179

as follows

```diff
-             "default_arg", [](Matrix1d a, Matrix1d b) { return a + b; },
+             "default_arg", [](Matrix1d a, Matrix1d b) -> Matrix1d { return a + b; },
```

but I think there's probably something more complicated happening here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,16 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'windows-2022', 'macos-13']
         python: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13.0-rc.2', 'pypy3.9-v7.3.16', 'pypy3.10-v7.3.17']
+        eigen-from-source: [false]
+        include:
+        - os: 'macos-13'
+          python: '3.12'
+          eigen-from-source: true
+        - os: 'ubuntu-latest'
+          python: '3.12'
+          eigen-from-source: true
 
-    name: "Python ${{ matrix.python }} / ${{ matrix.os }}"
+    name: "Python ${{ matrix.python }} / ${{ matrix.os }} ${{ (matrix.eigen-from-source && '(eigen from source)') || '' }}"
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -41,8 +49,17 @@ jobs:
       uses: lukka/get-cmake@latest
 
     - name: Install Eigen
-      if: matrix.os == 'ubuntu-latest'
+      if: ${{ matrix.os == 'ubuntu-latest'  && !matrix.eigen-from-source }}
       run: sudo apt-get -y install libeigen3-dev
+
+    - name: Install Eigen from source
+      if: matrix.eigen-from-source
+      run: |
+        git clone --depth 1 https://gitlab.com/libeigen/eigen.git
+        mkdir eigen/build
+        cd eigen/build
+        cmake ..
+        sudo make install
 
     - name: Install PyTest
       run: |


wjakob:
That looks concerning since it affects such basic types. Let me add @WKarel as well.

WKarel:
`a + b` should create an instance of a specialization of `Eigen::CwiseBinaryOp` for scalar sums and the given specialization of `Eigen::Matrix`. Upon return, this should call nanobind's specialization of `type_caster` for dense Eigen expressions, which should forward to the specialization of `type_caster` for `Eigen::Array<double, 1, 1>`.

What happens if you replace the trailing return type `Eigen::Matrix<double,1,1>` with `Eigen::Array<double,1,1>`?

I do not see why `type_caster<dense expression>::from_cpp` is a function template:
```cpp
    template <typename T2>
    static handle from_cpp(T2 &&v, rv_policy policy, cleanup_list *cleanup) noexcept {
        return Caster::from_cpp(std::forward<T2>(v), policy, cleanup);
    }
```

Replacing it with 
```cpp
    static handle from_cpp(T&& v, rv_policy policy, cleanup_list* cleanup) noexcept {
      return Caster::from_cpp(std::move(v), policy, cleanup);
    }
    static handle from_cpp(const T& v, rv_policy policy, cleanup_list* cleanup) noexcept {
      return Caster::from_cpp(v, policy, cleanup);
    }
```
might actually fix the issue with CLang on MacOS.

dfm:
Thanks for the suggestions, @WKarel!

> What happens if you replace the trailing return type `Eigen::Matrix<double,1,1>` with `Eigen::Array<double,1,1>`?

I can confirm that updating the test as follows also fixes the error:

```diff
-             "default_arg", [](Matrix1d a, Matrix1d b) { return a + b; },
+             "default_arg", [](Matrix1d a, Matrix1d b) -> Eigen::Array<double, 1, 1> { return a + b; },
```

However, the error persists if `Matrix1d` in the test is redefined as `Eigen::Array<double, 1, 1>`, but the return type is omitted.

> Replacing it with ... might actually fix the issue with CLang on MacOS.

I tried updating the following lines (I think this is what you were referring to):

https://github.com/wjakob/nanobind/blob/5ead8f0d7f0edc6684920bc2a6f91b516aa7d2e9/include/nanobind/eigen/dense.h#L228-L231

with the version you suggest, unfortunately that doesn't seem to fix the problem!

wjakob:
@dfm: could you please provide the `typeid()` output as suggested by @WKarel?

dfm:
@wjakob, @WKarel — I'm so sorry about disappearing here! I'll see if I can make any progress debugging this later today or tomorrow, but in the meantime, here's what I get for that `typeid` output when the test fails:

```
N5Eigen13CwiseBinaryOpINS_8internal13scalar_sum_opIddEEKNS_6MatrixIdLi1ELi1ELi0ELi1ELi1EEES6_EE
```

If that immediately suggests anything to try, let me know!

wjakob:
Can you run that through `c++filt` or `c++filt -t` on your machine? Just run the command, enter that long string, and post what this outputs.

dfm:
Absolutely! That prints:

```
Eigen::CwiseBinaryOp<Eigen::internal::scalar_sum_op<double, double>, Eigen::Matrix<double, 1, 1, 0, 1, 1> const, Eigen::Matrix<double, 1, 1, 0, 1, 1> const>
```

dfm:
> Does this test also fail in an unoptimized build?

Good question and sorry for the slow response! It looks like switching to a Debug build _fixes_ the error.

wjakob:
I finally had some time to look into this. This is buggy code and not a nanobind issue -- see here for an explanation: f153ae1976ee87c704c01f1447ddcb05a3120e05

WKarel:
One note: I now understand the underlying problem. In fact, the type_caster for dense Eigen expressions must never be called if the expression references Eigen instances that are local variables, including function arguments. Instead:
- explicitly specify a plain Eigen type as return type (lambda function or not). Or
- call `Eigen::DenseBase::eval()`, e.g. `(a + b).eval()`, which is equivalent, but seems more compact and flexible.
- Or if the expression to be returned only references function arguments, then expect references as arguments, not instances. This is safe, because the tuple of in-casters keeps the referenced objects alive until after the out-caster has evaluated the returned expression.

wjakob:
Exactly. These are actually some good points, I will try to incorporate that into the docs.

PR #753, opened 2024-10-06
Add partial specialization dtype_traits<const T>.

hpkfft:
If a developer provides a specialization for, say, `dtype_traits<__fp16>` then this commit avoids his having also to provide a specialization for `dtype_traits<const __fp16>`.

### DIFF
--- a/include/nanobind/ndarray.h
+++ b/include/nanobind/ndarray.h
@@ -137,9 +137,9 @@ template <> struct dtype_traits<void> {
     static constexpr auto name = descr<0>();
 };
 
-template <> struct dtype_traits<const void> {
-    static constexpr dlpack::dtype value{ 0, 0, 0 };
-    static constexpr auto name = descr<0>();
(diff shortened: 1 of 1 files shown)


hpkfft:
As an aside, the C language has `_Float16` defined in technical specification ISO/IEC TS 18661-3:2015.
Arm recommends that new code use the `_Float16` data type instead of the `__fp16` data type. 
See: [Arm Half-precision](https://developer.arm.com/documentation/102284/6-16-2LTS/armclang-Reference/Other-Compiler-specific-Features/Half-precision-floating-point-data-types?lang=en)

C++23 has `std::float16_t` as well as `std::bfloat16_t`.  See: [Fixed width floating-point types](https://en.cppreference.com/w/cpp/types/floating-point)

wjakob:
This is great, thank you!

I am happy to accept further specializations, but they must be valid in c++17.

PR #756, opened 2024-10-08
Add IWYU export pragmas to protect internal headers

adzenith:
We use [Include What You Use](https://github.com/include-what-you-use/include-what-you-use) to manage `#include`s across a large codebase. IWYU automatically adds and removes `#include` statements in each source file to match what symbols are being used in that file. By default it tries to directly include internal nanobind headers when it detects symbols from those files.

This PR adds [`export` pragmas](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export) to tell IWYU basically the same thing as the comment on line 38 tells humans: that `nanobind.h` itself exports all of the internal symbols, and that the transitive headers should not be directly included.

We've been carrying this patch for a while and it's been working great for us.

Thanks for making nanobind!

### DIFF
--- a/include/nanobind/nanobind.h
+++ b/include/nanobind/nanobind.h
@@ -36,6 +36,7 @@
 #include <new>
 
 // Implementation. The nb_*.h files should only be included through nanobind.h
+// IWYU pragma: begin_exports
 #include "nb_python.h"
 #include "nb_defs.h"
 #include "nb_enums.h"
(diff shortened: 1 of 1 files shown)


wojdyr:
These pragmas are also respected by Include Cleaner from clangd.

Recently, I've been evaluating Zed and got warnings about headers without installing or configuring anything. Apparently, the trend in code editors is to come with batteries included, and for C++ on Linux that's the clangd language server.

wjakob:
It looks kind of ugly :-( . Does `// IWYU pragma: begin_exports` also work?

adzenith:
Yeah, sounds great, lemme change it real quick.

adzenith:
Thanks!

PR #765, opened 2024-10-18
Adds an option LD_PRELOAD to cmake function to pre-load a shared library prior launching stub genreator.

adamryczkowski:
It can be used to pre-load address sanitizer shared object.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -398,7 +398,7 @@ function(nanobind_add_module name)
 endfunction()
 
 function (nanobind_add_stub name)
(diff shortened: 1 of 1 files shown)


wjakob:
Such a feature would need to be portable, but this AFAIK won't work on macOS or Windows. Also missing documentation.

adamryczkowski:
I understand your concern with documentation. Once you are OK with the change, I will write it.

Are you aware of any other way to use Nanobind with address sanitizer? If yes, then I kindly ask you to share.

I hope there is a way, because my organization has a policy of enforcing using of address sanitizers in unit tests, and that includes use of nanobind. 

What's wrong with providing non-protable solutions to problems that are platfrom-specific?

wjakob:
One option is to also compile Python with sanitizers, then the injection hack isn't needed. For some sanitizers (e.g. thread sanitizers in free-threaded builds), this is actually the only way of doing things that works.

Contributions to nanobind are held to a relatively high standard: if you want to add a feature, I kindly ask that you look into how to get it to work on all platforms. Partial solutions are difficult to explain and frustrating for users and maintainers alike.

I will close this PR for now.

PR #769, opened 2024-10-24
Perform two passes in the variant caster

tjstum:
When converting a Python object that wraps a `T` to a `std::variant<U,T>`, where `U` is implicitly convertible from `T`, the variant caster will cast to the `U` (even though the Python object is definitely a `T`)

The included test case demonstrates the issue. pybind11 does two pass conversion in this case. One can work around the issue by using `noconvert()` on the argument. But it seems that it would be a friendlier default to make type conversions in a variant work as expected.


wjakob:
LGTM, thanks! Can you rebase/resolve the merge conflict?

tjstum:
Just done, thanks!

wjakob:
Thanks!

PR #771, opened 2024-10-29
Add a non-const overload of bytearray::data

tjstum:
Unlike `bytes`, a `bytearray` is mutable. You are allowed to write to it, and the underlying API function returns a `char*`

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -1081,6 +1081,10 @@ Wrapper classes
 
       Return the size in bytes.
 
+   .. cpp:function:: void * data()
+
+      Convert a Python ``bytearray`` object into a byte buffer of length :cpp:func:`bytearray::size()` bytes.
+
    .. cpp:function:: const void * data() const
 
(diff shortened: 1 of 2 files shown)


tjstum:
This seemed simple enough to skip a test and changelog entry, but I'm happy to add those if that's desired. Thanks!

Hnasar (approved):
LGTM

wjakob:
In this case, I would expect to have a `const` and a non-`const` version of the cast operator, with only the latter having a non-`const` return value.

tjstum:
Thanks for the feedback! Just to be clear, since these are named methods, not operators, are you saying that you'd prefer if this adds `c_str_mutable()` and `data_mutable()` functions to the `bytearray` class?

oremanj:
I imagine the desire was for overloading on constness:
```
char* data() { return …; }
const char* data() const { return …; }
```
I would personally keep the `c_str` name const-only, but it would also be possible to do the same with it.

wjakob:
Thanks!

PR #772, opened 2024-10-31
Fix unhandled exception in set_caster

tjstum:
`obj_iter` throws `python_error` on failure, but type casters aren't allowed to throw! If you try to from-Python cast an object of a non-iterable type, Python will abort from the unhandled exception coming from a `noexcept` function

It looks like the intention here is to handle failure manually (a NULL return from the CPython call), so I did just that


wjakob:
Thanks!

Issue #773, opened 2024-11-01
[BUG]: Incorrect file extension generated for .pyd file when specifying Python version


MrDengDengDeng:
### Problem description

When specifying a specific Python version, the generated .pyd file has an incorrect suffix. The extension does not match the expected format for the targeted Python version, which leads to compatibility issues during module loading.
The .pyd file suffix issue was resolved by adjusting the regular expression in the if statement. I changed ".+-.+-.+" to ".+-.+_.+", and with this modification, the generated file name now has the correct suffix. Here’s the updated code:
`if(DEFINED NB_SOABI AND "${NB_SOABI}" MATCHES ".+-.+_.+")`
 `  set(NB_SUFFIX ".${NB_SOABI}${NB_SUFFIX_EXT}")`
`endif()`

### Reproducible example code

```text
cmake_minimum_required(VERSION 3.21)

project(Test VERSION 0.1.0 LANGUAGES C CXX)

set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_STANDARD 17)

set(Python_ROOT_DIR "${CMAKE_SOURCE_DIR}/venv312")

find_package(Python 3.12 COMPONENTS Interpreter Development REQUIRED)

find_package(Nanobind REQUIRED PATHS "${CMAKE_SOURCE_DIR}/venv312/Lib/site-packages/nanobind/cmake" NO_DEFAULT_PATH)

# if(DEFINED NB_SOABI AND "${NB_SOABI}" MATCHES ".+-.+_.+")
#   set(NB_SUFFIX ".${NB_SOABI}${NB_SUFFIX_EXT}")
# endif()

Message(STATUS "NB_SOABI: ${NB_SOABI}") # prints: cp312-win_amd64
Message(STATUS "NB_SUFFIX: ${NB_SUFFIX}") # prints:.cp311-win_amd64.pyd
```

wjakob:
Could I ask you to make a PR with this change to `cmake/nanobind-config.cmake`? Then we can see if this impacts all the various targets.

wjakob:
I don't know how to solve this issue on my own, and things seem stuck. It does not seem to affect many people (first time this came up). Will close for now, please reopen if you have additional information or a proposed fix.

mtsokol:
[...]

PR #776, opened 2024-11-04
Resolve issue with incorrect .pyd file extension generation

MrDengDengDeng:


wjakob:
Related to issue #773.

wjakob:
I am hesitant to merge this PR because there seem to be some platform-specific differences that aren't handled by this change.

On my macOS machine, the SOABI tag is `cpython-312-darwin`.  So with two dashes, rather than a dash and an underline.

MrDengDengDeng:
> I am hesitant to merge this PR because there seem to be some platform-specific differences that aren't handled by this change.
> 
> On my macOS machine, the SOABI tag is `cpython-312-darwin`. So with two dashes, rather than a dash and an underline.
If so, you should use different rules in cmake for different platforms

MrDengDengDeng:
If so, you should use different rules in cmake for different platforms

wjakob:
Let me add @henryiii, who wrote this line originally. I started to look into this and became quite confused.

On my macOS laptop, the suffix is `.cpython-312-darwin.so`. On Windows, it seems to be of the form `.cp311-win_amd64.pyd`. This page presents yet another naming scheme https://peps.python.org/pep-0425/ (but is outdated), and the current version seems in disagreement  as well: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#platform-compatibility-tags.

What are these version tags actually supposed to look like? What's the right combination of dashes and underscores to detect a compatible form and have the right mechanism for Scikit-build and PyPy?

wjakob:
I don't know how to solve this issue on my own, and things seem stuck. It does not seem to affect many people (first time this came up). Will close for now, please reopen if you have additional information or a proposed fix.

PR #779, opened 2024-11-06
Add <cstddef> to nanobind includes.

hawkinsp:
nb_types.h uses `std::ptrdiff_t`, which is defined in `cstddef`.

Fixes a build failure in Google's build environment, in which this header does not get included transitively by the existing headers.


wjakob:
Thanks!

PR #788, opened 2024-11-13
make_iterator: accept args by value, not forwarding reference, so that passing lvalue iterators works

oremanj:
This is a reasonable-looking way to expose `std::map::equal_range()` to Python:
```
    .def("equal_range", [](MapType& self, KeyType key) {
        auto range = self.equal_range(key);
        return nb::make_iterator(nb::type<MapType>(), "range_iterator", range.first, range.second);
    })
```
Unfortunately, it dangles: `range.first` has type `MapType::iterator&`, so deducing `Iterator&&` from it yields `Iterator = MapType::iterator&`, so a reference gets stored in the iterator `state` object. But that reference refers to the stack-allocated `range` object, which will be destroyed before the Python iterator is used. This PR changes `make_iterator` (and friends) arguments to accept their iterator and sentinel by value, to avoid this dangling risk. You can still get reference semantics by specifying the template arguments explicitly.

Also resolve ambiguity between the iterator and range overloads of `make_iterator`, and clean up the docs to reflect the iterator/sentinel split.


wjakob:
That seems dangerous indeed. Thank you very much!

PR #789, opened 2024-11-14
Fix handling of overloaded __new__ that takes no arguments

### LINKED ISSUE #786: [BUG]: Custom __new__ with no arguments
### Problem description

The docs provide an [example](https://nanobind.readthedocs.io/en/latest/classes.html#customizing-python-object-creation) on how to customize object creation, including an overload with no arguments. But without arguments, instantiation fails - see code below.

There seems to be some code and/or comments out of sync:
- if new_ is called on a function without arguments, then wrap_base_new creates a wrapper with 1 argument, while
- nb_type_vectorcall seems to assume that this wrapper has none.
- test46_custom_new even tests that this wrapper can be called with the class as argument.

Given the above, I do not see how using new_ on a function without arguments is meant to work.

### Reproducible example code

```cpp
class Pet
{
    Pet(){}
public:
    static std::unique_ptr<Pet> make(std::string name, int age) {
        return std::unique_ptr<Pet>{ new Pet() };
    };
};

nb::class_<Pet>(test, "Pet")
    .def(nb::new_([]() { return Pet::make("charly", 0); }))
    .def(nb::new_(&Pet::make), "name"_a, "age"_a);
```

```py
>>> from mod import Pet
>>> p = Pet("charly", 0)
>>> p = Pet()
TypeError: __new__(): incompatible function arguments. The following argument types are supported:
    1. __new__(cls: object, name: str, age: int) -> mod.Pet

Invoked with types:

```
```

### PR DESCRIPTION
**Pull request by oremanj**
Fixes #786. We now consider whether it's possible to invoke the overload with no arguments, rather than just whether it is declared with no parameters; the difference is relevant in the presence of `nb::args`/`nb::kwargs` and default values. The previous logic didn't work anyway - it checked whether there was an overload with zero args, but (a) this couldn't distinguish the auto-added wrapper overload from a user-specified one and (b) neither would have zero args, because new always takes at least the type arg.

### DIFF
--- a/src/nb_func.cpp
+++ b/src/nb_func.cpp
@@ -453,9 +453,8 @@ PyObject *nb_func_new(const void *in_) noexcept {
                     noargs_ok = false;
                     break;
                 }
-                if (noargs_ok) {
+                if (noargs_ok)
                     td->flags |= (uint32_t) type_flags::has_nullary_new;
-                }
             }
         } else if (is_new) {
             td->init = func;
--- a/src/nb_func.cpp
+++ b/src/nb_func.cpp
@@ -432,8 +432,31 @@ PyObject *nb_func_new(const void *in_) noexcept {
         type_data *td = nb_type_data((PyTypeObject *) f->scope);
         bool has_new = td->flags & (uint32_t) type_flags::has_new;
 
-        if (is_init && !has_new) {
-            td->init = func;
+        if (is_init) {
+            if (!has_new) {
+                td->init = func;
+            } else {
+                // Keep track of whether we have a __init__ overload that
+                // accepts no arguments (except self). If not, then we
+                // shouldn't allow calling the type object with no arguments,
+                // even though (for unpickling support) we probably do have
+                // a __new__ overload that accepts no arguments (except cls).
+                // This check is necessary because our type vectorcall shortcut
+                // skips Python's usual logic where __init__ is always called
+                // if __new__ returns an instance of the type.
+                bool noargs_ok = true;
+                for (size_t i = 1; i < fc->nargs - has_var_kwargs; ++i) {
+                    if (has_var_args && i == fc->nargs_pos)
+                        continue; // skip `nb::args` since it can be empty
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/nb_func.cpp [hunk 1]
```diff
@@ -432,8 +432,31 @@ PyObject *nb_func_new(const void *in_) noexcept {
...
+                // Keep track of whether we have a __init__ overload that
+                // accepts no arguments (except self). If not, then we
+                // shouldn't allow calling the type object with no arguments,
+                // even though (for unpickling support) we probably do have
+                // a __new__ overload that accepts no arguments (except cls).
+                // This check is necessary because our type vectorcall shortcut
+                // skips Python's usual logic where __init__ is always called
+                // if __new__ returns an instance of the type.
+                bool noargs_ok = true;
+                for (size_t i = 1; i < fc->nargs - has_var_kwargs; ++i) {
```
wjakob, line 448: Could the loop here also use `i < fc->nargs - has_var_kwargs - has_var_args` as stopping criterion?

src/nb_func.cpp
```diff
@@ -432,8 +432,31 @@ PyObject *nb_func_new(const void *in_) noexcept {
...
+                    if (has_var_args && i == fc->nargs_pos)
+                        continue; // skip `nb::args` since it can be empty
+                    if (has_args && fc->args[i].value != nullptr)
+                        continue; // arg with default is OK
+                    noargs_ok = false;
+                    break;
+                }
+                if (noargs_ok) {
```
wjakob, line 456: Can omit braces here

wjakob:
Thank you very much for the fix! I added two small comments.

src/nb_func.cpp
oremanj: No, because there can be keyword-only parameters in between `nb::args` and `nb::kwargs`, and the function is still invocable with no arguments as long as they all have defaults.

wjakob: Right, thank you!

PR #796, opened 2024-11-21
type_caster<ndarray>: accept None only if flags say so

WKarel:
Since https://github.com/wjakob/nanobind/commit/577942c8a13c931320db89c99d35778fc99be3d6 / https://github.com/wjakob/nanobind/pull/526, `type_caster<ndarray>::from_python` accepts None as argument, yielding an invalid ndarray. It does so independently of the cast-flags. This is fine, if the caster is called by the function dispatch mechanism, which handles cast_flags::accepts_none beforehand.
However, `type_caster<Eigen::Matrix/Array>` and `type_caster<Eigen::Map>` use `type_caster<ndarray>` as a local caster. Since these Eigen types are no wrappers, their type casters must not accept None. Part of this PR is that the local caster considers cast_flags::accepts_none, and the outer casters enforce the absence of this flag.
My use case is to avoid a read access violation when casting None returned by a virtual method overridden in Python, expected to return some Eigen::Matrix - see the according test.


wjakob:
Thank you, this looks great!

PR #805, opened 2024-12-02
Add an API isinstance(inst, cls) that mirrors the Python isinstance.

hawkinsp:
The current `isinstance<T>(inst)` tests if a Python object is an instance of a bound C++ class. This new overload tests if a Python object is an instance of a Python class. I have needed this helper in several nanobind-using codebases, and the same API exists in pybind11.

### DIFF
--- a/include/nanobind/nb_misc.h
+++ b/include/nanobind/nb_misc.h
@@ -111,6 +111,13 @@ inline Py_hash_t hash(handle h) {
     return rv;
 }
 
+inline bool isinstance(handle inst, handle cls) {
+    int ret = PyObject_IsInstance(inst.ptr(), cls.ptr());
+    if (ret == -1)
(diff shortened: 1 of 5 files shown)


include/nanobind/nb_misc.h
```diff
@@ -111,6 +111,13 @@ inline Py_hash_t hash(handle h) {
     return rv;
 }
 
+inline bool isinstance(handle inst, handle cls) {
+  int ret = PyObject_IsInstance(inst.ptr(), cls.ptr());
```
wjakob, line 115: indentation is different

wjakob:
This looks great. Could I ask you to add a changelog entry?

hawkinsp:
> This looks great. Could I ask you to add a changelog entry?

Done!

wjakob:
Thanks!

PR #806, opened 2024-12-03
Support overriding static properties defined via def_prop_ro_static.

hawkinsp:
See
https://github.com/pybind/pybind11/commit/e0e2ea33788e29ff5efca5842e92c6b179ca9378 for the analogous change to pybind11.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -49,6 +49,10 @@ Version TBD (unreleased)
   Previously, after ``a += b``, ``a`` would be replaced with a copy.
   (PR `#803 <https://github.com/wjakob/nanobind/pull/803>`__)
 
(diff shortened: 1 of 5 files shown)


wjakob:
This looks great, thank you! Could I ask you to also add a changelog entry?

hawkinsp:
BTW, I don't suppose you have a sense for when the next release will be?

I want to land a LLVM change that depends on this one (switching the upstream MLIR Python bindings to nanobind from pybind11... because nanobind is a lot faster: https://github.com/llvm/llvm-project/pull/118583).

wjakob:
That's awesome. Yes, I can push out a release soon.

wjakob:
@hawkinsp I pushed a release with the accumulated changes.

PR #829, opened 2024-12-20
Improve type_caster for floating-point types.

hpkfft:
I would like to propose that `nb::arg().noconvert()` not allow value-changing conversions of floating-point arguments.

Given
```
NB_MODULE(my_extension, m) {
    m.def("d",   [](double d) { printf("%.8f\n", d); });
    m.def("dnc", [](double d) { printf("%.8f\n", d); },
                 nb::arg().noconvert());

    m.def("f",   [](float f) { printf("%.8f\n", f); });
    m.def("fnc", [](float f) { printf("%.8f\n", f); },
                 nb::arg().noconvert());

#if 0
    m.def("ld",   [](long double x) { printf("%.8Lf\n", x); });
    m.def("ldnc", [](long double x) { printf("%.8Lf\n", x); },
                  nb::arg().noconvert());
#endif
}
```
we currently have the following:

|    x    |  "hello"  |     3      |    3.0     |  math.pi   |   1e40     |
| ------- | --------- | ---------- | ---------- | ---------- | ---------- |
| d(x)    | TypeError | 3.00000000 | 3.00000000 | 3.14159265 | 100....752 |
| dnc(x)  | TypeError | TypeError  | 3.00000000 | 3.14159265 | 100....752 |
| f(x)    | TypeError | 3.00000000 | 3.00000000 | 3.14159274 | inf        |
| fnc(x)  | TypeError | TypeError  | 3.00000000 | 3.14159274 | inf        |

and the two `long double` functions would not compile if they were uncommented.

I suggest the following is preferable:

|    x    |  "hello"  |     3      |    3.0     |  math.pi   |   1e40     |
| ------- | --------- | ---------- | ---------- | ---------- | ---------- |
| d(x)    | TypeError | 3.00000000 | 3.00000000 | 3.14159265 | 100....752 |
| dnc(x)  | TypeError | TypeError  | 3.00000000 | 3.14159265 | 100....752 |
| f(x)    | TypeError | 3.00000000 | 3.00000000 | 3.14159274 | inf        |
| fnc(x)  | TypeError | TypeError  | 3.00000000 | TypeError  | TypeError  |
| ld(x)   | TypeError | 3.00000000 | 3.00000000 | 3.14159265 | 100....752 |
| ldnc(x) | TypeError | TypeError  | 3.00000000 | 3.14159265 | 100....752 |

The code in this PR is not well tested, but I wanted to show some code to help spur feedback.
If the idea seems good, I'll work on this in January.


include/nanobind/nb_cast.h
```diff
@@ -126,10 +126,22 @@ template <typename T>
...
+                // Assume T, double, and Python float are all IEEE 754 binary64
+                return detail::load_f64(src.ptr(), flags, (double *) &value);
+            } else {
+                double d;
+                if (!detail::load_f64(src.ptr(), flags, &d))
+                    return false;
+                T result = static_cast<T>(d);
+                if ((flags & (uint8_t) cast_flags::convert)
+                        || static_cast<double>(result) == d
+                        || (result != result && d != d)) {
```
wjakob, line 139: What does `result != result && d != d` accomplish that `result != result` does not do?

include/nanobind/nb_cast.h
```diff
@@ -126,10 +126,22 @@ template <typename T>
...
-            else
-                return detail::load_f32(src.ptr(), flags, &value);
+            if constexpr (sizeof(T) == 8) {
+                // Assume T, double, and Python float are all IEEE 754 binary64
+                return detail::load_f64(src.ptr(), flags, (double *) &value);
+            } else {
+                double d;
+                if (!detail::load_f64(src.ptr(), flags, &d))
+                    return false;
+                T result = static_cast<T>(d);
```
wjakob, line 136: I think that this would be better to still keep in a dedicated `load_f32` routine with the double precision bits inlined. The goal is to keep binding code small that calls `load_f32` thousands of times.

wjakob:
I like this, the idea seems useful.

include/nanobind/nb_cast.h
hpkfft: I intend for the caster to work for any floating-point type.  The type `T` may not have `Inf`.  If that is the case, then a double precision `Inf` would be converted to `NaN`.   So, `result != result` but `d == d`.   This is a value-changing conversion, so we want it to fail if `noconvert()` was specified.  The same can happen if `d` is large.  Then the conversion would overflow (depending on rounding mode), and although `d` is finite, `result` is `NaN`.
This is a possible scenario.  Nvidia, Intel, Arm, Google, AMD, and Meta have "approved" an 8-bit floating-point specification E4M3 which does not have `Inf` but does have `NaN`.  (E5M2 has both `Inf` and `NaN`.)
https://www.opencompute.org/documents/ocp-8-bit-floating-point-specification-ofp8-revision-1-0-2023-12-01-pdf-1

Maybe we could just check `d != d`.   But that would be wrong if `T` does not support `NaN`.  I cannot immediately think of a system relevant to nanobind that has such a type, but I'd rather play it safe.   The `NaN` comparison check is at the end of all the short-circuiting, so I think it won't affect performance in practical usage.

include/nanobind/nb_cast.h
hpkfft: I restored `load_f32`.  Note that I statically assert that both `double` and `float` adhere to ISO/IEC 60559 [as documented here](https://eel.is/c++draft/numeric.limits#members-53.sentence-1), so I only check `d != d` since I know that if `d` is `NaN`, then the conversion to `float` will give `NaN`.  Hopefully, these assertions are true everywhere, or else I have some thinking to do....

In the case of `double`, the caster only checks `sizeof(T) == sizeof(double)`.  The assumption (as documented in the comment) is that this is ISO/IEC 60559 (i.e., IEEE 754) binary64.   Hopefully, this is always true for systems of interest.  The good news is this branch will be taken for `std::float64_t` as well as for `double`.   If you like, I'm happy to use `std::numeric_limits` in the test, but I hesitated to include `<limits>` since it's 1900 lines.

I used `std::is_same_v<T, float>` in the test for `float` since [TensorFloat-32](https://en.wikipedia.org/wiki/TensorFloat-32) is the same size as `float` but is a different representation.  So, `std::float32_t` will not take this branch.   (Of course, it will still be correct, but it will use the last branch.  (Without this PR, it doesn't work at all.))

I **did** include `<limits>` in `common.cpp` since it's only one file and it's already included transitively by `nb_internals.h`, which includes `tsl/robin_map.h`, which includes `tsl/robin_hash.h`, which includes `<limits>`.

tests/test_functions_ext.pyi.ref
```diff
@@ -120,8 +120,12 @@ def test_20(arg: str, /) -> object: ...
 
 def test_21(arg: int, /) -> int: ...
 
+def test_21_dnc(arg: float) -> float: ...
+
 def test_21_f(arg: float, /) -> int: ...
 
+def test_21_fnc(arg: float) -> float: ...
+
```
hpkfft, line 128: I noticed that `test_21_f` has a slash before the closing paren, but `test_21_fnc` does not.  I was not expecting `nb::arg().noconvert()` to have this effect since I did not give the argument a name.

wjakob: Yep, that's a known problem of the current code that generates the function signatures. It could likely be fixed but is not high on my priority list.

include/nanobind/nb_cast.h
```diff
@@ -126,10 +126,24 @@ template <typename T>
 struct type_caster<T, enable_if_t<std::is_arithmetic_v<T> && !is_std_char_v<T>>> {
     NB_INLINE bool from_python(handle src, uint8_t flags, cleanup_list *) noexcept {
         if constexpr (std::is_floating_point_v<T>) {
-            if constexpr (sizeof(T) == 8)
-                return detail::load_f64(src.ptr(), flags, &value);
-            else
+            if constexpr (sizeof(T) == sizeof(double)) {
```
wjakob, line 129: What about replacing the first line with `std::is_same_v<T, double>`?

src/common.cpp
```diff
@@ -926,23 +925,32 @@ bool load_f64(PyObject *o, uint8_t flags, double *out) noexcept {
 }
 
 bool load_f32(PyObject *o, uint8_t flags, float *out) noexcept {
+    static_assert(std::numeric_limits<float>::is_iec559);
```
wjakob, line 928: I think you can remove these. Nanobind is specifically scoped to architectures and compielrs that satisfy these assertions.

src/common.cpp
```diff
@@ -926,23 +925,32 @@ bool load_f64(PyObject *o, uint8_t flags, double *out) noexcept {
...
+    static_assert(std::numeric_limits<double>::is_iec559);
     bool is_float = PyFloat_CheckExact(o);
+    bool convert = flags & (uint8_t) cast_flags::convert;
 
 #if !defined(Py_LIMITED_API)
     if (NB_LIKELY(is_float)) {
-        *out = (float) PyFloat_AS_DOUBLE(o);
-        return true;
+        double d = PyFloat_AS_DOUBLE(o);
+        float result = static_cast<float>(d);
```
wjakob, line 936: Normal (non-`static_cast`) is fine for arithmetic conversions. I prefer to keep these fancy C++ casts to things that actually need them (generally the nanobind codebase is written that way).

include/nanobind/nb_cast.h
hpkfft: Yes, I like that.   It is future-proof if new 64-bit floating-point types are invented.

src/common.cpp
hpkfft: Done.

src/common.cpp
```diff
@@ -927,22 +925,29 @@ bool load_f64(PyObject *o, uint8_t flags, double *out) noexcept {
...
+        float result = (float) d;
+        if (convert || (double) result == d || d != d) {
+            *out = result;
+            return true;
+        } else {
+            return false;
+        }
     }
-
-    is_float = false;
```
wjakob, line 937: Can you re-enable this assigment? I am not sure that all compilers will understand that `is_float` can only be false following this conditional. Having the assignment gurantees that constant propagation will remove the check below.

src/common.cpp
```diff
@@ -904,18 +904,16 @@ bool load_f64(PyObject *o, uint8_t flags, double *out) noexcept {
 
 #if !defined(Py_LIMITED_API)
     if (NB_LIKELY(is_float)) {
-        *out = (double) PyFloat_AS_DOUBLE(o);
+        *out = PyFloat_AS_DOUBLE(o);
         return true;
     }
-
-    is_float = false;
```
wjakob, line 911: Can you re-enable this assigment? I am not sure that all compilers will understand that is_float can only be false following this conditional. Having the assignment gurantees that constant propagation will remove the check below.

hpkfft: Done. 

I had assumed this was an old work-around for a specific compiler issue and was no longer needed.
Clang does the right thing with nanobind's default -O3 optimization level.
Amusingly, with a debug build, clang performs the dead store and immediately reloads the value in the very next instruction to test whether it is false.  (No constant propagation, no dead store removal.)

Honestly, I think it's better not to have this since it only applies in a not NB_LIKELY code path.
But then I do not have any experience with non-Linux systems/compilers....
On Linux release builds, the dead store is removed, so having it is harmless.

Feel free to change your mind; I'm happy to revert this latest commit.   :)

wjakob: I prefer to haveit, I don't think it can do any harm in release mode, and debug mode performance is in any case meaningless.

PR #831, opened 2024-12-22
Simplify how to default CMAKE_BUILD_TYPE to Release.

hpkfft:
It is simpler to set the VALUE of the existing property to Release.

Note that this cannot be done so easily in nanobind's `CMakeLists.txt` because the `project()` specifies `LANGUAGES NONE`.
Probably, it could be done later in the file, after `enable_language(CXX)`.

See:  https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -44,8 +44,7 @@ addition, binding code may run slowly and produce large binaries.
 .. code-block:: cmake
 
    if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
-     set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
-     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY VALUE Release)
    endif()
 
 Finding nanobind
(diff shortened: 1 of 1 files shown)


wjakob:
I don't quite understand this change. What is nice with the current setting is that it gives a nice drop-down box in MSVC and cmake-gui.

hpkfft:
Yeah, I guess you're right.

After `project(my_project)`, the file `build/CMakeCache.txt` contains:
```
//Choose the type of build, options are: None Debug Release RelWithDebInfo
// MinSizeRel ...
CMAKE_BUILD_TYPE:STRING=
```
but `cmake-gui` ignores the comments.  (Though, the comments are useful if one edits the file with a text editor.)

After `set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)`, it becomes:
```
//Choose the type of build.
CMAKE_BUILD_TYPE:STRING=Release
```

Finally, `set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release"
    "MinSizeRel" "RelWithDebInfo")` adds the following towards the bottom of `build/CMakeCache.txt`:
```
//STRINGS property for variable: CMAKE_BUILD_TYPE
CMAKE_BUILD_TYPE-STRINGS:INTERNAL=Debug;Release;MinSizeRel;RelWithDebInfo
```
which is read by `cmake-gui` to give the nice drop-down menu.

The advice in this PR would change the default value to `Release` without doing anything else:
```
//Choose the type of build, options are: None Debug Release RelWithDebInfo
// MinSizeRel ...
CMAKE_BUILD_TYPE:STRING=Release
```

I suppose one could use two `set_property()` commands....
I'll make that change to this PR in case you like it better, but honestly, if I had realized that `cmake-gui` ignores the comments in `CMakeCache.txt` I would never have made this PR in the first place.   Feel free to close this as you like.

wjakob:
I will close it then. Thanks in any case, I am by no means a CMake expert and it is useful to see other ways of doing things.

PR #833, opened 2024-12-23
Fix -Wsign-compare gcc warning

tjstum:
gcc warns about this code:
`warning: comparison of integer expressions of different signedness: 'size_t' {aka 'long unsigned int'} and 'int' [-Wsign-compare]`

From what I can tell, this is a difference between clang and gcc, as explored [on godbolt](https://godbolt.org/z/T3TKsdvTr)

### DIFF
--- a/src/nb_func.cpp
+++ b/src/nb_func.cpp
@@ -445,7 +445,8 @@ PyObject *nb_func_new(const void *in_) noexcept {
                 // skips Python's usual logic where __init__ is always called
                 // if __new__ returns an instance of the type.
                 bool noargs_ok = true;
-                for (size_t i = 1; i < fc->nargs - has_var_kwargs; ++i) {
+                for (size_t i = 1;
+                     i < fc->nargs - static_cast<size_t>(has_var_kwargs); ++i) {
                     if (has_var_args && i == fc->nargs_pos)
                         continue; // skip `nb::args` since it can be empty
                     if (has_args && fc->args[i].value != nullptr)


tjstum:
I know that [the purification ritual comment](https://github.com/wjakob/nanobind/blob/b7c4f1abfcab9cc5a8f0ef758926d92ff5eac3a3/include/nanobind/nb_attr.h#L263-L288) mentions `-Wsign-conversion`, so I can understand if this is feeling too purification-y. But this does issue a warning with regular `-Wall` in modern gcc (even going back to gcc 8) as shown in that godbolt link.

Please also let me know if you have a better way you'd like this formatted. There are not a lot (any?) examples of multi-line for loops that I could find.

src/nb_func.cpp
```diff
@@ -445,7 +445,8 @@ PyObject *nb_func_new(const void *in_) noexcept {
                 // skips Python's usual logic where __init__ is always called
                 // if __new__ returns an instance of the type.
                 bool noargs_ok = true;
-                for (size_t i = 1; i < fc->nargs - has_var_kwargs; ++i) {
+                for (size_t i = 1;
+                     i < fc->nargs - static_cast<size_t>(has_var_kwargs); ++i) {
```
hpkfft, line 449: Since `int` can represent the entire range of values of the original types on the right, integer promotion converts each value to `int`.  Then subtraction results in an `int`.   So, may I suggest:
```
    for (int i = 1; i < fc->nargs - has_var_kwargs; ++i) {
```
I feel this is more natural/readable (and it fits on one line).  As a bonus, the generated code might be better.  On Compiler Explorer, I think it's nicer to uncheck the "Intel asm syntax" box.  Note that the `int` code is smaller using 32-bit integers than for 64-bit size_t, because `addq` takes one more byte than `addl`, etc.

  64bit: `48 83 c3 01  add $0x1,%rbx`    ==>  32bit: `83 c3 01  add $0x1,%ebx`
  64bit: `48 39 eb     cmp %rbp,%rbx`  ==>  32bit: `39 eb     cmp %ebp,%ebx`

wjakob:
I committed a fix directly. I am not a fan of `static_cast` for simple integer type conversions.

tjstum:
Sounds good. Thanks for the fix!

PR #835, opened 2024-12-24
Special forms for nb::typed

oremanj:
Add two special forms for `nb::typed`:

* `nb::typed<nb::object, T>` and `nb::typed<nb::handle, T>` render as just `T`, instead of the nonsensical `object[T]` that they would be otherwise. This allows recapture of the pre-2.0 functionality of `nb::typed` where it entirely replaces (rather than augments) the type of an argument.
* `nb::typed<nb::callable, Sig>` for some C++ function signature `Sig` produces an equivalent Python callable signature. For example, `nb::typed<nb::callable, void(float, bool)>` would become `Callable[[int, float], None]`, which there is currently no way to express due to the nested brackets.


wjakob:
That's pretty cool. Thanks!

PR #836, opened 2024-12-25
Enable subclassing `nb_type`

makslevental:
I would like to be able to subclass `nb_type`. Currently this is possible in `pybind` and I use this feature to perform some "magic" when a pybind class pops out from C++ (see [here](https://github.com/makslevental/mlir-python-extras/blob/af8bd5fd628d9ed61a516acc7133019ac8fa92f5/mlir_utils/dialects/ext/arith.py#L75)). The upstream project (MLIR) recently switched `pybind -> nanobind` so I'd like to restore this functionality.

Let me know if you'd like me to add a test case.

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -876,7 +876,7 @@ static PyTypeObject *nb_type_tp(size_t supplement) noexcept {
             /* .name = */ name,
             /* .basicsize = */ basicsize,
             /* .itemsize = */ itemsize,
-            /* .flags = */ Py_TPFLAGS_DEFAULT,
+            /* .flags = */ Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
             /* .slots = */ slots
         };
 


wjakob:
Doing anything with the `nb_type` metaclasses is dangerous and intentionally disallowed. Only `libnanobind` can safely create new nanobind types. First of all, new instances need to be correctly registered in the `internals` data structure. Secondly, nanobind types also use more memory than standard Python types, which the library internally uses for bookkeeping. This requires a special allocation sequence and correct initialization of this supplemental data. All of this is orchestrated by the metaclasses, which are part of the secret sauce to make nanobind fast, and that also means that they may change in unpredictable ways in the future. This is why they are locked away.

Things that required certain workarounds in pybind11 should often be done differently (potentially more directly) in nanobind, so a direct 1:1 port of "magic" is generally not recommended. I suggest that you create a discussion post that explains your use case (ideally reduced a bit, the MLIR implementation file you posted was difficult to follow)

PR #840, opened 2025-01-03
Add `nanobind.source_dir()` API to obtain the source path

nicholasjng:
This is useful when building with other tools or build systems, like `setuptools` in my case.

-------------

Right now, to obtain the source path, I am using `Path(nanobind.include_dir()).with_name("src")`, but I felt that this was a worthy addition for anyone needing the sources path when building nanobind from source themselves as part of a package / wheel build by means other than scikit-build-core.


wjakob:
Thanks

PR #842, opened 2025-01-05
Update nanobind-bazel docs with latest developments

nicholasjng:
`nanobind_bazel` v2.4.0 was released on the BCR on christmas. The MODULE.bazel code example is thus updated, as is tradition, along with two more user-facing changes:

1. The minimum Bazel version was increased to Bazel 7, since we need to use a sufficiently new `rules_python` for free-threading support.
2. The `@nanobind_bazel//:free_threading` flag was removed, since it was not that useful - you still needed to give `rules_python`'s free-threading flag separately to request a free-threaded toolchain, and supplying it did not even guarantee a free-threading build. Now, we just use `rules_python`'s free-threading flag value directly instead.

### DIFF
--- a/docs/api_bazel.rst
+++ b/docs/api_bazel.rst
@@ -155,15 +155,3 @@ following flag settings.
     version. Allowed values are ``"cp312"``, ``"cp313"``, which target the
     stable ABI starting from Python 3.12 or 3.13, respectively. By default, all
     extensions are built without any ABI limitations.
-
-.. py:function:: @nanobind_bazel//:free_threading (boolean)
-
-    Build nanobind extensions with a Python toolchain in free-threaded mode.
-    If given, the currently configured Python toolchain must support free-threading,
-    otherwise, the build will result in a compilation error.
-    Only relevant for CPython 3.13+, since support for free-threaded Python was
-    introduced in CPython 3.13.
-    For more information on free-threaded extension support in nanobind, refer to the
-    relevant :ref:`documentation section <free-threaded>`.
-
-    *New in nanobind-bazel version 2.2.0.*
--- a/docs/bazel.rst
+++ b/docs/bazel.rst
@@ -139,6 +139,30 @@ Naturally, since stub generation relies on the given shared object files, the
 actual extensions are built in the process before invocation of the stub
 generation script.
 
+Building extensions for free-threaded Python
+--------------------------------------------
+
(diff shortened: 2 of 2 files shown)


wjakob:
Actually, onecomment: it may be useful to keep the free-threading attribute to as an opt-in (similarly to how this is implemented in the CMake build system). Upgrading a package to correctly work in a free-threaded environment can be quite complex. One checkpoint along the way is just to run it as-is (not opted in), in which case the free-threaded features of the interpreter are turned off. Not very useful if your goal is parallelism, but helpful for bring-up of large multi-extension projects.

nicholasjng:
I agree - in fact, `rules_python`'s free-threading flag means that a free-threaded toolchain will be used instead of a "classic" one whenever possible (i.e. CPython 3.13+), so passing `@rules_python//config_settings:is_free_threaded=no` (name slightly different) just means regular CPython 3.13 will be used, and `-DNB_FREE_THREADED` will not be given.

Also, no free-threading is the default, so we can build regular CPython 3.13 just fine (in fact, that's what I'm still doing in nanobind_bazel CI).

Does that match your expectation?

wjakob:
That makes sense! Is this documented somewhere? Otherwise, it may not be easy to infer for users.

nicholasjng:
Afraid not. I can add a paragraph on free-threaded builds to the Bazel doc, either in this PR or a follow-up.

wjakob:
Let's put it here, we can keep this PR open a little longer.

nicholasjng:
Done.

On a side note, the docs venv setup is broken at least on my MacOS machine: When pip-installing docs/requirements.txt with Python<3.13, I get an error from sphinx about missing `pkg_resources`:

```
$ sphinx-build docs docbuild -a
Running Sphinx v6.1.3

Extension error:
Could not import extension sphinxcontrib.moderncmakedomain (exception: No module named 'pkg_resources')
```

That one can be fixed by running `pip install --upgrade setuptools`. On Python 3.13, it's another module that errors out:

```
$ sphinx-build docs docbuild -a
Running Sphinx v6.1.3

Extension error:
Could not import extension sphinx.builders.epub3 (exception: No module named 'imghdr')
```

Just wanted to put this out there, since there are other people working on the documentation as well.

wjakob:
Thank you! I'll make a note to transition to a newer version of Sphinx in the future.

PR #856, opened 2025-01-14
Fix [-Wcast-qual] warnings on GCC 14.2

smurfix:
GCC's C++ with -Wcast-qual enabled wants C++17's `const_cast`.

### DIFF
--- a/include/nanobind/nb_types.h
+++ b/include/nanobind/nb_types.h
@@ -179,8 +179,8 @@ public:
     handle &operator=(handle &&) noexcept = default;
     NB_INLINE handle(std::nullptr_t, detail::steal_t) : m_ptr(nullptr) { }
     NB_INLINE handle(std::nullptr_t) : m_ptr(nullptr) { }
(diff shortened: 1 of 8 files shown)


smurfix:
I hit a few more, sorry about the noise.

Unsure how to handle this one.
```
/usr/share/nanobind/include/nanobind/nb_cast.h:469:19: warning: cast from type ‘const CommandCost*’ to type ‘nanobind::detail::type_caster_base<CommandCost>::Type*’ {aka ‘CommandCost*’} casts away qualifiers [-Wcast-qual]
  469 |             ptr = (Type *) &value;
```

wjakob:
There is a comment specifically about this (https://github.com/wjakob/nanobind/blob/cee104db8606797a63752d2904b2f2795014a125/include/nanobind/nb_attr.h#L263). I should probably make it into an FAQ entry.

wjakob:
(FAQ entry added)

smurfix:
OK … your choice, of course.

My problem is integration into an existing project whose maintainers will not accept the (implied) regression.

PR #858, opened 2025-01-14
Added nanobind_setup_libname helper function

vfdev-5:
Context: https://github.com/wjakob/nanobind/discussions/844#discussioncomment-11828312

cc @wjakob

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -308,6 +308,35 @@ function (nanobind_musl_static_libcpp name)
   endif()
 endfunction()
 
+function (
+  nanobind_setup_libname
+  output_libname
+  ARG_NB_STATIC
+  ARG_STABLE_ABI
+  ARG_FREE_THREADED
+  ARG_NB_DOMAIN
+  ARG_NB_SHARED
+)
+  set(libname "nanobind")
+  if (ARG_NB_STATIC)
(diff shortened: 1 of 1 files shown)


wjakob:
How about giving this function the same signature as `nanobind_add_module`? Then the arguments can easily be forwarded, which I think is convenient. Let's call it `nanobind_libname`

vfdev-5:
> How about giving this function the same signature as `nanobind_add_module`? Then the arguments can easily be forwarded, which I think is convenient. Let's call it `nanobind_libname`

`nanobind_add_module` signature is just `name` and all other arguments are parsed with `cmake_parse_arguments`.
I'm not totally sure how can we forward the arguments (`"STABLE_ABI;FREE_THREADED;NB_STATIC;NB_SHARED;PROTECT_STACK;LTO;NOMINSIZE;NOSTRIP;MUSL_DYNAMIC_LIBCPP"`) which could be parsed first by `nanobind_add_module` and then also by `nanobind_libname`. 

Also parsed arguments like ARG_STABLE_ABI, ARG_NB_STATIC etc are then set to some default value depending on their mutual values:

https://github.com/wjakob/nanobind/blob/8badd40da0399e24e2482dc6b5307745022c59a8/cmake/nanobind-config.cmake#L322-L339

If you have few more details on the idea how to move forward I'm all ears.

PR #860, opened 2025-01-15
Remove redundant move in test

hpkfft:
An out-of-the box build with Debian trixie (i.e., with g++ 14.2.0) gives the following warning:
```
warning: redundant move in return statement [-Wredundant-move]
  508 |     m.def("test_type_object_t", [](nb::type_object_t<Struct> h) -> nb::object { return std::move(h); });
      |                                                                                        ~~~~~~~~~^~~
```
Since the return expression is move-eligible, the move constructor is selected without the `std::move`.


wjakob:
So AFAIK I added this `std::move()` because of a compiler warning saying that I needed an explicit move to avoid using a converting constructor. But I can't repro it anymore. Will merge this for now.

Issue #864, opened 2025-01-16
[BUG]: Race condition between wrapper deallocation and lookup


colesbury:
### Problem description

There's a race condition between wrapper lookup and wrapper deallocation where a Python wrapper may be returned that's in the process of being deallocated. I have a reproducer for the free threading build. I think the problem can also affect the default (GIL-enabled) build as well, but I don't have a reproducer yet.

This the counterpart of the pybind11 bug:

* https://github.com/pybind/pybind11/issues/5489

### Explanation

[`nb_type_put`](https://github.com/wjakob/nanobind/blob/8badd40da0399e24e2482dc6b5307745022c59a8/src/nb_type.cpp#L1769) will lookup an existing Python wrapper object in `inst_c2p`. The [`inst_dealloc`](https://github.com/wjakob/nanobind/blob/8badd40da0399e24e2482dc6b5307745022c59a8/src/nb_type.cpp#L207) function removes the wrapper from `inst2_cp` when the wrapper is deallocated.

During the `nb_type_put` call, it's possible that the found wrapper has a reference count of 0 and is in the process of being deallocated, but not yet removed from `inst2_cp`. In the free threading build, this can happen because `nb_type_put` can be run concurrently with  `inst_dealloc` *up to* the acquisition of the [shard lock](https://github.com/wjakob/nanobind/blob/8badd40da0399e24e2482dc6b5307745022c59a8/src/nb_type.cpp#L255-L256). I think this can also happen in the default (GIL-enabled) build, because things like [`Py_CLEAR(*dict)`](https://github.com/wjakob/nanobind/blob/8badd40da0399e24e2482dc6b5307745022c59a8/src/nb_type.cpp#L218) can call arbitrary code that may temporarily release the GIL.

### Suggested fix

`nb_type_put` should only incref and return a wrapper if the reference count is not zero. In the GIL-enabled build, this is roughly:

```c++
if (Py_REFCNT(seq.inst) > 0) {
    Py_INCREF(seq.inst);
    return seq.inst;
}
```

In the free threading build, we'll want to use [`PyUnstable_TryIncref`](https://github.com/python/cpython/issues/128844) when it's available, or implement that logical like we're doing in pybind11.

### See also

* https://github.com/python/cpython/issues/128844
* https://github.com/pybind/pybind11/pull/5494

### Reproducible example code

```c++
struct Data {};
Data MyData;

NB_MODULE(my_ext, m) {
     nb::class_<Data>(m, "Data")
[...]

PR #865, opened 2025-01-17
Fix race condition involving wrapper lookup

colesbury:
See #864

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -44,22 +44,10 @@ static void nb_enable_try_inc_ref(PyObject *obj) noexcept {
 #if 0 && defined(Py_GIL_DISABLED) && PY_VERSION_HEX >= 0x030E00A5
     PyUnstable_EnableTryIncRef(obj);
 #elif defined(Py_GIL_DISABLED)
-    // TODO: Replace with PyUnstable_Object_EnableTryIncRef when available.
-    // See https://github.com/python/cpython/issues/128844
-    if (_Py_IsImmortal(obj)) {
-        return;
-    }
-    for (;;) {
-        Py_ssize_t shared = _Py_atomic_load_ssize_relaxed(&obj->ob_ref_shared);
-        if ((shared & _Py_REF_SHARED_FLAG_MASK) != 0) {
-            // Nothing to do if it's in WEAKREFS, QUEUED, or MERGED states.
-            return;
-        }
-        if (_Py_atomic_compare_exchange_ssize(
-                &obj->ob_ref_shared, &shared, shared | _Py_REF_MAYBE_WEAKREF)) {
-            return;
-        }
-    }
+    // Since this is called during object construction, we know that we have
+    // the only reference to the object and can use a non-atomic write.
+    assert(obj->ob_ref_shared == 0);
+    obj->ob_ref_shared = _Py_REF_MAYBE_WEAKREF;
 #endif
 }
 
@@ -164,11 +152,11 @@ PyObject *inst_new_int(PyTypeObject *tp, PyObject * /* args */,
         self->clear_keep_alive = 0;
         self->intrusive = intrusive;
         self->unused = 0;
+        nb_enable_try_inc_ref((PyObject *)self);
 
         // Update hash table that maps from C++ to Python instance
         nb_shard &shard = internals->shard((void *) payload);
         lock_shard guard(shard);
-        nb_enable_try_inc_ref((PyObject *)self);
         auto [it, success] = shard.inst_c2p.try_emplace((void *) payload, self);
         check(success, "nanobind::detail::inst_new_int(): unexpected collision!");
     }
@@ -230,12 +218,12 @@ PyObject *inst_new_ext(PyTypeObject *tp, void *value) {
     self->clear_keep_alive = 0;
     self->intrusive = intrusive;
     self->unused = 0;
+    nb_enable_try_inc_ref((PyObject *)self);
 
     nb_shard &shard = internals->shard(value);
     lock_shard guard(shard);
 
     // Update hash table that maps from C++ to Python instance
-    nb_enable_try_inc_ref((PyObject *)self);
     auto [it, success] = shard.inst_c2p.try_emplace(value, self);
 
     if (NB_UNLIKELY(!success)) {
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -41,7 +41,7 @@ static PyObject **nb_weaklist_ptr(PyObject *self) {
 }
 
 static void nb_enable_try_inc_ref(PyObject *obj) noexcept {
-#if 0 && defined(Py_GIL_DISABLED) && Py_VERSION_HEX >= 0x031400a4
+#if 0 && defined(Py_GIL_DISABLED) && PY_VERSION_HEX >= 0x030E00A5
     PyUnstable_EnableTryIncRef(obj);
 #elif defined(Py_GIL_DISABLED)
     // TODO: Replace with PyUnstable_Object_EnableTryIncRef when available.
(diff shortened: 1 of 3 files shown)


src/nb_type.cpp
```diff
@@ -40,6 +40,72 @@ static PyObject **nb_weaklist_ptr(PyObject *self) {
...
+        if (_Py_atomic_compare_exchange_ssize(
+                &obj->ob_ref_shared, &shared, shared | _Py_REF_MAYBE_WEAKREF)) {
+            return;
+        }
+    }
+#endif
+}
+
+static bool nb_try_inc_ref(PyObject *obj) noexcept {
+#if 0 && defined(Py_GIL_DISABLED) && PY_VERSION_HEX >= 0x030E00A5
```
wjakob, line 67: Is the `0` intentional here?

src/nb_type.cpp
```diff
@@ -1766,16 +1834,18 @@ PyObject *nb_type_put(const std::type_info *cpp_type,
                 PyTypeObject *tp = Py_TYPE(seq.inst);
 
                 if (nb_type_data(tp)->type == cpp_type) {
-                    Py_INCREF(seq.inst);
-                    return seq.inst;
+                    if (nb_try_inc_ref(seq.inst)) {
```
wjakob, line 1837: I think that if `nb_try_inc_ref` fails, what we actually want to do is to `break` out of the loop that considers the different Python objects associated with that pointer address. We tried the only possible match, and it cannot be returned.

Ditto below.

There are also two other places, where this `nb_try_inc_ref` will be needed, which is in `nb_type_put_p` below (mirrors the function here, but for polymorphic C++ objects)

src/nb_type.cpp
```diff
@@ -40,6 +40,72 @@ static PyObject **nb_weaklist_ptr(PyObject *self) {
     return weaklistoffset ? (PyObject **) ((uint8_t *) self + weaklistoffset) : nullptr;
 }
 
+static void nb_enable_try_inc_ref(PyObject *obj) noexcept {
+#if 0 && defined(Py_GIL_DISABLED) && PY_VERSION_HEX >= 0x030E00A5
+    PyUnstable_EnableTryIncRef(obj);
+#elif defined(Py_GIL_DISABLED)
+    // TODO: Replace with PyUnstable_Object_EnableTryIncRef when available.
```
wjakob, line 47: What's the difference between `PyUnstable_Object_EnableTryIncRef` and `PyUnstable_EnableTryIncRef`?

src/nb_type.cpp
```diff
@@ -40,6 +40,72 @@ static PyObject **nb_weaklist_ptr(PyObject *self) {
     return weaklistoffset ? (PyObject **) ((uint8_t *) self + weaklistoffset) : nullptr;
 }
 
+static void nb_enable_try_inc_ref(PyObject *obj) noexcept {
+#if 0 && defined(Py_GIL_DISABLED) && PY_VERSION_HEX >= 0x030E00A5
```
wjakob, line 44: Is the 0 intentional here?

wjakob:
Hi Sam, thank you for tracking this down and making a patch. I left some comments inside. This added requirement to invoke `nb_enable_try_inc_ref` on each new instance sounds expensive to me. Wouldn't it better to provide an API to directly allocate try-inc-ref-enabled objects via `PyObject_New()` and `PyType_GenericAlloc()`?

src/nb_type.cpp
colesbury: The APIs haven't been added to CPython yet, but I wanted to get this PR ready while the issue was still fresh in my mind. We can either:

1) Delete this code path for now
2) Wait until the APIs are added
3) Or anything else you prefer

src/nb_type.cpp
colesbury: No difference, just the name changed. I'll update the comment.

src/nb_type.cpp
colesbury: I think breaking out here would create more wrappers than necessary. The case I'm thinking about is where there are multiple lookups before the to-be-deallocated wrapper is removed from `inst_c2p`:

T1: starts to deallocate wrapper object, but hasn't removed it from `inst_c2p` yet.
T2: lookups wrapper, sees zero refcount. The entry is converted to a linked list and the new wrapper is inserted at the end.
...
T3: lookups wrapper. The first entry still the zero refcount wrapper. The second entry is the wrapper we want. If we break out early, we'll create a third wrapper, which doesn't seem ideal.

colesbury:
I'll measure the overhead `nb_enable_try_inc_ref`. Adding a whole new allocation API like `PyObject_New()` to allocate try-inc-ref-enabled objects seems like it would be a tough sell.

hawkinsp:
I confirmed that this PR fixes a race that @vfdev-5 and I were trying to track down in the JAX test suite when run under free-threading (reduced example: https://gist.github.com/vfdev-5/d46bcdc73231bd1e2b2f85c40de9f890)

colesbury:
I benchmarked some simple wrapping code using the current `main` branch of CPython (with free threading), roughly:

```c++
struct Foo {};
nb::class_<Foo>(m, "Foo");
m.def("new_foo", []() -> Foo * { return new Foo(); });
```

And then calling `t.new_foo()` in a loop to measure the combined time for wrapper creation and deallcation.

On my M1 mac, the difference is basically in the noise margin: ~80 ns. The extra overhead seems close maybe 1-2 ns.
On my Intel Linux machine: I'm seeing ~129 ns with the extra cost of ~6 ns.

We can optimize `nb_enable_try_inc_ref` (and `PyUnstable_EnableTryIncRef`) a bit more for the case where it's called immediately after object construction. That brings down the overhead on Intel so the iteration cost is ~129.5 ns (well within the noise margin; not sure 0.5 ns is meaningful here) with the inlined implementation of ~131 ns with the out-of-line call to `PyUnstable_EnableTryIncRef`.

src/nb_type.cpp
wjakob: In nanobind (and pybind11, for that matter), a single pointer can be associated with multiple Python instances.

For example, you could have
```
struct A {
    B b;
    /* ... */
}
```
with both an `A*` pointer and a `B*` pointer being simultaneously alive through Python wrappers. These can be created and GCed independently from each other.

The `nb_type_put()` function converts a C++ pointer into an associated Python object (if one exists) or it creates a new one. It walks through a chain of objects that are associated with a given pointer address until it finds one of the right type. There can only be one of the right type.

Your commit adds a failure mechanism, where it turns out that one wrapper is in the process of being deleted, and we need to create a new one. In this case, it (AFAIK) doesn't make sense to try the other types associated with the current pointer, since we already found the one. Hence the suggestion to `break` and immediately create a new wrapper. Arguably this is pretty minor and unlikely to have a big performance impact.

src/nb_type.cpp
wjakob: I think it's OK as is, we can remove the `#if 0` once the change goes in.

Do I understand correctly that the version below now has an optimization (specific to object construction) that won't be in the official API?

wjakob:
Thank you @colesbury, I left 2 more comments.

By the way, in `src/common.cpp:1182`, there is the `maybe_make_immortal` function that contains FT implementation details.
Is there an official API for this step that could be used?

src/nb_type.cpp
wjakob: Hum, thinking more about this, I think that is currently necessary continue iterating. There could be multiple objects that are in the process of being cleaned up.

wjakob:
I will merge this for now, we can keep on iterating on the details (there seems to be a related race condition as well). Aside, it would be nice to reference official API for all free threading-related API implementation details such as immortalizing objects (on sufficiently new Python builds where such an API is available).

src/nb_type.cpp
colesbury: I intend to add a similar optimization in the implementation of the CPython API.

colesbury:
I'll take a look at the immortalization stuff again. Immortalization was, to some extent, a workaround for limitations in 3.13. I'm hoping we can avoid it in 3.14.

src/nb_type.cpp
oremanj: You can also get multiple nanobind instances wrapping the same pointer, with the same type, by using `nb::inst_wrap` from the low-level API.

PR #878, opened 2025-01-23
Two small doc fixes

smurfix:
* Bytes are not unicode
* the word "section" was missing from a sentence in the free-threading section
* replaced "the project" with "nanobind" in the same sentence,
   'cause that's somewhat more clear prose to this reader


wjakob:
Thanks!

PR #883, opened 2025-01-27
Don't print too many "leaked instance" messages.

smurfix:
Most terminals don't have 10k line s of scrollback buffer ...

### DIFF
--- a/src/nb_internals.cpp
+++ b/src/nb_internals.cpp
@@ -295,6 +295,9 @@ static void internals_cleanup() {
                     break;
             }
         }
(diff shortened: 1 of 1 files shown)


wjakob:
Fair enough. Needs a "skipped remainder" part just like below.

And actually, let's make a change here: I would like the skipping behavior to be disabled when compiled in debug mode. It can be super-useful to see what kinds of instances are leaked to identify the potential source. Showing only 20 of them might hide the culprit.

wjakob:
Thanks!

PR #884, opened 2025-01-27
Introduce def_visitor abstraction for objects that provide custom binding logic when passed to def()

oremanj:
This has two purposes:
* I'm planning to contribute some `class_` extensions (for pickling and buffer protocol support) that I would like to be able to put in their own headers so they don't slow down compilation for those that don't use them. This PR allows `class_::def()` to perform an operation that was not foreseen by the code inside `class_`.
* I've found it quite ergonomically useful (and am carrying a local patch) to be able to add customizations by chaining `.def(my_thing())` to a class binding, rather than the existing supported alternative of giving the binding a name and calling a helper function on it (`auto someCls = nb::class_<...>(...); my_thing(someCls);`).

Previously `class_filler`, renamed to `def_visitor` per feedback.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -2217,6 +2217,13 @@ Class binding
       where possible. See the discussion of :ref:`customizing object creation
       <custom_new>` for more details.
 
+   .. cpp:function:: template <typename Visitor, typename... Extra> class_ &def(def_visitor<Visitor> arg, const Extra &... extra)
+
+      Dispatch to custom user-provided binding logic implemented by the type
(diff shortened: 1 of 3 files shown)


wjakob:
This is great! I like `def_visitor` as name.

oremanj:
Great, renamed!

wjakob:
Thanks!

ManifoldFR:
Wow, thanks for adding this!!
This should make the transition for some of us who still use Boost.Python (and hadn't moved to Pybind11) much easier, since BP also had this `def_visitor` method-chaining pattern.

Issue #867, opened 2025-01-20
[BUG]: Race on `inst->state` between nb_type_get / nb_type_put_common under free-threading


vfdev-5:
### Problem description

We see few data races when accessing class attribute of another class attribute:
```python
pshard.sharding_spec.sharding
```
- one race is similar to Sam (@colesbury) reported here: https://github.com/wjakob/nanobind/issues/864
- second one is between nb_type_get / nb_type_put_common which try to read and write into `inst->state`.

The first race we saw being fixed by https://github.com/wjakob/nanobind/pull/865 .

@wjakob how would you like this second race to be fixed?

Full TSAN report and code reproducer is here: https://gist.github.com/vfdev-5/d46bcdc73231bd1e2b2f85c40de9f890

Another similar TSAN report by @hawkinsp : https://gist.github.com/hawkinsp/5f3997c72c2c6781c1d8a90225f3eddd

### Reproducible example code

```text
Python code: https://gist.github.com/vfdev-5/d46bcdc73231bd1e2b2f85c40de9f890#file-check-py
C++ bindings: https://gist.github.com/vfdev-5/d46bcdc73231bd1e2b2f85c40de9f890#file-binding-cpp
CMakeLists: https://gist.github.com/vfdev-5/d46bcdc73231bd1e2b2f85c40de9f890#file-cmakelists-txt
```

wjakob:
I suspect the following: `inst_new_ext` (which creates a Python object wrapping an existing C++ instance) right now acquires a lock to the shard containing this object. However, there are some uses of this API which continue to modify the associated Python object following its creation. In the meantime, another thread who is accessing that same pointer can look up the partially created object from the hash table.

What I think needs to change is that locking the shard becomes the responsibility of the caller of `inst_new_ext` rather than having the function do this. The `inst_new_int` is unaffected since here we don't know ahead of time what the associated pointer address is.

This race condition aside is also another fundamental race that will need to be figured out: if two threads simultaneously access the same C++ pointer, which currently does not exist on the Python side, there are two outcomes.

Case 1:
Thread 1: No Python object found. Starts to creates Python object
Thread 2: Finds Python object created by Thread 1 & returns that

Case 2:
Thread 1: No Python object found. Starts to creates Python object
Thread 2: No Python object found. Starts to create Python object

That second case sounds like it could be a problem and would be nice to avoid.

wjakob:
Is it easy for you to repro these? If I send you a patch, can you tell me if that fixed it?

[...]

PR #887, opened 2025-01-28
Fix race condition in free-threaded Python (fixes issue #867)

wjakob:
This commit addresses an issue arising when multiple threads want to access the Python object associated with the same C++ instance, which does not exist yet and therefore must be created. @vfdev-5 reported that TSAN detects a race condition in code that uses this pattern, caused by concurrent unprotected reads/writes of internal ``nb_inst`` fields.

To fix this issue, we split instance creation and registration into a two-step process. The registration is only done when the object is fully constructed.

### DIFF
--- a/docs/free_threaded.rst
+++ b/docs/free_threaded.rst
@@ -103,6 +103,11 @@ supplemental locking. The :ref:`next section <free-threaded-locks>` explains a
 Python-specific locking primitive that can be used in binding code besides
 the solutions mentioned above.
 
+Multi-threaded code that concurrently returns the same C++ instance via the
+:cpp:enumerator:`nb::rv_policy::reference` policy may observe situations, where
+multiple Python objects are created that all wrap the same C++ instance
+(however, this is harmless aside from the duplication).
+
 .. _free-threaded-locks:
 
 Python locks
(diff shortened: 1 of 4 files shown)


wjakob:
This commit fixes the race condition in free-threaded Python. 

cc @vfdev-5 @hawkinsp @oremanj 

I punted on the idea of strictly enforcing an invariant that a (C++ instance pointer, type) pair can map to at most one Python object. See branch `ft-race-alt` / commit 1415e0522570cbf691eac2826a82072256ab0eda for a failed (?) attempt of such an implementation. In that alternative implementation, I check the instance list for duplicates while registering a new instance. If a duplicate exists, I return that one instead and destroy the unregistered instance. This passes the tests but makes me worry that it's going to be difficult to properly enforce the return value policies.

For example, what if some threads return an object via `rv_policy::reference`, and then one single thread takes ownership via `rv_policy::take_ownership`. If we potentially return some other thread's instance, we would have to "patch in" the update the flags, which sounds complicated. (This is kind of a silly example because it is not handled correctly at the moment even without this patch. But it seems fixable if the creation of duplicate instances is allowed)

wjakob:
In `inst_new_int`, that whole issue cannot appear because we're registering a new instance that nobody has seen yet. So there cannot be a race on the pointer identity of this instance.

makslevental:
@wjakob "peanut gallery"-level question: can we get a `PATCH` version bump for this change so that we can require it downstream (in [MLIR](https://github.com/llvm/llvm-project/blob/main/mlir/python/requirements.txt#L1)).

Issue #890, opened 2025-01-29
[BUG]: Unwarranted "noexcept" declaration


smurfix:
### Problem description

Consider

``` C++
/// Simplified nb_func_vectorcall variant for functions w/o keyword arguments
static PyObject *nb_func_vectorcall_simple(PyObject *self,
                                           PyObject *const *args_in,
                                           size_t nargsf,
                                           PyObject *kwargs_in) noexcept {
```
That `noexcept` is a mistake, I just encountered a crash because of it. GDB finds that you call, in line 914,
``` C++
                // Found a suitable overload, let's try calling it
                result = f->impl(const_cast<void **>(f->capture), const_cast<PyObject **+ >(args_in),
                                 args_flags, (rv_policy) (f->flags & 0b111),
                                 &cleanup);
```
which goes to `nb_func.h` line 219:
``` C++
    f.impl = [](void *p, PyObject **args, uint8_t *args_flags, rv_policy policy,
                cleanup_list *cleanup) NB_INLINE_LAMBDA -> PyObject * {
```
to line 275, same file
``` C++
            result = cast_out::from_cpp(
                       cap->func((in.template get<Is>())
                                     .operator cast_t<Args>()...),
                       policy, cleanup).ptr();

```
… which calls user code, which in my case calls back to Python, which may of course throw an exception.

### Reproducible example code

I didn't write a reproducer, sorry. The more interesting question is, which **other** `noexcept`s, besides the obvious `nb_vectorcall_complex`, have to go …

wjakob:
The `noexcept` is a non-negotiable policy. Nanobind does this to avoid generating large amounts of exception handling code. Usually any code that can raise an exception needs to be wrapped into code that converts it into the convention expected by the caller (a `bool` return value, or a Python error status code).

You gave an example of some dispatch code. The `catch` blocks for those are in `nb_func_vectorcall_*`.

smurfix:
OK, I might be able to work around that, but it  turns out that I also have a worse problem.

The destructor for `gil_scoped_release` calls `PyEval_RestoreThread`. Your guess what this can do if the thread's state is no longer valid for whatever reason would probably be correct.

wjakob:
Is that happening somewhere in nanobind, or in your code? In the latter case, you can check `nb::is_alive`.

smurfix:
[...]

PR #891, opened 2025-01-29
Added a comment about GIL acquisition vs. multithreading vs. program exit.

m-u-xyz:
I added a paragraph warning about this problem. Maybe it saves somebody else a few hours of work.

Cf. #890 where one of my problems was exactly this …

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -2737,6 +2737,12 @@ running it in parallel from multiple Python threads.
 
       Reacquire the GIL
 
+If you use nanobind in a multi-threaded program, you **must** ensure that
+all threads have terminated **before** the interpreter exits. Otherwise any
+code that re-acquires the GIL (e.g. for running destructors) will throw a
+cancellation exception. This will result in a hard crash because C++ does
+not support exceptions in destructors, nor ignoring a cancellation.
+
 Free-threading
 --------------
 


wjakob:
What does "nor ignoring a cancellation" mean?

I will note that ``nb::is_alive()`` can be used to skip deallocation/destructor code in some cases. It's not so nice because that causes leaks, but still infinitely preferable to segfaults.

smurfix:
When Python detects, as you try to acquire the GIL that a thread needs to die, it self-cancels. You can't catch this exception without re-throwing it.

I can't fix the problem by calling `is_alive` in my destructors because that's not sufficient if the object contains one of nanobind's. To that effect I created PR #894 which adds aliveness checks to all of nanobind's destructors if they directly acquire the GIL. I didn't yet check whether that's sufficient.

wjakob:
Hi @smurfix,

sorry for the long delay in getting to this. The information in this message is not quite accurate anymore: the crashing behavior was changed in recent Python versions. The issue is also not in threads being alive in general -- it's threads that work with nanobind objects, and specifically in such a way that the GIL is dropped and reacquired (which is the step that crashes on old python versions, and pauses those threads on newer versions). Would you consider updating the text to make this more clear, or shall I close the PR?

PR #894, opened 2025-01-30
Add missing "is_alive" checks before acquiring the GIL in destructors.

m-u-xyz:
cf. #891

### DIFF
--- a/docs/api_extra.rst
+++ b/docs/api_extra.rst
@@ -1412,7 +1412,7 @@ functions:
                   Py_INCREF(o);
               },
               [](PyObject * o) noexcept {
-                  if (! nb::is_alive())
+                  if (!nb::is_alive())
                       return;
                   nb::gil_scoped_acquire guard;
                   Py_DECREF(o);
--- a/docs/api_extra.rst
+++ b/docs/api_extra.rst
@@ -1412,12 +1412,18 @@ functions:
                   Py_INCREF(o);
               },
               [](PyObject * o) noexcept {
+                  if (! nb::is_alive())
+                      return;
                   nb::gil_scoped_acquire guard;
                   Py_DECREF(o);
               });
 
           // ...
       }
+
+   The liveness check in the destructor is necessary in a multi-threaded
+   environment because when the interpreter is shutting down, trying to
+   (re-) take the GIL will throw a cancellation exception.
 
 .. cpp:function:: inline void inc_ref(intrusive_base * o) noexcept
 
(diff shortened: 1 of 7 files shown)


src/nb_ndarray.cpp
```diff
@@ -662,6 +664,8 @@ ndarray_handle *ndarray_create(void *value, size_t ndim, const size_t *shape_in,
     scoped_pymalloc<int64_t> shape(ndim), strides(ndim);
 
     auto deleter = [](managed_dltensor *mt) {
+        if (! is_alive())
```
wjakob, line 667: very minor: without space after `!`

src/nb_ndarray.cpp
```diff
@@ -293,6 +293,8 @@ static PyObject *dlpack_from_buffer_protocol(PyObject *o, bool ro) {
     }
 
     mt->deleter = [](managed_dltensor *mt2) {
+        if(! is_alive())
```
wjakob, line 296: very minor: without space after `!`

docs/api_extra.rst
```diff
@@ -1412,13 +1412,19 @@ functions:
                   Py_INCREF(o);
               },
               [](PyObject * o) noexcept {
+                  if (! nb::is_alive())
```
wjakob, line 1415: and here

wjakob:
Could you add a changelog as well?

smurfix:
Done.

We'll see whether that is sufficient …

src/nb_ndarray.cpp
wjakob: still there?

smurfix: Eh. Missed that one. Fixed.

wjakob:
After some more thought, I don't think that this really fixes the issue.

While `nb::is_alive()` might return `True` when the condition is checked, there is no guarantee that this is still the case at the next instruction. Basically this is a race condition. If you want to avoid undefined behavior, you will need to prevent this problem in another way.

smurfix:
That is true in general. But what happens when the Python program throws an uncaught exception or otherwise ends abnormally, or simply is signalled to end?

I'm not concerned with preventing a 0.1% chance of a hard coredump and whatnot due to a race condition in this case. I'm concerned with preventing a 100% chance of getting one.

wjakob:
By having a condition variable, mutex, or similar synchronization mechanism, you can guarantee that the right order is enforced during shutdown. If the application is killed, then the kernel will shut things down and none of this code will run at all.

A crash that happens in 0.1% of runs is super annoying because it is so hard to reproduce. I would rather have software fail spectacularly than accumulate lots of issues in the long tail.

wjakob:
-> https://discuss.python.org/t/safely-using-the-c-api-when-python-might-shut-down/78850

PR #900, opened 2025-01-31
The stub generator needs to emit aliases *after* the item they're aliasing.

### LINKED ISSUE #899: [BUG]: stubgen generates aliases out-of-order
### Problem description

This is unlikely to work:

``` python3
class _ttd_enum_DirDiff(enum.IntFlag):
    __str__ = __repr__
    def __repr__(self, /):
        """Return repr(self)."""
    SAME = 0
    ...
```

### Reproducible example code

```text
py::enum_<DirDiff>(m, "DirDiff" ,py::is_flag(),py::is_arithmetic())
        .value("SAME", 2)
```

smurfix:
What a surprise. ;-)

Fixes #899

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -376,7 +376,7 @@ class StubGen:
                 self.write_ln(f"@{overload}")
                 self.put_nb_overload(fn, s, name)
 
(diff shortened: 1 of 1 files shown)


wjakob:
I believe this issue is now fixed indirectly by #938. Please chime in if you find this to not be the case.

PR #911, opened 2025-02-05
Fix off-by-one error in nb::call_policy documentation

oremanj:
As noted by @wojdyr in https://github.com/wjakob/nanobind/discussions/816#discussioncomment-11561279 - thanks for catching!


wjakob:
Thanks!

PR #913, opened 2025-02-06
Fix underalignment of nanobind-bound classes.

hawkinsp:
See https://github.com/python/cpython/issues/129675#issuecomment-2638095421
`tp_basicsize` needs to be a multiple of the alignment of `PyObject`, which is in effect pointer aligned.

This fixes a SIGBUS alignment crash found when testing free-threaded nanobind-using code on aarch64.


wjakob:
Thanks!

PR #914, opened 2025-02-06
Using typing_extensions.CapsuleType as fallback for types.CapsuleType

### LINKED ISSUE #912: [BUG]: Capsules are reported as types.CapsuleType, which only exists on 3.13+
### Problem description

Title. My use case is with stubgen, where the generated stubfiles will be used back to 3.9. So importing them with those type annotations will cause errors.

### Possible solution

Would it be reasonable to have this name be conditioned on the python version, that is, if we're on <3.13 then use [typing_extensions.CapsuleType](https://github.com/python/typing_extensions/pull/390)? Then as long as the stubfiles are generated with the lowest supported version then things shouldn't be an issue. That does come with the caveat that typing_extensions is now an implicit dependency, but for me personally that's perfectly fine.

Ahajha:
Fixes #912


Ahajha:
@wjakob The tests fail, but the result is now conditioned on Python version, which is new. Should I find a way to make the tests conditional as well (probably having a reference version per Python version, or just 3.13+ and 3.12- for now), or just unconditionally use typing_extensions?

wjakob:
Let's make the test conditional as well. Can you move the definition of the name of this type to `include/nb_defs.h`? You will see that there are many similar definitions at the top of this file.

Ahajha:
Sure thing - will update shortly.

Ahajha:
The test split is a little hacky, but could probably be made extendable with a bit more work. Indifferent to leaving it as is (for the next person to extend, whenever that is) or fixing it up now.

wjakob:
That's a little too hacky :-). How about "canonicalizing" the two strings being compared to each other? We could convert all `typing_extension.Capsule` substrings to `types.Capsule`, and then they will match.

Ahajha:
Ah, so basically
- Read file
- Conditionally replace `typing_extensions.CapsuleType` with `types.CapsuleType` in the string (or vice versa, depending on which file we make the 'canonical' one)
- Then compare

Changes incoming...

Ahajha:
Much happier with this version.

Ahajha:
@wjakob Can you please take another look when you get a chance?

wjakob:
Great, thank you!

PR #916, opened 2025-02-06
ndarray: use clone to copy return pytorch tensors

tstastny:
Using `rv_policy::copy` for torch tensors is currently broken. Produces the following error:
```
RuntimeError: nanobind::detail::ndarray_export(): copy failed: AttributeError: 'Tensor' object has no attribute 'copy'
```

The error message is correct, torch tensors don't have an out-of-place `copy` attribute, they have `clone()`.

I added a simple if statement to change out the attribute name in the ndarray export function. If other frameworks have similar issues, could be expanded to a switch case or so.

### DIFF
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -803,8 +803,12 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
     }
 
     if (copy) {
+        const char* copy_str = "copy";
+        if (framework == pytorch::value)
+            copy_str = "clone";
+        
         try {
-            o = o.attr("copy")();
+            o = o.attr(copy_str)();
         } catch (std::exception &e) {
             PyErr_Format(PyExc_RuntimeError,
                          "nanobind::detail::ndarray_export(): copy failed: %s",


src/nb_ndarray.cpp
```diff
@@ -803,8 +804,12 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
     }
 
     if (copy) {
+        std::string copy_str("copy");
```
wjakob, line 807: This looks good, but can you please use a `const char *` here? There is no need for a `std::string`.

You can remove the braches (`{`) for the `if` statement, this codebase doesn't use them if the `then` condition is a simple one-liner.

tstastny: used const char*, removed braces, and squashed

wjakob:
Thanks!

wbthomason:
Would it be possible to get a v2.5.1 tag with this fix included?

tstastny:
> Would it be possible to get a v2.5.1 tag with this fix included?

@wjakob my team would also be happy for a tag that rolls into the pypi package if it's possible :)

wjakob:
I made a v2.6.0 release with the changes accumulated so far (which includes an ABI break, hence the minor version bump -- changing the ABI is not something I wish to do in a patch version).

tstastny:
thank you!

wthomason-bdai:
Excellent, thanks very much!

wbthomason:
Excellent, thanks very much!

PR #928, opened 2025-02-13
Add missing #pragma once

marktsuchida:
This adds `#pragma once` to the following headers:

- The public header `<nanobind/typing.h>`, which would result in an error if included twice.

- The private header `src/nb_ft.h`, which _is_ included twice by `nb_combined.cpp`, resulting in an error (at least in my environment).

(There are additional private headers that lack `#pragma once`; I'm leaving those alone here.)


wjakob:
Thanks!

PR #931, opened 2025-02-13
Add input types to chrono type casters

tjstum:
The type casters in `stl/chrono.h` accept various types in their
`from_python` calls, but they currently just use a name for their
output type.

This uses `io_name` to differentiate, adding the various acceptable
types that the casters permit at runtime

This also includes type stub testing for the `test_chrono_ext` module. The
first commit in this PR contains the stubs only.

### DIFF
--- a/include/nanobind/stl/chrono.h
+++ b/include/nanobind/stl/chrono.h
@@ -99,7 +99,13 @@ public:
         return pack_timedelta(dd.count(), ss.count(), us.count());
     }
 
-    NB_TYPE_CASTER(type, const_name("datetime.timedelta"))
+    #if PY_VERSION_HEX < 0x03090000
+        NB_TYPE_CASTER(type, io_name("typing.Union[datetime.timedelta, float]",
+                                     "datetime.timedelta"))
+    #else
+        NB_TYPE_CASTER(type, io_name("datetime.timedelta | float",
+                                     "datetime.timedelta"))
+    #endif
 };
 
 template <class... Args>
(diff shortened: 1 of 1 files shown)


tjstum:
I now understand why there aren't stubs for these functions. :-)

Is there a straightforward way to tell stubgen to emit the module attribute value as `= ...` in the stubs (instead of as `= True` or `= False`)? It seems like, in general, we would always want that (since the runtime value isn't relevant: just the type). I could make that change as a separate PR.

Failing that, I can reduce this PR to just the type caster change and then just leave this early version as "proof" of its correctness.

include/nanobind/stl/chrono.h
```diff
@@ -99,7 +99,13 @@ template <typename type> class duration_caster {
         return pack_timedelta(dd.count(), ss.count(), us.count());
     }
 
-    NB_TYPE_CASTER(type, const_name("datetime.timedelta"))
+    #if PY_VERSION_HEX < 0x03090000
+        NB_TYPE_CASTER(type, io_name("typing.Union[datetime.timedelta, float]",
+                                     "datetime.timedelta"))
```
Hnasar, line 104: in stubs, it's fine to use the modern syntax all the time. 
Typecheckers treat stubs as all having `from __future__ import annotations`

wjakob: In the past, there were users who wanted old-style annotations when working with old Python versions, and to import various things from `typing_extensions` that potentially don't exist on them (even though it's just the stubs!) 🤷.

tjstum: The rest of nanobind's type casters use this form (`typing.Union`), so I'm going to stay consistent with that.

tjstum:
I've reduced the PR to just the type caster change. Hopefully this is amenable!

wjakob:
Thanks!

PR #936, opened 2025-02-17
Support for implicit bool casting and np.bool

havogt:
Adds support for implicit conversion of bools and support for np.bool -> bool.

This is copied over from pybind11, originally implemented in https://github.com/pybind/pybind11/pull/925.


tests/test_functions.py
```diff
@@ -745,8 +754,47 @@ def case(arg1, arg2, expect_ret):  # type: (str, str, str | None) -> str
...
+    # straight up bool
+    assert t.test_11_b(True) is True
+    assert t.test_11_b(False) is False
+    assert t.test_11_bnc(True) is True
+    assert t.test_11_bnc(False) is False
+
+    # None requires implicit conversion
+    with pytest.raises(TypeError):
+        assert t.test_11_bnc(None)
+    # t.test_11_b(None) is False
```
havogt, line 775: I couldn't get the `None` case to work directly, it seems we don't end up in the type caster for `bool`, but didn't debug further. Maybe someone with better understanding of the code paths, can quickly spot the problem, otherwise I could spend more time debugging.

include/nanobind/nb_cast.h
```diff
@@ -254,13 +254,41 @@ template <typename T> struct none_caster {
...
         } else if (src.ptr() == Py_False) {
             value = false;
             return true;
+        } else if (flags & (uint8_t)cast_flags::convert ||
+                   // allow non-implicit conversion for numpy booleans
+                   is_numpy_bool(src)) {
+            Py_ssize_t res = -1;
+            if (src.is_none()) {
+                res = 0; // None is implicitly converted to False
+            }
```
havogt, line 270: This is currently not working, see comment in test.

tests/test_functions.py
wojdyr: You'd need `nb::arg().none()` in test_11_b, but perhaps wait for Wenzel's opinion on enabling such conversion.

wjakob:
I don't think I _want_ implicit bool casting for C++ bindings. I have always found this super error prone -- it is an intentional omission from nanobind.

havogt:
Thanks for the feedback! What about the np.bool part?

PR #938, opened 2025-02-17
fix: fix stubgen to sort stub module members according to module __dict__ 

tlambert03:
hey @wjakob, thanks for the awesome package.

this relates to #899 (but may not be exactly the same thing).  My stubs are getting generated out of order, and therefore have `Undefined name` errors for some default values when running things like ruff on the output,  due to `inspect.getmembers()` sorting members alphabetically.  

I recognize this is a tricky topic, so this PR is designed to *do nothing* by default, but leave a little bit of agency to the end user.  It adds a new `member_sort` parameter to `StubGen`

```python
class StubGen:
    def __init__(
        self,
        ...
        member_sort: Optional[Callable[[tuple[str, Any]], Any]] | Literal['as-defined'] = None,
    ) -> None:
```

if a Callable or None (the default) is passed, it is used directly in `sorted(getmembers(value), key=self. member_sort)`

if the literal string `'as-defined'` is passed, then the module members are sorted as they are defined in the module (using the module `__dict__`).  

let me know what you think


src/stubgen.py
```diff
@@ -178,7 +178,8 @@ def __init__(
         max_expr_length: int = 50,
         patterns: List[ReplacePattern] = [],
         quiet: bool = True,
-        output_file: Optional[Path] = None
+        output_file: Optional[Path] = None,
+        member_sort: Optional[Union[Callable[[Tuple[str, Any]], Any], Literal['as-defined']] = None,
```
riley-kinahan, line 182:
```suggestion
        member_sort: Optional[Union[Callable[[Tuple[str, Any]], Any], Literal['as-defined']]] = None,
```
My bad, I dropped a closing bracket in my suggestion

nurpax:
Can this also be turned on the CLI or enabled in cmake rules?

I'd like to use this in the generated pyi file as this makes the output much more useful if it can retain the original module's logical API grouping. I
imagine the pyi file being one of the primary ways the users of an extension can browse the API ref.

tlambert03:
I can look at adding that.  But will wait to see what @wjakob thinks before going further

nurpax:
I wonder if the original problem statement is stated too verbosely for this to get traction. :) I still think this is something that is valuable.

The problem is that due to how Python APIs return module members, the symbols in the stubgen pyi output are by default sorted in alphabetical order.  I would argue that this is the wrong default but that's how the Python API works.  IMO a more useful default order would be the output exactly in the same order as the symbols are defined in the Nanobind extension code.

This change adds the option to keep the original order.

wjakob:
Sorry about the delay in looking at this, I had a stressful deadline. Can you explain what "generated out of order" means" (as opposed to dictionary assignment order)? What is the custom callback you use in your project, and how does this help improve the situation? The ability to pass custom callbacks will be too complex for most, and I am wondering if it's not better to offer more ordering options.

tlambert03:
Thanks @wjakob, sorry for the unclear original explanation:  here's are the important points:

1. by "generated out of order" I mean that module members appear in *alphabetical order* in stub files generated by Stubgen, rather than the order in which they were declared in the module's `__dict__`.  
2. This is caused by the use of `inspect.getmembers` [here](https://github.com/wjakob/nanobind/blob/e4ad3dfb00225467e5eb557a85e57c9983ae0ad0/src/stubgen.py#L848C21-L849). `inspect.getmembers` states in the [docs](https://docs.python.org/3/library/inspect.html#inspect.getmembers) that it "returns all the members of an object in a list of (name, value) pairs sorted by name."
3. That creates `UndefinedName` errors if you attempt to parse/format the type stubs using (for example) `ruff`.

> The ability to pass custom callbacks will be too complex for most, and I am wondering if it's not better to offer more ordering options.

that's fine by me.  I could change the signature to:

```python
member_sort: Literal['as-defined', 'alphabetical'] = 'as-defined',
```

can you think of any other orders you'd want to offer?

wjakob:
Hi @tlambert03,

how about we switch to `as-defined` as new default? In that case, is the customization ability still needed? It was not my intention to use alphabetic sorting. If this sounds okay, would you be able to change the commit so that it adopts this ordering?

tlambert03:
Absolutely, will do tomorrow 👍

tlambert03:
ok @wjakob, I changed this PR to be the bare minimal changeset to sort stub module members according to `__dict__` insertion, and removed all user-adjustable features.   This *does* now change the default behavior, but if alphabetical was never intended, then I think it's a good bugfix that needs no user option.  let me know if you'd like any additional changes

tlambert03:
i see that tests are failing and will fix soon

tlambert03:
I think this is all set now.  Note i deleted [one stub file in the tests](https://github.com/wjakob/nanobind/pull/938/files#diff-8dee0d24e890f2c86fd4134ccac499d5b5b926e67857d8ee9ec82a6e657e7be4) that seemed to be unnecessary.  lemme know if that was incorrect.

wjakob:
Awesome, thanks! Let's merge this and wait for a bit to see if anyone runs into issues with it.

tlambert03:
Thanks!

PR #939, opened 2025-02-18
stubgen: Fix module search of nested modules

ingomueller-net:
This PR fixes the logic that tries to find which dot (`.`) separates the module from the class in a nested type name in `simplify_types`. The current logic flipped the components, creating invalid type names.


wjakob:
Subsumed by #940.

PR #950, opened 2025-02-27
Allow taking optional by value in function arguments

Walter-Reactor:
This appears to be a bug, or perhaps just an oversight?


wjakob:
See issue #951. The fact that `.none()` has to be specified is indeed somewhat redundant, but is a consequence of the design of this library. I don't see a way of fixing this without adding quite a bit of complexity and/or adding runtime cost.

Walter-Reactor:
Gotcha. Would it be possible to note this in the documentation, or at least in a comment in the `stl/option.h` header?

PR #972, opened 2025-03-14
Create a metaclass for each supplemental data type

oremanj:
This allows many metaclass customizations without incurring the performance cost of allowing arbitrary metaclasses. As a side benefit, we get nontrivial construction and destruction of supplement types, and can check whether a particular nanobind type uses a particular supplement type. There are no additional per-instance performance or storage costs, no additional per-type storage costs, and negligible per-type performance costs.

The information about a particular supplement type is stored in a new `nb::detail::supplement_data` structure. An inline variable template is used to automatically create one of these in static data for each supplement type used in a particular extension module. Instances of `nb_meta` (i.e., metaclass types) gain a pointer to their `supplement_data`, which is null for the default metaclass `nb_type_0`. The non-default metaclass types `nb_type_xyz` are now stored in a `nb_type_map_slow` rather than a Python dictionary, and (except on free-threaded builds where everything is immortal) they are properly reference-counted so that they will be destroyed once all their instances are. This avoids leak warnings when assigning nanobind functions as their attributes.

Size cost:
* On my laptop this adds 781 bytes to a release build of libnanobind-static.a.
* One pointer per nanobind metatype, including the default one.
* One `nb::detail::supplement_data` structure (five pointers) stored in static data per supplement type used in an extension module.

I originally added some caching in static data to limit lookups by supplement type to one per supplement type per extension module. I removed this because it was a bit complex and I don't think it's necessary (we already do two type_map_slow lookups per nb_type_new, avoiding a third doesn't save all that much) but it would be easy to re-add. See commit 34739e3. The additional cost from that caching would be 104 bytes of code and two more pointers in `supplement_data`.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -2059,9 +2059,20 @@ declarations in generated :ref:`stubs <stubs>`,
 
 .. cpp:struct:: template <typename T> supplement
 
-   Indicate that ``sizeof(T)`` bytes of memory should be set aside to
-   store supplemental data in the type object. See :ref:`Supplemental
-   type data <supplement>` for more information.
+   Indicate that the type being created (not its instances) should contain
+   a subobject of type ``T``, which can later be accessed using
+   :cpp:func:`nb::type_supplement\<T\>() <type_supplement>`. nanobind will
+   create a separate metaclass per type ``T``. See
+   :ref:`Supplemental type data <supplement>` for more information.
+
+   .. cpp::function:: supplement&& inheritable() &&
+
+      By default, specifying a supplement for a type prevents that type from
+      being inherited, since there would be no general way to initialize the
+      supplemental data for the derived type. If you instead pass the binding
+      annotation ``nb::supplement<T>().inheritable()``, inheritance will be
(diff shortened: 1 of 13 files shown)


oremanj:
The tests aren't running because the tests for a previous iteration hung. I don't have permission to stop the previous run https://github.com/wjakob/nanobind/actions/runs/13862493485. Maybe worth setting a timeout in github actions?

oremanj:
Well, I guess that saves me from needing to figure out the reference leak on earlier Pythons! Thanks for your consideration. I came up with this when I noticed that nanobind *did* have a way to do metaclass customizations — “use a supplement type of a size that no one else uses, then grab the metaclass and assign attributes on it” — and figured there must be a better way. If there are simplifications (such as not supporting non-trivial supplement types) that would make you more comfortable with this change, that would be useful to know, but I understand if you’d prefer to just not go there at all.

I’m not sure what you mean by “Transport law, do we present it?” though.

wjakob:
> I’m not sure what you mean by “Transport law, do we present it?” though.

A copy paste error, sorry. I think that I would like the supplement feature at the rather minimalist level that it currently has now.

Issue #976, opened 2025-03-21
[BUG]: STABLE API support till python 3.12


petrasvestartas:
### Problem description

Hi,

We had a long discussion for building nanobind project with cibuildwheels due to github action automation to support various python version.

It was really hard to understand what was happening when trying to building multiple wheels for python 3.9, 3.10, 3.11, 3.12. The problem is that Nanobind does not support stable API until python 3.12.

Will this be fixed in the future?

### Reproducible example code

Here is the discussion that might be interesting to read:

https://github.com/scikit-build/scikit-build-core/issues/1017#issuecomment-2743503126

PR #978, opened 2025-03-21
fix(cmake): error if SKBUILD_SABI_VERSION set and too low

henryiii:
This was confusing a user in https://github.com/scikit-build/scikit-build-core/issues/1017 and https://github.com/wjakob/nanobind/issues/976. I think this would really help, since it would have been a clear, readable error instead of a wheel that worked on the CPython it was built on but claimed it was Stable ABI.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -26,6 +26,12 @@ elseif(DEFINED Python_SOSABI)
   set(NB_SOSABI "${Python_SOSABI}")
 endif()
 
+
+# Error if scikit-build-core is trying to build Stable ABI < 3.12 wheels
+if(DEFINED SKBUILD_SABI_VERSION AND SKBUILD_SABI_VERSION VERSION_LESS "3.12")
+  message(FATAL_ERROR "You must set 'tool.scikit-build.wheel.py-api to cp312 or later when using scikit-build-core with nanobind.")
+endif()
+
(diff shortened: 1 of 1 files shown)


cmake/nanobind-config.cmake
```diff
@@ -26,6 +26,12 @@ elseif(DEFINED Python_SOSABI)
   set(NB_SOSABI "${Python_SOSABI}")
 endif()
 
+
+# Error if scikit-build-core is trying to build Stable ABI < 3.12 wheels
+if(DEFINED SKBUILD_SABI_VERSION AND SKBUILD_SABI_VERSION VERSION_LESS "3.12")
+  message(FATAL_ERROR "You must set 'tool.scikit-build.wheel.py-api to cp312 or later when using scikit-build-core with nanobind.")
```
henryiii, line 32:
```suggestion
  message(FATAL_ERROR "You must set tool.scikit-build.wheel.py-api to 'cp312' or later when using scikit-build-core with nanobind.")
```

wjakob:
Sure, thanks!

henryiii:
Maybe I should have printed out what it actually gets too.  This shouldn't have tried to do a Stable ABI build at all. Do you have a repo I can try out? Otherwise, this is easy to set up, I think.

henryiii:
Ahh, I see it, we define it to `""` when it's not building Limited ABI wheels. Sorry, will fix.

PR #980, opened 2025-03-24
fix(cmake): Handle empty SKBUILD_SABI_VERSION

henryiii:
This is an empty string when it's not building for the Stable ABI.

I chose to check if it's defined and check if it's empty in two steps, though if you'd prefer, the empty check also is False if it's not defined. The form I used is slightly clearer, and might work a little better if someone passes `--warn-uninitialized` (maybe the second one passes that too, not sure).

Fix for regression in #978.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -28,8 +28,9 @@ endif()
 
 
 # Error if scikit-build-core is trying to build Stable ABI < 3.12 wheels
-if(DEFINED SKBUILD_SABI_VERSION AND SKBUILD_SABI_VERSION VERSION_LESS "3.12")
-  message(FATAL_ERROR "You must set tool.scikit-build.wheel.py-api to 'cp312' or later when using scikit-build-core with nanobind.")
+if(DEFINED SKBUILD_SABI_VERSION AND SKBUILD_ABI_VERSION AND SKBUILD_SABI_VERSION VERSION_LESS "3.12")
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @henryiii,

I unfortunately saw this PR too late and had pushed out v2.6.0. Is this a serious regression, i.e., do we need a v2.6.1 ASAP?

garth-wells:
v2.6.1 ASAP would be very helpful. Builds against v2.6.0 when using  scikit-build-core fail.

hawkinsp:
We noticed this caused a failure in one of our CI builds also, so a +1 on "worth a new point release".

pthom:
+1 on my side too, my [CI](https://github.com/pthom/imgui_bundle/actions/runs/14128029833/job/39581443761#step:6:403
) just started to fail with a message "You must set tool.scikit-build.wheel.py-api to 'cp312' or later when using scikit-build-core with nanobind."

henryiii:
Yes, making a point release and maybe yanking would be a good idea.

wjakob:
Done just now.

kwk:
We ran into similar problems and hopefully [the packaging](https://src.fedoraproject.org/rpms/python-nanobind/pull-request/2) is not complaining with 2.6.1.

pthom:
> Done just now.

I confirm it works on my CI, many thanks for the quick fix!

kwk:
> Done just now.

Our CI is also working with 2.6.1 now! Thank you for the quick release.

PR #987, opened 2025-04-02
Support from-Python conversion of non-default-constructible std::variants

oremanj:


wjakob:
This looks great, thank you!

PR #988, opened 2025-04-02
Support casts of const-qualified unique_ptrs

tjstum:
This uses the same technique that the `shared_ptr` type caster does. The updated test previously failed because of an invalid conversion from `const void*` to `void*`


wjakob:
LGTM, thanks!

PR #990, opened 2025-04-04
Fix type in put_nb_static_property stubgen

mrexodia:
This was generating invalid `.pyi` files with code like this:

```cpp
nb::class_<NbDatabase>(m, "Database")
	.def_prop_ro_static("current", &NbDatabase::get_current, "Get the current database")
```


wjakob:
@mrexodia This is breaking the test suite. Could you instead create an issue with a self-contained reproducer demonstrating the problem you encountered?

Issue #994, opened 2025-04-05
[BUG]: NB_SUPPRESS_WARNINGS should also disable clang-tidy


JeWe37:
### Problem description

Recently, #868 introduced `NB_SUPPRESS_WARNINGS` to suppress warnings from pybind11. However, unlike the solution proposed in https://github.com/wjakob/nanobind/issues/850#issuecomment-2583398359, it does not disable clang-tidy via the `CXX_CLANG_TIDY` target property, causing warnings/errors from it to still prevent successful builds.

Can be fixed by setting
```
set_target_properties(nanobind-static PROPERTIES CXX_CLANG_TIDY "")
```
currently manually, but this should likely happen in nanobind_build_library directly.

### Reproducible example code

```text

```

PR #997, opened 2025-04-08
Disable clang-tidy on nanobind library

JeWe37:
Pull request for #994 as requested. Not disabling this could lead to build failures in projects where clang-tidy is globally enabled if one included the project as a submodule.


wjakob:
This sets the flag unconditionally? Wasn't the plan to only set this if `nanobind_add_module()` was called with `NB_SUPPRESS_WARNINGS`?

JeWe37:
Fixed that, I had forgotten about that, sorry.

wjakob:
Still looks the same to me, did you push the change?

wjakob:
Thanks!

PR #1000, opened 2025-04-08
Make stubgen compatible with sanitizers

wjakob:
In CMake build systems using ``nanobind_add_stub()``, a tricky issue arises when the extension is built with sanitizers. Unless Python itself is built with sanitizers as well, this generally requires injecting a special sanitizer library into the Python executable at startup time. Failing to do so causes the Python interpreter to abort when the sanitized extension is imported. stubgen needs to run Python and import the built extension, so this combination simply did not work so far.

This commit extends the CMake implementation of ``nanobind_add_stub()`` so that it searches the dependency list (``DEPENDS``) for any targets that use sanitizers. It figures out the right injection library, deals with a tricky issue on macOS (we must find out the true Python executable, which is often hidden behind several layers of shims), and then runs stubgen with sanitizers pre-injected.

This feature only works on macOS and Linux. I don't know enough about sanitizers on Windows, external contributions in this regard are welcome.


maximiliank:
Great idea to handle the `LD_PRELOAD` automatically. Unfortunately, extra care needs to be taken in case of Linux systems using the [INPUT Syntax](https://unix.stackexchange.com/questions/302402/what-is-the-syntax-of-input-in-so-file), e.g. Fedora/RHEL. There it looks like:

```shell
cat `g++ -print-file-name=libasan.so`
/* GNU ld script */
OUTPUT_FORMAT(elf64-x86-64)
INPUT ( /usr/lib64/libasan.so.8 )
```

One way to resolve this is to compile a program with `-fsanitize=address;-fPIC;-shared` and then use `ldd` to parse for the real location. A cmake example can be found on [stackoverflow](https://stackoverflow.com/a/76836372/9759769).

I am also not aware of sanitizers on Windows.

wjakob:
Thanks for making me aware of this.

> One way to resolve this is to compile a program with -fsanitize=address;-fPIC;-shared and then use ldd to parse for the real location. A cmake example can be found on [stackoverflow](https://stackoverflow.com/a/76836372/9759769).

That sounds really complicated. Is it also an acceptable solution to just extract the argument of the `INPUT(..)` line?

Do you have access to RHEL/Fedora machine and could test that this PR works there?

maximiliank:
I checked it and here are my remarks:

* Add `INTERFACE_LINK_OPTIONS` and `INTERFACE_COMPILE_OPTIONS` to the properties which are searched in case the sanitizer was added via `INTERFACE`, e.g. `target_compile_options(python_ext INTERFACE -fsanitize=address,undefined)`
* If possible, add `TRANSITIVE_LINK_PROPERTIES` and `TRANSITIVE_COMPILE_PROPERTIES` if cmake >= 3.30 is available
* Remove quotes from detected `san_flags` to allow `-fsanitizer="address,undefined"`
* Add `ASAN_OPTIONS=detect_leaks=0` to the `NB_STUBGEN_CMD` command, otherwise the generation may fail in case of `pymalloc` or other things
* `set(is_clang CMAKE_CXX_COMPILER_ID MATCHES "Clang")` behaves strange on Linux with gcc, my `CMAKE_CXX_COMPILER_ID` is `GNU` but it still enters `elseif(is_clang)`. It should set the `san_libname` like in this branch, but still it is confusing to enter `is_clang` with `gcc`

wjakob:
@maximiliank Thanks for the feedback, could you take another look?

maximiliank:
Almost. In the `APPLE` and `Clang` check there is an `elseif()` only and no `else()`. Further the `san_libname` should be also different for `Clang` on Linux. The block should be

```cmake
      if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
        if(APPLE)
          set(san_libname "libclang_rt.${san}_osx_dynamic.dylib")
        else()
          set(san_libname "libclang_rt.${san}.so")
        endif()
      else()
        set(san_libname "lib${san}.so")
      endif()
```

With this it works for me on `RHEL` with `gcc` and `clang`. Also the props check works and the `detekt_leaks=0`.

The only case which currently does not work is to resolve transitive `INTERFACE` link options and compiler options, but I think there is no possibility in `CMake` todo so.

cmake/nanobind-config.cmake
```diff
@@ -422,6 +421,142 @@ function(nanobind_add_module name)
...
+  if (detected_san)
+    list(REMOVE_DUPLICATES detected_san)
+    set(libs "")
+
+    foreach(san ${detected_san})
+      set(san_libname "")
+
+      if (APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        set(san_libname "libclang_rt.${san}_osx_dynamic.dylib")
+      elseif()
```
maximiliank, line 488: Should be

```cmake
      if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
        if(APPLE)
          set(san_libname "libclang_rt.${san}_osx_dynamic.dylib")
        else()
          set(san_libname "libclang_rt.${san}.so")
        endif()
      else()
        set(san_libname "lib${san}.so")
      endif()
```

wjakob:
Thanks a lot @maximiliank !

starball5:
I use generator expressions for sanitizer flags. Like this: `"$<$<CONFIG:Debug,RelWithDebInfo>:-fsanitize=address,undefined>"`. So `san_list` is `address;undefined>` and `ubsan` isn't getting added to `detected_san`. Generator expressions are tricky or impossible to handle robustly at configure time, but I feel like my use case is worth handling. Could it please be handled?

starball5:
I'm also interested in the ldd option (or `objdump -p /path/to/program | grep NEEDED` which ldd recommends for untrusted binaries to prevent potential execution of untrusted code).

CMake has [`file(GET_RUNTIME_DEPENDENCIES [...])`](https://cmake.org/cmake/help/latest/command/file.html#get-runtime-dependencies) which you could use as a cross-platform way to get the library dependencies. That could be used in a wrapper script in `NB_STUBGEN_CMD` that gets the list of dependencies and then runs `cmake -E env ...`.

Aside: Is this feature necessary when Clang is using static sanitizer libraries? As far as I know, using shared library versions of sanitizers on clang is opt-in with `-shared-libasan`.

PR #1003, opened 2025-04-09
Support `Eigen::Map<Eigen::SparseMatrix>` [updated]

wjakob:
This PR adds support for taking and returning sparse matrix maps (including complex-valued ones). As with passing dense maps, the type caster is extra picky and only accepts the zero-copy case. There is now also some extra documentation that explains caveats about returning maps in general.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -201,6 +201,10 @@ Version 2.3.0
 
 There is no version 2.3.0 due to a deployment mishap.
 
+- Added casters for `Eigen::Map<Eigen::SparseMatrix<...>` types from the `Eigen library
+  <https://eigen.tuxfamily.org/index.php?title=Main_Page>`__. (PR `#782
+  <https://github.com/wjakob/nanobind/pull/782>`_).
+
 Version 2.2.0 (October 3, 2024)
 -------------------------------
 
--- a/docs/eigen.rst
+++ b/docs/eigen.rst
@@ -145,6 +145,20 @@ apply:
 
      void f4(nb::DRef<Eigen::MatrixXf> x) { x *= 2; }
 
+Maps
+----
+
+Besides ``Eigen::Ref<...>``, nanobind also supports binding functions that take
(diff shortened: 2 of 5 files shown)


wjakob:
@alecjacobson Could I ask you to give this a quick try and see if it addresses the copying issues in libigl?

alecjacobson:
I'll be very happy to. I'm crunching for a deadline tomorrow, but should be able to try it shortly after.

alecjacobson:
I made some quick test code at https://github.com/libigl/libigl-python-bindings/tree/alecjacobson/bump-nanobind

The example C++ bindings using `Eigen::Map<Eigen::SparseMatrix` (in the style of the libigl bindings)

https://github.com/libigl/libigl-python-bindings/blob/alecjacobson/bump-nanobind/src/sparse_map_noop.cpp

https://github.com/libigl/libigl-python-bindings/blob/alecjacobson/bump-nanobind/src/sparse_map_shape.cpp

(there are corresponding ones without `_map` using `Eigen::SparseMatrix`)

Specifically the python test code is in https://github.com/libigl/libigl-python-bindings/blob/alecjacobson/bump-nanobind/test_sparse.py

This prints out timing info `[matrix dim] [time sparse_shape] [time sparse_noop] [time_sparse_map_shape] [time_sparse_map_noop]`

```
40962 0.000920057 0.00110698 5.96046e-06 6.00815e-05
163842 0.0034461 0.004601 9.05991e-06 0.000133038
655362 0.0171809 0.020237 9.05991e-06 0.000243902
2621442 0.0688672 0.092025 2.00272e-05 0.00264215
10485762 0.284097 0.365652 2.5034e-05 0.00990915
41943042 1.02176 1.46509 4.19617e-05 0.0410988
```

Way faster with Map but seems it's not quite "no cost" even with map (and it scales with matrix size). Maybe I'm doing something wrong in the bindings. I'll continue to investigate.

wjakob:
Any other comments? If not, I will go ahead and merge this.

alecjacobson:
I think this is great, thank you!

PR #1004, opened 2025-04-10
Use Py_XINCREF,Py_XDECREF also with Py_LIMITED_API

hpkfft:
My understanding is that #500 concluded that it is fine to use `Py_XINCREF` and `Py_XDECREF` with `Py_LIMITED_API`.
If so, it seems preferable from a performance point of view.

Please review this PR carefully and correct me if I'm wrong.   I'm not a Python expert.


wjakob:
This looks fine to me. To be honest, I don't understand why it wasn't already implemented that way.

wjakob:
Thanks for catching this!

PR #1008, opened 2025-04-10
enum: __repr__ != __str__

wojdyr:
For Python enums, repr() differs from str()
for example:

Enum
str(): Color.RED
repr(): <Color.RED: 1>

Flag
str(): Permissions.READ|WRITE
repr(): <Permissions.READ|WRITE: 3>

IntFlag
str(): 3
repr(): <Colors.YELLOW: 3>

in nanobind, `__repr__` was set to `PythonEnumType.__str__`

This change sets it to `PythonEnumType.__repr__` instead

ping @skallweitNV @nicholasjng

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -72,8 +72,8 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
     scope.attr(name) = result;
     result.attr("__doc__") = ed->docstr ? str(ed->docstr) : none();
 
-    result.attr("__str__") = enum_mod.attr(is_flag ? factory_name : "Enum").attr("__str__");
-    result.attr("__repr__") = result.attr("__str__");
+    result.attr("__str__") = factory.attr("__str__");
+    result.attr("__repr__") = factory.attr("__repr__");
 
     type_init_data *t = new type_init_data();
     memset(t, 0, sizeof(type_data));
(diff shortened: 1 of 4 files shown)


src/nb_enum.cpp
```diff
@@ -72,8 +72,8 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
     scope.attr(name) = result;
     result.attr("__doc__") = ed->docstr ? str(ed->docstr) : none();
 
-    result.attr("__str__") = enum_mod.attr(is_flag ? factory_name : "Enum").attr("__str__");
-    result.attr("__repr__") = result.attr("__str__");
```
nicholasjng, line 76: I seem to remember that this was my first solution as well, but I was later asked to revert it, I think due to @wjakob's preference for how enum values render when using `__str__`. 

Not sure how things stand currently, just for your information.

wojdyr: ok, so it's more complicated than I thought. Now I also see that output isn't the same for python 3.8-3.10.
I started looking into it because I had the same problem as in #899 and I wondered why one is assigned to the other.

hpkfft: Some advocate for dropping support for older versions more aggressively.
https://scientific-python.org/specs/spec-0000/

wjakob:
The behavior in nanobind is intentional. It smooths over inconsistencies and provided a smoother update path from older nanobind versions where `enum_<>` was realized using `class_<>` bindings. If this bothers you, it should be easy to write a wrapper that restores the original implementation of these methods.

wojdyr:
Fair enough.
But let me add that it also creates inconsistencies (repr without `< >`) and it's a change when migrating from pybind11.
In pybind11 `repr(m.UnscopedEnum.EOne) == "<UnscopedEnum.EOne: 1>"`.

PR #1010, opened 2025-04-13
Add python script version.py

hpkfft:
It would be nice to have a python script to print out nanobind's version number (using the same logic as found in `CMakeLists.txt`), in particular, using `nanobind.h` as the source of truth and runnable from any directory.
For example:
```
$ ./version.py 
2.6.1
$ cd ../myproject
$ ../nanobind/version.py 
2.6.1
```

To make this more appealing, I added logic to write the version number so you can use it before/after a release:
```
$ ./version.py -w 2.7.0-dev1
$ ./version.py
2.7.0-dev1

$ ./version.py -w 2.7.0
$ ./version.py 
2.7.0

$ git status
	modified:   include/nanobind/nanobind.h
	modified:   pyproject.toml
	modified:   src/__init__.py
```

Note that the version syntax is from https://semver.org/ with nanobind's restriction that the `<pre-release>` after the hyphen (if present) is `devN` with N > 0, i.e., `X.Y.Z` or `X.Y.Z-devN`.  In summary:
```
$ ./version.py -h
Usage:  ./version.py
   or:  ./version.py -w X.Y.Z
   or:  ./version.py -w X.Y.Z-devN
```


wjakob:
This is great, but I prefer not to have it in the main directory. How about `src`? Can it be put into `src` in such a way that this doesn't end up in PyPI packages, etc.?

hpkfft:
Done.   Please see #1014.

hpkfft:
This process of constructing instruction tables should be very fascinating.
There need be no real danger of it ever becoming a drudge,
for any processes that are quite mechanical
may be turned over to the machine itself.

— ALAN M. TURING (1945)

PR #1012, opened 2025-04-14
Support easier construction of nb::typed forms

tjstum:
Currently, if you have a function where you want to specify the types manually, you have to write it sort of like this:

```c++
T someFunction();

m.def("f", []() -> nb::typed<nb::object, T> {
    return nb::borrow<nb::typed<nb::object, T>>(nb::cast(someFunction()));
});
```

This example is obviously contrived, but it does come up in some circumstances:
- You retrieved an `nb::object` by calling a Python function, and you happen to know the return type (e.g. `decimal.Decimal`)
- You may need to apply different arguments to `nb::cast` based on different values of a type, so a bound lambda is constrained to return `nb::object`, but we still know a proper type

Before this change, the test cases in would fail to build, since there was no conversion from an `nb::object`/`nb::tuple` to an `nb::typed`


docs/changelog.rst
```diff
@@ -54,6 +54,10 @@ Version TBD (unreleased)
   with unsorted indices. (PR `#981
   <https://github.com/wjakob/nanobind/pull/981>`__).
 
+- Functions that return ``nb::typed<nb::object, ...>`` now support copy and move
+  construction, making it more ergonomic to return that type. (PR `#1012
```
oremanj, line 58:
```suggestion
- ``nb::typed<T, ...>`` now supports construction from ``T``, making it more
  ergonomic to return values with type annotations. (PR `#1012
```

wjakob:
Thanks, this is great!

PR #1014, opened 2025-04-15
Add python script src/version.py

hpkfft:
This is a continuation of #1010, which is non-reopenable because I force pushed to avoid a merge commit.

FYI, my motivation is improved Meson support.  Meson's [WrapDB](https://mesonbuild.com/Using-the-WrapDB.html) adds an overlay (or patch) on top of the nanobind source code to provide the file [meson.build](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/nanobind/meson.build).  This file currently contains the nanobind version number:
```
project( 'nanobind',   'cpp',
  version: '2.4.0',
  meson_version: '>=0.53.0',
  license: 'BSD-3-Clause',
)
```
After the next release of nanobind (and there's no hurry as far as I'm concerned) and with this PR, this can become
```
project( 'nanobind',   'cpp',
  version: run_command('src/version.py', check: true).stdout().strip(),
  meson_version: '>=1.0.0',
  license: 'BSD-3-Clause',
)
```
which makes is easier to keep the WrapDB up-to-date with nanobind releases (only the [wrap file](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/nanobind.wrap) and the [releases.json](https://github.com/mesonbuild/wrapdb/blob/master/releases.json#L2648C3-L2659C5) index need to be updated).
Also, the improved `meson.build` can be used with a git clone of the nanobind master branch.

### DIFF
--- a/src/version.py
+++ b/src/version.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+
+# With no command line flag, this prints the nanobind version.
+# With flags -w semver, this writes the new version to where it's needed.
+
+import os
+import re
+import sys
+
+# Parse the header file <nanobind/nanobind.h> and print the version.
+def get_version(root):
+    major = ''
+    minor = ''
+    patch = ''
+    dev = ''
+    with open(os.path.join(root, "include/nanobind/nanobind.h"), 'r') as f:
+        for line in f:
+            if m := re.match(r'#define NB_VERSION_(.+)', line):
+                if m_major := re.match(r'MAJOR\s+([0-9]+)', m.group(1)):
+                    major = m_major.group(1)
+                if m_minor := re.match(r'MINOR\s+([0-9]+)', m.group(1)):
+                    minor = m_minor.group(1)
+                if m_patch := re.match(r'PATCH\s+([0-9]+)', m.group(1)):
+                    patch = m_patch.group(1)
+                if m_dev := re.match(r'DEV\s+([0-9]+)', m.group(1)):
+                    dev = m_dev.group(1)
+
+    version_core = '.'.join([major, minor, patch])
+    if int(dev) > 0:
+        print(version_core, '-dev', dev, sep='')
+    else:
+        print(version_core)
+
+# Write the semantic version to nanobind.h, pyproject.toml, and __init__.py.
+# The semver string must be either 'X.Y.Z' or 'X.Y.Z-devN', where X, Y, Z are
+# non-negative integers and N is a positive integer.
+def write_version(root, semver):
+    major = 0
+    minor = 0
+    patch = 0
+    dev = 0
+    try:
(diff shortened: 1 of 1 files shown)


wjakob:
This file ends up in wheels built with `pipx run build`. Could I ask you look into a way to avoid this? Thanks!

hpkfft:
I have to step away for the next month or so, but I can follow up after, or perhaps someone else can help sooner.

If the wheels you're making are source distributions of nanobind (which I guess they are if they include the `src` subdirectory), perhaps you will consider allowing this script to be included as well?  The script is less than 4KB, and it is intended to be useful to anyone building from source and using an alternative build system (such as Meson).  It seems awkward to create a situation in which installing nanobind sources from `nanobind-2.7.0.tar.gz` or cloning the github repository works but using pip does not (because the `version.py` script is missing).  I haven't installed nanobind myself using pip, and Meson's WrapDB service doesn't use it, but it seems unkind to make the pip installation less capable.

I haven't tested it, but I am led to believe that the following line in `pyproject.toml` will do as you request:
```
[tool.scikit-build]
sdist.exclude = ["src/version.py"]
```

wjakob:
I actually don't mind having this script in `sdist` tarballs, I think it makes sense to have there as this is supposed to represent the complete source code of a project. But I don't thin it makes sense to have in binary wheels.

hpkfft:
The "binary" wheel is actually source code, so it's a little confusing.   I'm not disagreeing (nor agreeing for that matter).  I'm not an expert in Python nor in its ecosystem and tools.

Anyway, it looks like the following in  `CMakeLists.txt` installs the Python files from `src`:
```
  install(
    DIRECTORY src/
    DESTINATION "${NB_INSTALL_MODULE_DIR}"
    FILES_MATCHING PATTERN "*\.py"
  )
```
and can be changed to either:
```
  install(
    DIRECTORY src/
    DESTINATION "${NB_INSTALL_MODULE_DIR}"
    FILES_MATCHING PATTERN "*\.py"
    PATTERN "version.py" EXCLUDE
  )
```
or
```
  install(
    FILES "src/__init__.py"
          "src/__main__.py"
          "src/stubgen.py"
    DESTINATION "${NB_INSTALL_MODULE_DIR}"
  )
```

hpkfft:
Note that something has to be done, because the existing `install(DIRECTORY......` will move `version.py` from the `src` directory into the top-level directory.
The `version.py` script doesn't work if it is moved there because it expects itself to be in `src`:
```
 root = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
Another option is to move the script into the top-level directory of the git repo (and remove the outermost `os.path.dirname`), where it will be ignored by the `install()` functions in `CMakeLists.txt`.

PR #1015, opened 2025-04-15
Fix bugs in tp_traverse examples and documentation.

hawkinsp:
* In wrapper_tp_traverse in test_classes.cpp, make sure to visit the type.
* In both the tp_traverse docs and tests, do not visit the contents of the C++ object if the instance is not ready. A nanobind object is tracked for GC at the end of PyType_GenericAlloc, but that is before the constructor is called. If a GC occurs between the two, we may end up trying to traverse an unconstructed C++ object. We found this to happen in practice in JAX.
* Update the Wrapper type in the docs to use a std::shared_ptr<Wrapper>. It makes no sense to use nb::find if the type is an nb::object.

### DIFF
--- a/docs/refleaks.rst
+++ b/docs/refleaks.rst
@@ -330,11 +330,26 @@ In particular, *all* types in the cycle must implement the ``tp_traverse``
 *type slot*, and *at least one* of them must implement the ``tp_clear`` type
 slot.
 
-Here is an example of the required code for the prior ``Wrapper`` type:
+Here is an example of the required code for a ``Wrapper`` type:
 
 .. code-block:: cpp
 
+   struct Wrapper { std::shared_ptr<Wrapper> value; };
+
    int wrapper_tp_traverse(PyObject *self, visitproc visit, void *arg) {
+       // On Python 3.9+, we must traverse the implicit dependency
+       // of an object on its associated type object.
+       #if PY_VERSION_HEX >= 0x03090000
+           Py_VISIT(Py_TYPE(self));
+       #endif
+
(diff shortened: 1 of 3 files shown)


hawkinsp:
BTW, looks like your ubuntu 20.04 images have been turned down by github as of today.

"This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101"

wojdyr:
I proposed the `old-compilers` section in ci.yml last year to monitor which oldest compilers are still supported.
Since these compilers are less and less relevant, I guess setting them up in docker would be an overkill and it's better to remove them.

wjakob:
Oh wow, that's concerning. Thank you for finding and reporting this issue. (cc @awni who was using this pattern in a large base).

awni:
For some reason updating to this (ie checking the ready flag) breaks our leak checking tests 🤔 . I'm not aware of any situation in the test where we destroy an object before its fully initialized.  The way we use this is by wrapping `nb::cpp_func` in our own [`mlx_func`](https://github.com/ml-explore/mlx/blob/main/python/src/mlx_func.cpp#L95-L101) and install the type slots for the GC. 

Do you happen to know, is there something we're missing that should set the ready flag to true?

See here: https://github.com/ml-explore/mlx/pull/2079

wjakob:
@awni Your `gc_func` is not a nanobind object. Therefore, calling `nb::inst_ready()` on it is UB.

awni:
Makes sense, thanks!!

PR #1017, opened 2025-04-17
Add `private_exceptions` parameter to stubgen.

cqc-alec:
Adds an option to the stubgen CLI and Python class to allow inclusion of selected private members of a module without including all of them.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -173,6 +173,7 @@ class StubGen:
         recursive: bool = False,
         include_docstrings: bool = True,
         include_private: bool = False,
+        private_exceptions: set[str] = set(),
         include_internal_imports: bool = True,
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @cqc-alec -- This change targets a relatively specific use case that seems somewhat specific to your project. At the same time, it represents a step towards a large space of similar customizations what could significantly increase the size and complexity of the stub generator. Stubgen targets a common set of features and the goal is not to cover everything. My suggestion would be to subclass the `.put()` method of the stub generator in your downstream project. The stub generator can be subclassed specifically to permit this kind of hot-patching in end applications.

cqc-alec:
Hi @wjakob , that is very reasonable. I will do as you suggest. Thanks!

PR #1022, opened 2025-04-21
docs: Add doc on new `nanobind_static_library` macro, cp314 SABI option

nicholasjng:
In particular, notes the experimental nature of the underlying `cc_static_library` rule. The `"cp314"` value for the stable ABI flag was added to the list of legal input values.

Also removes linebreaks in some links to the Bazel docs, which were broken at least in VSCode.


wjakob:
Thanks!

PR #1032, opened 2025-04-30
Bypass python bug that causes compilation targeting stable ABI of python 3.12 not to work on python 3.12 when compiled with python 3.13+

somik861:
To be more specific:
Python 3.12 uses macros defined in abstract.h to rewrite PyObject_DelAttr (and PyObject_DelAttrString) to PyObject_SetAttr (PyObject_SetAttrString).
Python 3.13 defines new symbols PyObject_DelAttr and PyObject_DelAttrString and removes the macro.

This change is however not guarded by PY_LIMITED_API check.
Library compiled on 3.13 will therefore search for PyObject_DelAttr, regardless of PY_LIMITED_API. This causes the library to be unloadble in Python 3.12. 

I added simple macro check, that mimics the macro rewrite of Python 3.12 when limited API for 3.12 is requested.

I have observed this bug on windows and linux.
Fix tested on windows as well as linux.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -463,13 +463,21 @@ void setattr(PyObject *obj, PyObject *key, PyObject *value) {
 }
 
 void delattr(PyObject *obj, const char *key) {
+    #if (defined(Py_LIMITED_API) && PY_LIMITED_API < 0x030d0000)
+    int rv = PyObject_SetAttrString(obj, key, nullptr);
+    #else
     int rv = PyObject_DelAttrString(obj, key);
+    #endif
     if (rv)
         raise_python_error();
 }
 
 void delattr(PyObject *obj, PyObject *key) {
+    #if (defined(Py_LIMITED_API) && PY_LIMITED_API < 0x030d0000)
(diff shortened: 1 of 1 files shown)


wjakob:
Isn't this is a bug in Python then? I'm wondering whether nanobind is the right place to fix it.

somik861:
Hi,

yes it is bug in python itself :D 

Never the less, I don't think python will ever fix it.

This hot-fix will however work even if python developers decides to fix it some day.

As this is a simple change, which does not cause any difference for users that are not using Stable API. And users that will use Stable API will most likely target 3.12+ (as it is the lowest supported), i would still kindly ask merge it into nanobind.

wjakob:
Dear @somik861,

could I ask you to still report this issue on the Python bug tracker? I am very interested in what the Python C API experts will say, which will also influence how nanobind is going to respond to the issue.

wjakob:
Never mind, I reported it myself here: https://github.com/python/cpython/issues/134989

Angelcarrillo11 (approved):

somik861:
Looks like python will fix the problem after all. Thank you

PR #1037, opened 2025-05-04
Add constructors of enum that take member definitions

wenjuno:
The benefit of the new API is a simpler and safer implementation that uses only the public API of the Python Enum class, mimicking how it would be done in Python code. The exisiting API `value()` incrementally adds member to a Enum class, entailing modifying internal states of the Enum class.

### DIFF
--- a/include/nanobind/nanobind.h
+++ b/include/nanobind/nanobind.h
@@ -35,6 +35,7 @@
 #include <typeinfo>
 #include <utility>
 #include <new>
+#include <vector>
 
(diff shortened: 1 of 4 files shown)


wjakob:
For context: the existing API exists to retain compatibility with the previous version of nanobind (and classic pybind11 enums) that constructed enumerations in an incremental manner. Besides compatibility, I much prefer this way of specifying declarations. This API works flawlessly across Python versions and will be adapted if an issue ever arises. I am not in favor of supporting a second API to do the same thing in a different way.

PR #1045, opened 2025-05-15
stubgen: add `LIB_PATH` option to locate dependent shared libraries

Doekin:
This PR addresses issues such as `ModuleNotFoundError` that could occur during stub generation when extension modules depend on external shared libraries (e.g., `.dll` on Windows, `.so` on Linux) located outside of the standard search paths.

### Background

On Windows, when stubgen attempts to import an extension module to generate stubs, `importlib.import_module()` may fail if the Python interpreter cannot locate the required dependent DLLs—for example, if they are not in the same directory as the `.pyd` file.

Example error:
```console
 Generating my_ext.pyi
   File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-snyzxcow\overlay\Lib\site-packages\nanobind\stubgen.py", line 1440, in <module>
     main()
   File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-snyzxcow\overlay\Lib\site-packages\nanobind\stubgen.py", line 1367, in main
     mod_imported = importlib.import_module(mod)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ...
 ModuleNotFoundError: No module named 'my_ext'
```
In contrast, on Linux, the dynamic linker can use the extension module's embedded RPATH to find dependencies.

### Proposed Change

A new `LIB_PATH` argument has been introduced to `nanobind_add_stub`. This option allows you to specify a list of directories containing dependent shared libraries. During stub generation:

- On **Windows**, these directories are added to the DLL search path using `os.add_dll_directory()`.
- On **Linux**, the paths are prepended to the `LD_LIBRARY_PATH` environment variable (on macOS, to `DYLD_LIBRARY_PATH`).

### Example Usage

```cmake
# Suppose my_ext.pyd links against some_dynamic_lib.dll,
# and some_dynamic_lib.dll is in a separate output directory.
target_link_libraries(my_ext PRIVATE some_dynamic_lib)

nanobind_add_stub(
    my_ext_stub
    MODULE my_ext
    OUTPUT my_ext.pyi
    PYTHON_PATH $<TARGET_FILE_DIR:my_ext>
    # Provide the directory containing some_dynamic_lib.dll or .so
    LIB_PATH $<TARGET_FILE_DIR:some_dynamic_lib>
    DEPENDS my_ext
)
```

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -618,7 +618,7 @@ endfunction()
 # ---------------------------------------------------------------------------
 
 function (nanobind_add_stub name)
-  cmake_parse_arguments(PARSE_ARGV 1 ARG "VERBOSE;INCLUDE_PRIVATE;EXCLUDE_DOCSTRINGS;EXCLUDE_VALUES;INSTALL_TIME;RECURSIVE;EXCLUDE_FROM_ALL" "MODULE;COMPONENT;PATTERN_FILE;OUTPUT_PATH" "PYTHON_PATH;DEPENDS;MARKER_FILE;OUTPUT")
+  cmake_parse_arguments(PARSE_ARGV 1 ARG "VERBOSE;INCLUDE_PRIVATE;EXCLUDE_DOCSTRINGS;EXCLUDE_VALUES;INSTALL_TIME;RECURSIVE;EXCLUDE_FROM_ALL" "MODULE;COMPONENT;PATTERN_FILE;OUTPUT_PATH" "PYTHON_PATH;LIB_PATH;DEPENDS;MARKER_FILE;OUTPUT")
 
   if (EXISTS ${NB_DIR}/src/stubgen.py)
     set(NB_STUBGEN "${NB_DIR}/src/stubgen.py")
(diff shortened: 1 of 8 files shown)


wjakob:
I am generally open to this feature, but what bothers me is the windows-specific nature. The two other supported platforms (Linux, macOS) have analogous ways of adding dynamic library search paths.

Doekin:
Thanks for raising that point! I updated the code to handle dependent shared libraries on Linux and macOS as well. The option has been renamed from `DLL_PATH` to `LIB_PATH`, and `LIB_PATH` now prepends the specified paths to `LD_LIBRARY_PATH` (on Linux) or `DYLD_LIBRARY_PATH` (on macOS).

Doekin:
I've rebased the branch on the latest `master` and simplified the change. `LIB_PATH` now keeps the `os.add_dll_directory()` handles alive on Windows, avoids the unreliable in-process `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH` handling, and the docs now recommend rpath for Linux/macOS. I also replaced the previous `inter_module` test with a smaller focused test using a dependent shared library.

izzalDev:
@Doekin, I believe it would be better to implement automated dependency detection. For example, if torch is already specified in `pyproject.toml` or installed in the active virtual environment, the user shouldn't need to manually define the `LIB_PATH` pointing to `$VIRTUAL_ENV/Lib/site-packages/torch/lib`.

Doekin:
Thanks for the suggestion, @izzalDev! Auto-detection would definitely be awesome for UX, but I fear it might be too fragile to implement in a lightweight script like `stubgen.py`:

1. **No standard DLL layout:** Python packages place DLLs everywhere (e.g., `torch` puts them in `/lib`, others in the root). Automatic scanning is highly error-prone.
2. **CMake already knows the exact paths:** Many projects using `nanobind` are standard C++ projects without a pyproject.toml. Since CMake knows the exact location of the linked dependencies (e.g., via `$<TARGET_FILE_DIR:lib>`), passing them explicitly avoids guesswork and "DLL hell".

Keeping `stubgen.py` predictable and explicit seems safer for edge cases.

PR #1055, opened 2025-05-22
Fix stub generation for imported methods

inducer:
Currently, stub generation mishandles the case where a method is actually a function imported from a different module. (It simply imports the name as the name of the method and calls it a day. A name never gets added to the class namespace.)

This can arise, for example, if one adds methods to nanobind-generated classes via monkeypatching.

Full disclosure: this is a bit of a drive-by PR. I did not investigate if/how type stubs are tested. If you'd like me to add testing or make other revisions, I would be happy to.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -388,7 +388,12 @@ class StubGen:
 
         # Check if this function is an alias from *another* module
         if name and fn_module and fn_module != self.module.__name__:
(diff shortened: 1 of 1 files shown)


wjakob:
Could you please add a test that covers this new feature? You will also need to update the stub reference files.

PR #1065, opened 2025-05-30
Loop over dependent targets to detect sanitizers if one of those targets has the flags added

maximiliank:
Fixed sanitizer detection for stub creation in case the sanitizer flags are added via LINK_LIBRARIES


wjakob:
This looks good, but can you add a comment explaining the addition?

maximiliank:
Should I add it in the code or in the PR comments?

The addition is required if the sanitizer flags are not passed directly to the extension target but via another linked target. E.g.

```cmake
target_compile_options(project_options INTERFACE -fsanitize=address,undefined)
target_link_options(project_options INTERFACE -fsanitize=address,undefined)

# define nanobind target my_ext

target_link_libraries(my_ext project_options)
```

In the current implementation it does not detect that the extension is build with sanitizers. The fix implements looping over all dependent targets to search for dependencies on sanitizers.

maximiliank:
Done

Issue #989, opened 2025-04-03
[BUG]: Recursive stubgen creates invalid tree


Ahajha:
### Problem description

When generating stubfiles recursively, the output tree doesn't match the project structure.

Given a submodule structure looking like:
```
mymod
- sub1
- sub2
```
stubgen will create an output tree looking like
```
mymod.pyi
sub1/__init__.pyi
sub2/__init__.pyi
```
This is incorrect mainly because of the name for `mymod.pyi`, in this case it should be `__init__.pyi` (and then everything needs to be placed in a `mymod` folder).

As an additional QOL feature related to this, anytime we see a folder with just a `__init__.pyi` in it, we can collapse to just a file with `<folder>.pyi`.

If this makes sense, I can attempt a PR.

wjakob:
So I'm puzzled because AFAIK nanobind's stubgen handles all of this already.  (For example, Dr.Jit, one of the projects where I am using nanobind and stubgen heavily, contains both of those cases). If this is somehow broken then a PR would be appreciated. I will close this issue for now because it's not really clear what is broken.

Ahajha:
This one I have a public repro that I can push. Will also probably be a few days.

Ahajha:
Okay I'm just getting back around to this issue - I've make a simple repro that doesn't involve cloning anything:
```sh
uv venv # or pick your favorite venv
source .venv/bin/activate # for Unix
# Source is here: https://github.com/mlc-ai/xgrammar
uv pip install nanobind xgrammar==0.1.19
python3 -m nanobind.stubgen -m xgrammar.xgrammar_bindings -O . -r
```
The key note is the `-r` flag that has this incorrect behavior. This creates 4 files:
```
testing/grammar_functor/__init__.pyi
testing/__init__.pyi
kernels/__init__.pyi
xgrammar_bindings.pyi
```
The issue is the layout, as is it isn't in the correct shape for installation. I would expect it to be laid out like this:
```
xgrammar_bindings/__init__.pyi
xgrammar_bindings/testing/grammar_functor.pyi
xgrammar_bindings/testing/__init__.pyi
xgrammar_bindings/kernels.pyi
```
At the very least, this would be acceptable:
```
xgrammar_bindings/__init__.pyi
xgrammar_bindings/testing/grammar_functor/__init__.pyi
xgrammar_bindings/testing/__init__.pyi
xgrammar_bindings/kernels/__init__.pyi
```

[...]

PR #1082, opened 2025-06-21
Fix recursive stubgen output structure

Ahajha:
Context here: https://github.com/wjakob/nanobind/issues/989#issuecomment-2979084175, but going to add here too.

I'm using https://github.com/mlc-ai/xgrammar as an example here, but I've run into this with an internal codebase as well. I have a PR there to add stubfile generation (https://github.com/mlc-ai/xgrammar/pull/341/files), and I need to work around this issue by fixing up the output tree after building. What happens is that with the `-r` flag, the output tree looks like:
```
testing/grammar_functor/__init__.pyi
testing/__init__.pyi
kernels/__init__.pyi
xgrammar_bindings.pyi
```
This output tree doesn't match the project structure - I would accept one of the two following outputs as valid:
```
xgrammar_bindings/__init__.pyi
xgrammar_bindings/testing/grammar_functor/__init__.pyi
xgrammar_bindings/testing/__init__.pyi
xgrammar_bindings/kernels/__init__.pyi
```
or
```
xgrammar_bindings/__init__.pyi
xgrammar_bindings/testing/grammar_functor.pyi
xgrammar_bindings/testing/__init__.pyi
xgrammar_bindings/kernels.pyi
```
The first is more verbose, but valid. The second is a simpler form which I would generally prefer.

This makes the output tree look like the second, by generating the first structure and then simplifying the paths.

To be super clear - this is a breaking change. I'm happy to put this behind a flag. Or even two, one for this new structure and one for the tree simplification (though I think the simplification should be the default, and adding a second flag from the getgo is likely premature).

I'm also happy to add tests, just want to make sure that the direction of this PR is okay first before I do that.

Some design notes:
 - This places another folder in the specified output folder, for two reasons. First, if omitting the output directory flag, this correctly uses the name of the module as the subfolder, making the output tree correct next to the bindings file. Second, if using recursive stubgen without submodules (odd but fine to do), without this a bare __init__.pyi would be put in the folder and then tranformed into `foldername.pyi`, which would not be inside the user specified folder.
 - Generally, I think it makes sense to be as consistent as possible between recursive and nonrecursive, and with vs without submodules.
 - I added an assert here as I accidentally broke some stuff locally when I wrote to a folder with other files in it. The check just makes sure we don't modify anything other than stubfiles.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -818,7 +818,7 @@ class StubGen:
 
                         if not module_file or module_file.endswith('__init__.py'):
                             dir_name = self.output_file.parents[0] / value_name_s[-1]
-                            dir_name.mkdir(parents=False, exist_ok=True)
+                            dir_name.mkdir(parents=True, exist_ok=True)
                             output_file = dir_name / '__init__.pyi'
                         else:
                             output_file = self.output_file.parents[0] / (value_name_s[-1] + '.py')
@@ -1399,6 +1399,9 @@ def main(args: Optional[List[str]] = None) -> None:
 
         file.parents[0].mkdir(parents=True, exist_ok=True)
 
+        if opt.recursive:
+            file = file.with_suffix("") / "__init__.pyi"
+
         sg = StubGen(
             module=mod_imported,
             quiet=opt.quiet,
@@ -1433,6 +1436,20 @@ def main(args: Optional[List[str]] = None) -> None:
 
         with open(file, "w", encoding='utf-8') as f:
             f.write(sg.get())
+
+        # Simplify mod/__init__.pyi to mod.pyi if there are no other files in the folder
+        if opt.recursive:
+            for folder in file.parent.glob("**/"):
+                files = list(folder.iterdir())
+                if len(files) == 1:
+                    old_name = files[0]
+                    assert (old_name.suffix == ".pyi"), \
+                        "Tried to rename some files in the output folder that were not .pyi files, is this intended?"
+                    new_name = old_name.parent.with_suffix(".pyi")
+                    if not opt.quiet:
+                        print(f"  - renaming {old_name} to {new_name} ..")
+                    old_name.rename(new_name)
+                    folder.rmdir()
 
     if opt.marker_file:
         if not opt.quiet:


Ahajha:
@wjakob Makes sense, I had considered that too. The issue I wasn't sure how to solve is that we would need to know ahead of time the tree structure, to know if a certain path can be simplified. This method seemed much more obvious how to implement. I think in order to do this, we would need to count how many submodules we generate in the recursive calls, and then condition the path on that. I'll explore that strategy and get back to you.

Ahajha:
@wjakob I updated with the new logic - it ended up being a bit simpler than post-patching.

Ahajha:
I fixed the Eigen issue here: https://github.com/wjakob/nanobind/pull/1088

Ahajha:
bump

keltonhalbert:
+1 for this here too!

Ahajha:
Very interesting, I'll try to repro/fix that ~soonish.

RE: original source directory, is that information available at stubgen-time? I would assume we don't have that information. Unless you mean output directory?

Ahajha:
Actually, wait, it might. Let me double check, there's a related change I need to make.

wjakob:
In my project, the logic on whether to generate ``submodule.pyi`` or ``submodule/__init__.pyi`` works well now. If you still have problems, could I ask you to make a tiny demo repository or push a testcase that demonstrates the problem?

Ahajha:
Okay, it's definitely better but there's one more issue. The submodule naming is correct now, but the top-level stubfile is still incorrectly named. Using `xgrammar` (I used this as a repro in the linked issue), it would still give me
```
testing/grammar_functor.pyi
testing/__init__.pyi
kernels.pyi
xgrammar_bindings.pyi
```
instead of
```
testing/grammar_functor.pyi
testing/__init__.pyi
kernels.pyi
__init__.pyi # <- this one is the broken one
```
I can push a test case later today

wjakob:
I think I understand. In the case of a file that is entirely declared in nanobind (no .py files at all), things go wrong at the first level. Can you see if the following patch fixes this (untested)
[out.patch](https://github.com/user-attachments/files/22105273/out.patch)

Ahajha:
That would make sense, that is indeed the case for us. I'll try that patch in a bit.

Ahajha:
Just tested out the patch, it works!

wjakob:
Awesome, thanks!

PR #1092, opened 2025-06-26
Remove space from user defined string literal

hpkfft:
The Intel `icpx` 2025.2.0 compiler, which is based on clang, warns that identifier '_s' preceded by whitespace in a literal operator declaration is deprecated.

That seems to be correct.  https://en.cppreference.com/w/cpp/language/user_literal.html documents
> `user-defined-string-literal`  the character sequence "" followed, without a space, by the character sequence that becomes the `ud-suffix`

Note that nanobind's definition of `operator""_a` in `nb_attr.h` does not have whitespace, so this PR is consistent.

### DIFF
--- a/include/nanobind/nb_types.h
+++ b/include/nanobind/nb_types.h
@@ -452,7 +452,7 @@ class bytes : public object {
 };
 
 NAMESPACE_BEGIN(literals)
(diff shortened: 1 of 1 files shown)


wjakob:
Thanks!

PR #1097, opened 2025-07-03
Add casting route from float to int

venkataram-nv:
In the current implementation, `cast` for scalar Python types is able to convert from `float` to `int` but not the other way around. These changes add a route to convert floating types to integer types to make the casting process a bit more symmetric. On the whole, this would make scalar conversions more forgiving.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1115,15 +1115,24 @@ NB_INLINE bool load_int(PyObject *o, uint32_t flags, T *out) noexcept {
         return true;
     }
 
-    if constexpr (Recurse) {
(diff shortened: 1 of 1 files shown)


hpkfft:
I strongly recommend this PR be rejected.   The current implementation, which is able to convert from `int` to `float` but not the other way around, is preferable and prevents bugs in user extensions.   Also, the current behavior is consistent with Python:
```
>>> x = 3.14
>>> s = 'hello'
>>> s[x]
TypeError: string indices must be integers, not 'float'
```

venkataram-nv:
I see, that makes sense. Thanks for quick response!

hpkfft:
Thanks!  (My own extension relies on the current behavior.)

PR #1107, opened 2025-07-17
Opaque enums are ok

riclarsson:
This makes it OK again to have enums be created from names.

The steps that are perhaps not immediately obvious as follows:

1.  There exist a ``nanobind::detail::is_opaque`` struct whose value is ``false`` for all variadic types.
2. When ``NB_MAKE_OPAQUE`` is called, the value of ``nanobind::detail::is_opaque`` becomes ``true``.
3. When that static assertion inside ``implicitly_convertible`` would now normally trigger for ``enum``-types, the opacity is checked instead to allow the type to pass through.

The whole thing is tested in one of the enum-files, I think the diff-log to this PR showcase how to use it better than repeating things here.

We have had to held our project to version 2.4 for a long time now since that's when the ``implicitly_convertible``-assertion was added.  I am not sure what type of foot-gun I am yielding here, but allowing conversions from python ``str`` to C++ ``enum`` is incredibly practical to pass hard flags around with little typing.

I have one worry about this commit: ``NB_MAKE_OPAQUE`` takes a variadic set of arguments.  This ``__VA_ARGS__`` are all I have to work with.  I would quite honestly prefer if the macro was changed to be as [documented](https://nanobind.readthedocs.io/en/latest/api_core.html#c.NB_MAKE_OPAQUE) instead.  I am not sure if I am breaking someone's code because of this.


riclarsson:
First of all, I am very sorry for abusing your CI here.  I do not have access to a Windows machine and did not expect them to disable keywords.

I mentioned this problem to you, @wjakob, some time ago.  To repeat: I want our users to create enum-flags from strings.  We have many flags that are not human-friendly in our code because of C++ language limitations.  Take the ion "NO+", it is appalling to imagine calling it "NOplus", but C++ makes the proper name impossible.  And while it is possible to call it "NO+" in python, it is a bit much to expect people to write stuff like ``getattr(X, "NO+")`` to use it.

The solution?  A python "NO+" is automatically turned into a C++ ``X::NOplus``.  Simple.

Anyways, this all worked very well until that static-assertion was added some releases ago in the implicit conversion call.  You just needed to make the enum-class opaque first.  I want this functionality back.

Are you with me on this premise?

If so, this PR adds one way of doing what I want.  What would you need me to clean-up before you consider accepting it?

PR #1117, opened 2025-07-23
Improve fail_unspecified()'s default error message

hpkfft:
When `NB_COMPACT_ASSERTIONS` is defined and `fail_unspecified()` is called, this changes the printed output from:
```
Critical nanobind error: nanobind: encountered an unrecoverable error condition. Recompile using the 'Debug' or 'RelWithDebInfo' modes to obtain further information about this problem.
```
to
```
Critical nanobind error: encountered an unrecoverable error condition. Recompile using the 'Debug' mode to obtain further information about this problem.
```


wjakob:
Thanks!

PR #1119, opened 2025-07-25
accessor: Overload operators to allow in-place assignments on `m_base`

nicholasjng:
This enables users to write syntactic sugar for in-place ops as

```cpp
nb::object obj = ...;
// increment a (mutable) counter value inline.
obj.attr("count") += nb::int(1);
```

This is useful in bindings and wrappers because it avoids the duplication of `obj.attr()` calls. Templating the operator declaration allows their use in all access implementations.

-----------

Hey! This is a change I used for a small personal project lately. We previously discussed this briefly in https://github.com/wjakob/nanobind/pull/688#discussion_r1736413125,  and I'd be happy about a review. 

If you are interested, I will add more tests on other `Impl` specializations. Thanks!


wjakob:
This looks good. I can't think of any substantially better way of implementing it. Can you increase test coverage? It would also be good to cover the other accessors (e.g. dict key-based or index-based indexing).

wjakob:
Thanks!

PR #1132, opened 2025-08-05
docs: Update example to use "_Float16"

hpkfft:
Since it is recommended that portable code use the `_Float16` type defined by ISO/IEC TS 18661-3:2015, it is probably better to update the example in the nanobind docs accordingly.


wjakob:
(Sorry for closing the PR, I got confused between tabs)

wjakob:
OK, fair enough!

PR #1135, opened 2025-08-08
Enable copying of detail::tuple objects

hpkfft:
Given `nb::detail::tuple<int, int> tpl = {3, 17};` neither `auto [aa, bb] = tpl;` nor `auto t2 = tpl;` currently compiles.
To fix this, this PR conditionally enables the templated constructor so it is less greedy.

Also, this adds a sentence to slice's compute function to describe how to obtain the elements of a `detail::tuple`.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -480,7 +480,6 @@ With reference counting
 
    .. code-block:: cpp
 
-
        PyObject* list = ...;
        Py_ssize_t index = ...;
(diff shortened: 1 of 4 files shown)


wjakob:
Hi @hpkfft,

thank you for the PR. I am wondering: won't this PR interfere with perfect forwarding of the tuple arguments?

I didn't understand why structured binding is broken by the tuple constructor. Doesn't this just call the getter behind the scene?

wjakob:
Beautiful. Thank you!

Issue #1130, opened 2025-08-05
[BUG]: bad_cast when no numpy installed, but default argument needs numpy


thetwom:
### Problem description

This is kind of an exotic issue, not sure whether something can be done here.
Assume I have the following kind of wrapper
~~~
some_class.def("test", [](const Vec &a) { return a;}, "a"_a=Vec(0,0,0));
~~~
Where `Vec` is some fixed vector class which has a typecaster to a ndarray. When I import the resulting module into python without having numpy installed, I get a `bad_cast` error. 

Now it is totally fine to fail, but `bad_cast` is an error which is hard to track down. Maybe it is possible to throw a more meaningful error like "numpy not available"?

### Reproducible example code

```text

```

hpkfft:
An `ndarray` does not necessarily need (or use) NumPy.
If you are specifying `nb::numpy`, you might want to check (in your module) that numpy is imported.
As a personal preference, I'd prefer that nanobind remain as small as is reasonably possible and not check for numpy on every cast.

hpkfft:
Actually, I think I'm not understanding your scenario.  Can you post a complete (minimal) reproducer, both C++ and Python?
For example, if you are casting from a NumPy array, then obviously the Python code must have already successfully loaded NumPy....

hpkfft:
Thank you for posting this example.  I will look at it this weekend.  I think it's interesting, and also I had not considered what would happen with my own extension if somebody uses it without NumPy installed.  I agree that `bad_cast` can be hard to debug, so I too would like to make things nice for my users if I can.

No, I did not have a specific reason for mentioning that link, only that it mentions some possible ways to handle errors, such as `PyErr_Warn()`, which I thought was interesting.

wjakob:
Fixed by #1137

PR #1137, opened 2025-08-11
Preserve earlier RuntimeError in from_cpp()

hpkfft:
Currently, if a type caster sets a Python error in `from_cpp`, it is replaced by `cast_error()`.
This PR avoids calling `raise_cast_error()` if a RuntimeError has already occurred.

For example, if the NumPy module cannot be imported by the `ndarray` type caster, the following:
```
>>> z = hpk.empty(8, "numpy.float32")
RuntimeError: std::bad_cast
```
with this PR, becomes:
```
>>> z = hpk.empty(8, "numpy.float32")
RuntimeError: could not export nb::ndarray: ModuleNotFoundError: No module named 'numpy'
```

Note that this is an error on the part of the final user (his not having installed the NumPy module) rather than an error on the part of the nanobind extension's developer.  So, clear messages are even more than usually desirable.

Careful review and comments would be appreciated.
For example, is it good to check that the earlier error matches `PyExc_RuntimeError`, or should any earlier error take priority?


hpkfft:
Thanks for getting this into the next release!
Note that this PR fixes #1130

thetwom:
Thanks a lot for your work and taking this issue into account!

PR #1140, opened 2025-08-17
[WIP] Interoperability with other Python binding frameworks

oremanj:
See also https://github.com/pybind/pybind11/pull/5800, the same feature for pybind11.

[pymetabind](https://github.com/hudson-trading/pymetabind) is a proposed standard for Python <-> C-ish binding frameworks to be able to find and work with each other's types. For example, assuming versions of both nanobind and pybind11 that have adopted this standard, it would allow a nanobind-bound function to accept a parameter whose type is bound using pybind11, or to return such a type, or vice versa. Interoperability between different ABI versions or different domains of the same framework is supported under the same terms as interoperability between different frameworks. Compared to pybind11's `_pybind11_conduit_v1_` API, this one also supports implicit conversions and to-Python conversions, and should have significantly less overhead.

The essence of this technique has been in use in production by my employer for a couple of years now to enable a large amount of pybind11 binding code to be ported to nanobind one compilation unit at a time. Almost everything that works natively works across framework boundaries too, at only a minor performance cost. Inheritance relationships and relinquishment (from-Python conversion of `unique_ptr<T>`) don't work cross-framework, but those are the only limitations I'm aware of.

This PR adds nanobind support for exposing nanobind types to pymetabind for other frameworks to use ("exporting") and using other frameworks' types that they have exposed to pymetabind ("importing"). Types bound by a different framework than the extension module's own nanobind domain are called "foreign". There are some internals changes to allow foreign types to be represented in the same type maps as native nanobind types; this also includes an updated version of the per-thread fast c2p map that allows safe removal of types (since we can make our own types immortal but we can't force everyone else to make their types immortal). It is possible to compile nanobind without the code to support interop, using the new cmake option `NO_INTEROP`.

Current status: nominally code complete and existing tests pass, but I haven't added interop-specific tests or public-facing docs yet.

Performance: I have not yet measured the performance impact of this change, but I expect it to be quite low in situations where the foreign bindings don't need to be used. The new type_c2p_fast caches negative lookups, and we note whether any foreign bindings exist for a C++ type at the same time as we look up the nanobind type for it. If any foreign bindings have been imported, we do need to look up in type_c2p_fast before failing in some cases where we previously could avoid a lookup completely. When the foreign bindings do need to be used to perform a cast, they require a second c2p_fast lookup and some likely-modest indirection overhead.

Memory cost: Exporting a type allocates a 56-byte structure, a capsule object to wrap it, and adds that capsule object to the type's dictionary. Importing a type adds a new entry to the type_c2p_slow map.

Code size: With `NO_INTEROP`, `size libnanobind.a` adds up to 8533 bytes *smaller* than baseline on my machine (an arm64 mac), probably due to reusing `nb_ptr_map` for the type_c2p_fast map. Without `NO_INTEROP` libnanobind.a is 8983 bytes larger than baseline.

Things that need to happen before this can be released:
[x] add user-facing documentation
[x] add unit tests
[ ] test correctness of nanobind/pybind11 interop
[ ] test performance
[ ] solicit feedback from maintainers of other binding libraries
[ ] release pymetabind v1.0, incorporating said feedback


wjakob:
I am so excited! Thank you for your hard work on this @oremanj. I will do a thorough review in the coming week.

Two quick questions just based on the summary: do you plan to also make such a PR for pybind11? Would the idea there be to remove the "conduit" feature and replace it with pymetabind?

oremanj:
> Do you plan to also make such a PR for pybind11?

Yes; I've started on it already. It's more awkward and less zero-cost than this one due to the presumed need to avoid a pybind11 ABI version bump, but I haven't hit any blockers yet. (An additional type map lookup will be needed on every failed load if there are any imported foreign bindings, rather than knowing whether the particular type of interest has foreign bindings.)

> Would the idea there be to remove the "conduit" feature and replace it with pymetabind?

I don't know if that would be palatable, since the conduit feature has already been released and might need to be supported for a long time. Unfortunately I missed the window to be included in pybind11's most recent ABI break, which occurred with the 3.0 release in July; I'm guessing the next one might not be for a long time after that. The two features don't clash, though of course there's some cost to doing the attribute lookup for the conduit method when it's not needed.

src/nb_foreign.cpp
```diff
@@ -0,0 +1,541 @@
...
+                                                            PyObject *obj),
+                                    void *keep_referenced_ctx) noexcept {
+    cleanup_list cleanup{nullptr};
+    auto *td = (type_data *) binding->context;
+    void *result = nullptr;
+    bool ok = nb_type_get(td->type, pyobj,
+                          convert ? uint8_t(cast_flags::convert) : 0,
+                          keep_referenced ? &cleanup : nullptr, &result);
+    if (keep_referenced) {
+        for (uint32_t idx = 1; idx < cleanup.size(); ++idx)
```
wjakob, line 53: I am wondering if there are situations where it might be unsafe to deconstruct the temporaries until after the binding call has finished. 🤔  Here, you do this earlier than nanobind would ordinarily do for its own function bindings.

src/nb_foreign.cpp
```diff
@@ -0,0 +1,541 @@
...
+}
+
+static PyObject *nb_foreign_to_python(pymb_binding *binding,
+                                      void *cobj,
+                                      enum pymb_rv_policy rvp_,
+                                      PyObject *parent) noexcept {
+    cleanup_list cleanup{parent};
+    auto *td = (type_data *) binding->context;
+    rv_policy rvp = (rv_policy) rvp_;
+    if (rvp > rv_policy::none)
```
wjakob, line 68: What is the purpose of this?

src/nb_foreign.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+// Add the given `binding` to our type maps so that we can use it to satisfy
+// from- and to-Python requests for the given C++ type
+static void nb_type_import_binding(pymb_binding *binding,
+                                   const std::type_info *cpptype) noexcept {
+    // Caller must hold the internals lock
+    internals->foreign_imported_any = true;
+
+    auto add_to_list = [binding](void *list_head) -> void* {
+        if (!list_head) {
```
wjakob, line 251: can remove braces

src/nb_foreign.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+// Learn to satisfy from- and to-Python requests for `cpptype` using the
+// foreign binding provided by the given `pytype`. If cpptype is nullptr, infer
+// the C++ type by looking at the binding, and require that its ABI match ours.
+// Throws an exception on failure. Caller must hold the internals lock.
+void nb_type_import_impl(PyObject *pytype, const std::type_info *cpptype) {
+    if (!internals->foreign_registry)
+        register_with_pymetabind(internals);
+    pymb_framework* foreign_self = internals->foreign_self;
+    pymb_binding* binding = pymb_get_binding(pytype);
```
wjakob, line 294: So this is a naïve question. Why would we need to look up ``__pymetabind_binding__``? Won't we be notified of new frameworks/bindings using the hooks?

Or is the idea that the metabind feature is enabled lazily, and if we join the party late then that registry is empty to start with? Still then, I am wondering why we can't populate our own tables from the list of types in the registry, without having to touch the ``__pymetabind_binding__`` member.

src/nb_internals.h
```diff
@@ -181,30 +227,182 @@ struct std_typeinfo_eq {
...
+ *                     Any foreign bindings that it has (whether provided by
+ *                     this nanobind domain or provided by other frameworks)
+ *                     are accessible via type_data::foreign_bindings, with the
+ *                     discriminant semantics of the output of nb_get_foreign()
+ *                     (low 0b00 for a single binding or 0b01 for a list).
+ *
+ * - 0b10: pymb_binding*: This C++ type is not bound in the current nanobind
+ *                        domain but is bound by a single other framework.
+ *
+ * - 0b11: nb_foreign_seq*: This C++ type is not bound in the current nanobind
```
wjakob, line 249: Naïve question: what is the purpose of mapping a type to multiple frameworks?

Previously, from-python or to-python conversions might not work out, because a type is not registered at all. With pymetabind, there is now a way out because we can use the other framework to do the conversion for us. If multiple frameworks bind the same type, then this adds complexity. (e.g. the alloca() / complex locking code path in nb_foreign.cpp). I am wondering if we can end up with a simpler solution when scrapping this.

src/nb_internals.h
```diff
@@ -181,30 +227,182 @@ struct std_typeinfo_eq {
...
+ *                          domain but is bound by one or more other frameworks.
+ *                          (Once we have a list, we don't switch back to a
+ *                          singleton; see nb_type_try_foreign() for why.)
+ *
+ * The `slow` map is the source of truth. Its lookups perform string comparisons
```
wjakob, line 254: This is great. I should have added this.

src/nb_internals.h
```diff
@@ -149,19 +150,64 @@ template <typename T> class py_allocator {
     void deallocate(T *p, size_type /*n*/) noexcept { PyMem_Free(p); }
 };
 
-// Linked list of instances with the same pointer address. Usually just 1.
-struct nb_inst_seq {
-    PyObject *inst;
-    nb_inst_seq *next;
+/// nanobind maintains several maps where there is usually a single entry,
```
wjakob, line 153: This is great. I should have added this.

wjakob (commented):
Dear @oremanj,

this is a very impressive piece of work. I did a first pass over the PR, please see my feedback and questions attached.

Besides these, here are three additional high level questions:

This PR lacks tests. I am sure you have them on your end as part of the development effort. What would be to test this feature in practice (via CI) so that we can ensure it runs and keeps on running. Would it make sense to have a separate test repository (to avoid duplication) that gets pulled into the CI matrices of both nanobind and pybind11 so that any breaking changes of either project can be caught before shipping a new revision?

Are there features of nanobind that are not supported by pymetabind? Any caveats?

Is there anything to watch out regarding leak tracking when multiple frameworks are involved? I noticed that pymetabind's `leak_safe` flag is not used in the actual implementation.

Thanks,
Wenzel

src/nb_foreign.cpp
```diff
@@ -0,0 +1,541 @@
+/*
```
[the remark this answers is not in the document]
hpkfft: Git submodule contents are not included in `nanobind-2.8.0.tar.gz`, at least not for `robin-map`.
If pymetabind is not included in `nanobind-2.9.0.tar.gz`, would you kindly add it to the [Meson Wrap Database](https://github.com/mesonbuild/wrapdb/) as done here for [robin-map](https://github.com/mesonbuild/wrapdb/blob/master/subprojects/robin-map.wrap)?
Though for simplicity, I would prefer the file be copied into the nanobind source directory as is currently done.

cmake/nanobind-config.cmake
```diff
@@ -330,7 +335,7 @@ endfunction()
 
 function(nanobind_add_module name)
   cmake_parse_arguments(PARSE_ARGV 1 ARG
-    "STABLE_ABI;FREE_THREADED;NB_STATIC;NB_SHARED;PROTECT_STACK;LTO;NOMINSIZE;NOSTRIP;MUSL_DYNAMIC_LIBCPP;NB_SUPPRESS_WARNINGS"
+    "STABLE_ABI;FREE_THREADED;NB_STATIC;NB_SHARED;PROTECT_STACK;LTO;NOMINSIZE;NOSTRIP;MUSL_DYNAMIC_LIBCPP;NB_SUPPRESS_WARNINGS;NO_INTEROP"
```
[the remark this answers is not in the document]
hpkfft: If the feature becomes opt-in, would you reverse the polarity of the macro as well?  In other words, `NB_DISABLE_FOREIGN` becomes `NB_ENABLE_FOREIGN`.
Obviously, other build systems do not use `nanobind-config.cmake`.  By default, any macros you add would not be defined.  Developers would opt-in by defining the new macro.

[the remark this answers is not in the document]
Ahajha: Just chiming in for another vote for opt-in. I imagine that most projects don't need to pay the cost (as the bindings will be self contained), and the ones that do would probably just use it to use as a transition period and then turn it off again.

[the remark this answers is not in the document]
oremanj: The authors of a particular extension module don't generally know when they build it whether anyone will want to use its types from a different framework (or a different ABI version of the same framework). I think this is what pybind11 was referring to in their rationale for adding the `_pybind11_conduit_v1_` methods unconditionally -- "to avoid "oh, too late!" situations" (https://github.com/pybind/pybind11/pull/5296). I'm happy to switch the default, but I wonder if we might want to leave this question open until we have a better quantification of the cost? Speaking of which, @wjakob if you still have a copy of the benchmark that you used to obtain the performance comparison numbers in the nanobind docs, I think that might be useful here.

include/nanobind/stl/unique_ptr.h
```diff
@@ -94,6 +94,11 @@ struct type_caster<std::unique_ptr<T, Deleter>> {
         // Stash source python object
         src = src_;
 
+        // Don't accept foreign types; they can't relinquish ownership
```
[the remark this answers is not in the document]
oremanj: I only put the new #ifdefs in libnanobind, because I wanted to avoid "infecting" every piece of client code with a new flag dependency. One way to avoid the extra `inst_check` overhead without adding an #ifdef here would be to add a new cast flag that disables use of foreign types; how would you feel about that?

include/nanobind/nb_class.h
```diff
@@ -114,7 +111,7 @@ struct type_data {
     const char *name;
     const std::type_info *type;
     PyTypeObject *type_py;
-    nb_alias_chain *alias_chain;
+    void *foreign_bindings;
```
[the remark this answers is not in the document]
oremanj: The interop-disabled flag doesn't change the ABI version string, so we can't conditionally include fields based on its presence. Will add a comment.

The original `alias_chain` functionality is now served by the `types_in_c2p_fast` map in `nb_internals`, so that we can track aliases for both our types and foreign types.

oremanj:
Thank you for the thorough review! I responded to some of the inlines; still working on the others.

> This PR lacks tests. I am sure you have them on your end as part of the development effort. What would be to test this feature in practice (via CI) so that we can ensure it runs and keeps on running. Would it make sense to have a separate test repository (to avoid duplication) that gets pulled into the CI matrices of both nanobind and pybind11 so that any breaking changes of either project can be caught before shipping a new revision?

I haven't had a chance to write self-contained tests yet (most of my testing has been in the context of a large production system that uses both binding frameworks, which is useful but not really exportable), but should be able to get to that this week. I was planning to go the route that pybind11 used for the tests of their conduit feature, where the repository test suite contains a separate small extension module that demonstrates/exercises the API. I think there is some value in having both pybind11 and nanobind have self-contained tests that the functionality they advertise works, without either having any dependence on the other. Since pybind11/nanobind interop is the "headline feature", though, it probably also makes sense to have some specific tests of that. Maybe those should even live in the pymetabind repository so that they can be deduplicated and used by both clients.

> Are there features of nanobind that are not supported by pymetabind? Any caveats?

pymetabind doesn't expose the operation of relinquishing ownership from Python to C++ by passing a pyobject to a C++ function that takes a unique_ptr. That is the only missing piece I'm aware of from nanobind's perspective. Now that pybind11 also supports this, it might make sense to allow cross-framework relinquishment; curious for your thoughts there.

> Is there anything to watch out regarding leak tracking when multiple frameworks are involved? I noticed that pymetabind's leak_safe flag is not used in the actual implementation.

My plan was for nanobind to suppress its own leak warnings if any other framework sets leak_safe to false. My practical experience was that nanobind would otherwise issue lots of warnings once any nanobind default arguments had pybind11-bound types. Looks like I initially wrote this in terms of `!bindings_usable_forever` (which is required by `leak_safe` but not implied by it) and failed to update when I added `leak_safe`; thanks for the catch!

src/pymetabind.h
```diff
@@ -0,0 +1,719 @@
...
+    // framework. Return a pointer to the object, or NULL if no pointer of the
+    // appropriate type could be extracted.
+    //
+    // If `convert` is nonzero, be more willing to perform implicit conversions
+    // to make the cast succeed; the intent is that one could perform overload
+    // resolution by doing a first pass with convert=false to find an exact
+    // match, and then a second with convert=true to find an approximate match
+    // if there's no exact match.
+    //
+    // If `keep_referenced` is not NULL, then `from_python` may make calls to
```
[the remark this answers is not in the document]
oremanj: Added:
```
    // The pointer returned by
    // `from_python` may be invalidated once the `keep_referenced` references
    // are dropped. If you're converting a function argument, you should keep
    // any `keep_referenced` references alive until the function returns.
    // If you're converting for some other purpose, you probably want to make
    // a copy of the object to which `from_python`'s return value points before
    // you drop the references.
```

src/pymetabind.h
```diff
@@ -0,0 +1,719 @@
...
+    // or -1 and sets the Python error indicator on error.
+    //
+    // No synchronization is required to call this method.
+    int (*keep_alive)(PyObject* nurse, void* payload, void (*cb)(void*));
+
+    // Attempt to translate a C++ exception known to this framework to Python.
+    // This should translate only framework-specific exceptions or user-defined
+    // exceptions that were registered with the framework, not generic
+    // ones such as `std::exception`. If successful, return normally with the
+    // Python error indicator set; otherwise, reraise the provided exception.
```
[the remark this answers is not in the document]
oremanj: I'm going to make that change for future-proofing reasons: it will be more amenable to taking advantage of the efficiency improvements available from `std::exception_ptr_cast`, which was recently voted into C++26.

src/pymetabind.h
```diff
@@ -0,0 +1,719 @@
...
+
+    // The native identifier for this type in `framework->abi_lang`, if that is
+    // a concept that exists in that language. See the documentation of
+    // `enum pymb_abi_lang` for specific per-language semantics.
+    const void* native_type;
+
+    // The way that this type would be written in `framework->abi_lang` source
+    // code, as a NUL-terminated byte string without struct/class/enum words.
+    // Examples: "Foo", "Bar::Baz", "std::vector<int, std::allocator<int> >"
+    const char* source_name;
```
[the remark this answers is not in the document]
oremanj: Added: "If `framework->abi_lang` uses name mangling, this is the demangled, human-readable name. C++ users should note that the result of `typeid(x).name()` will need platform-specific alteration to produce one."

wjakob:
Hi Jason,

this sounds great! Just one quick thought about immortalization. The last time I looked into this, it seemed to me that deferred reference counting was a feature that is mainly usable by the Python bytecode interpreter. C++ binding code that increases/decreases reference counts of type objects does not benefit and would still access the global counter. It's possible that this changed in the meantime, or that I am simply confused.

The potential pitfall of of reference counting contention on a type object is quite severe, and that problem just goes away when making them immortal. That was the rationale for the current design.

oremanj:
@wjakob To offer a status update here, the pybind11 version of this PR has gone through some fairly extensive offline review and I will be making the relatively minor changes requested by that review in the coming week or two. Once that's reached a state of ready-to-merge-in-principle, I'll come back here and work on this one to get to a state that you're happy with.

I am currently planning to make the feature in nanobind be gated behind a build flag, but enabled by default at runtime if building in that mode. pybind11 is likely to take the feature as default-on due to the different tradeoffs they take around build size/cost vs "it just works".

There is an open question of whether CPython will accept responsibility for managing the bindings registry, rather than needing each project that uses it to carry a copy of pymetabind.h. I will consult with some CPython core devs that have expressed interest in pymetabind to see if this seems plausible to them. If it does, there will likely be a PEP-sized delay of a few months between this PR being ready-in-principle and it being able to be included in a nanobind release. (The PEP process would likely result in API changes to the functions and structures currently exported by pymetabind; pymetabind would want to accommodate those so it can be used as a backport library with low friction; it can't really do that if it's already committed to an ABI that's used in released code.)

cmake/nanobind-config.cmake
```diff
@@ -194,6 +190,7 @@ function (nanobind_build_library TARGET_NAME)
     ${NB_DIR}/src/nb_static_property.cpp
     ${NB_DIR}/src/nb_ft.h
     ${NB_DIR}/src/nb_ft.cpp
+    ${NB_DIR}/src/nb_foreign.cpp
```
wjakob: If this feature is officially called out as "metabind" or "pymetabind", should the file be named `nb_metabind.cpp` (for the nanobind specific metabind integration)

src/nb_type.cpp
```diff
@@ -847,6 +988,7 @@ static PyTypeObject *nb_type_tp(size_t supplement) noexcept {
 
         PyType_Slot slots[] = {
             { Py_tp_base, &PyType_Type },
+            { Py_tp_finalize, (void *) nb_type_finalize },
```
wjakob: I was curious about the split into finalization and deallocation. Why was this needed?

src/nb_type.cpp
```diff
@@ -336,101 +335,234 @@ static void inst_dealloc(PyObject *self) {
...
+                       const std::type_info *type,
+                       bool *has_foreign) {
 #if defined(NB_FREE_THREADED)
-    thread_local nb_type_map_fast type_c2p_fast;
+    auto per_thread_guard = nb_type_lock_c2p_fast(internals_);
+    nb_type_map_fast &type_c2p_fast = *per_thread_guard;
 #else
     nb_type_map_fast &type_c2p_fast = internals_->type_c2p_fast;
 #endif
+    void *const sentinel = (void *) (uintptr_t) 1;
```
wjakob: This sentinel construction is worthy of a comment, it seems more complicated than what we had here before.

src/nb_type.cpp
```diff
@@ -336,101 +335,234 @@ static void inst_dealloc(PyObject *self) {
...
-    // 'type_c2p_fast' TLS. This data structure is eventually deallocated
-    // when the thread terminates.
-    //
-    // In principle, this is dangerous because the user could delete a type
-    // binding from a module at runtime, causing the associated
-    // Python type object to be freed. If a function then attempts to return
-    // a value with such a de-registered type, nanobind should raise an
-    // exception, which requires knowing that the entry in 'type_c2p_fast'
-    // has become invalid in the meantime.
-    //
```
wjakob: Given the removal of this paragraph, does it mean that the new version is more robust wrt. (potentially non-immortal) type objects in free-threaded Python extensions in the future?

src/error.cpp
```diff
@@ -217,12 +217,28 @@ builtin_exception::~builtin_exception() { }
 
 NAMESPACE_BEGIN(detail)
 
-void register_exception_translator(exception_translator t, void *payload) {
-    nb_translator_seq *cur  = &internals->translators,
-                      *next = new nb_translator_seq(*cur);
-    cur->next = next;
-    cur->payload = payload;
-    cur->translator = t;
+void register_exception_translator(exception_translator t,
```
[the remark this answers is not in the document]
wjakob: That makes sense.

wjakob (commented):
Hi @oremanj,

sorry for taking a long time to get back to you. First, a few comments on your questions:

> How should this feature be referred to in public-facing documentation? Ralf pointed out that "interop" is confusingly generic, especially given pybind11's tagline of "Seamless interoperability between C++ and Python". Do we lift the "metabind" name into the public API?

I think that there would need to be a separate documentation page explaining the role and functionality of metabind. Calling it out by its name in both documentation and API seems perfectly fine with me. I agree that "interop" is potentially confusing.

> How much should be done by default, vs only on user request? If we go with only supporting pymetabind under a special build flag, should compilation with that flag imply that the user wants to export everything by default?

As mentioned before, my request would be that this is an opt-in feature. In this case, exporting everything automatically would be a reasonable choice. Is there any significant downside to this? The time spent running binding/export code has historically not been something that mattered.

Unrelated: even if pymetabind is an opt-in feature, I am fine with an ABI break of the regular nanobind in case re-shuffling data structure members make the code simpler for both versions.

> Should nanobind domains that support interop be able to share internals with those that don't, or should interop support be considered part of the ABI tag? Should defaults like nb::interoperate_by_default() apply at the domain level (like the current set_leak_warnings) or at the extension module level?

That's a good question. If it makes things easier, then we could have a separate ABI tag for builds with pymetabind enabled. In that case, it would apply to the entire domain.

Follow-up question: is there such a concept as a domain for pymetabind? (e.g. if a user wants meta-binding between specific sets of extensions while isolating from others). Perhaps this use case is contrived.

> How should multiple bindings for the same type (provided by different extension modules / frameworks) be put in a priority order? The order only matters when doing implicit conversions from-Python or converting to-Python, since from-Python without an implicit conversion will resolve based on the Python object's type. The current behavior is "order of importing" which is easy to implement but potentially makes program semantics dependent on import order in a maybe undesirable way.

There is no right answer to this question. Whatever is easy to implement or more efficient at runtime

> I am currently planning to make the feature in nanobind be gated behind a build flag, but enabled by default at runtime if building in that mode. pybind11 is likely to take the feature as default-on due to the different tradeoffs they take around build size/cost vs "it just works".

Great, this makes sense.

> There is an open question of whether CPython will accept responsibility for managing the bindings registry, rather than needing each project that uses it to carry a copy of pymetabind.h. I will consult with some CPython core devs that have expressed interest in pymetabind to see if this seems plausible to them.

I am curious: were there any developments here?

> How do you feel about the shape of the public API defined here? (import_for_interop, export_for_interop, interoperate_by_default, and the foreign-type options for type and isinstance)

Suppose that pymetabind is an opt-in feature, and that it automatically causes all types to be exported when turned on. That means that some of this API is no longer present, correct? I am assuming we would not need:

- export_for_interop
- interoperate_by_default

Would `import_for_interop` be needed in this case? I was looking at the docstring of this function, and its use case is not clear to me. Under what circumstances would my bindings call it with a type object (without explicit template argument) or with both a type object and a C++ template type parameter?


General observation: the changes to nb_internals.h/cpp, nb_type.h/cpp, etc., are rather complex. They make me worry that I will have a hard time understanding some parts of nanobind the next time I need to dig deeply into its guts. I understand that some eggs will have to be broken here since a minimal integration into interop will require messing with the internal data structures. But I am wondering if there are aspects of the current design that could be simplified.

PR #1153, opened 2025-09-04
Fix stubgen prefixing issue

Ahajha:
Situation: There are two submodules at the same "level", and the name of one of them is a prefix of the name of the other. When stubgen renders the types coming from the longer one in the shorter one, it naively checks that one is a prefix of the other, and ends up stripping away the prefix plus one extra character. I presume this is intended to be a submodule check.

Repro:
```c++
#include "nanobind/nanobind.h"
namespace nb = nanobind;

struct Type {};

NB_MODULE(_core, m) {
  nb::class_<Type>(m.def_submodule("mosh"), "Type");
  m.def_submodule("mo").def("func", [] { return Type{}; });
}
```
Results in 3 files:
`__init__.pyi`
```py
from . import mo as mo
from . import mosh as mosh
```
`mosh.pyi`
```py
class Type:
    pass
```
`mo.pyi`
```py
def func() -> h.Type: ...
```
Note the `h.Type` on the last file, this should be `_core.mosh.Type`.

To fix, we check for an extra `.`, since we're going to be stripping it anyways.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -683,7 +683,7 @@ class StubGen:
             if mod_name == "builtins":
                 # Simplify builtins
                 return cls_name if cls_name != "NoneType" else "None"
-            if full_name.startswith(self.module.__name__):
+            if full_name.startswith(self.module.__name__ + "."):
                 # Strip away the module prefix for local classes
                 return full_name[len(self.module.__name__) + 1 :]
             elif mod_name == "typing" or mod_name == "collections.abc":


wjakob:
Oops, that's terrible. Thank you!

Ahajha:
Hmm, that doesn't work. I think it's because in my specific case, it's actually a nested submodule (my specific case is `max._core.dialects.mo` and `max._core.dialects.mosh`, it's trying to use `.dialects` as a module name in the top level `__init__.pyi`... Which I think should be valid? But regardless the existing change is the most correct of the existing options.

wjakob:
I am not sure I follow -- the idea of this line is that we don't have to import types if they are actually from the same module. If your type is from a nested submodule, it seems to me that this simplification is not correct. Could you give a full example?

Ahajha:
Nudge @wjakob, I can add some tests here if it would help move this forward

tests/test_prefix_module.cpp
Ahajha: Whoops, need to delete this file

wjakob:
Ok. It looks reasonable. Let me merge this for now and see if anyone objects.

PR #1161, opened 2025-09-14
Bump nanobind_bazel version to v2.9.2

nicholasjng:
Same as always, Bazel central registry PR is https://github.com/bazelbuild/bazel-central-registry/pull/5850.


wjakob:
Thanks!

nicholasjng:
That is doable, but I usually take a few days after a nanobind release to push new `nanobind_bazel` versions to the BCR, for different reasons. Until then, the docs would technically contain invalid `MODULE.bazel` code.

But maybe it's not a big deal, since downgrading to latest stable is mostly seamless, and using `git_override()` instead is documented directly below. I'll send you a PR.

PR #1164, opened 2025-09-16
Extend `src/version.py` script to `nanobind_bazel` docs

nicholasjng:
This relieves the need to update the nanobind_bazel documentation entry by hand after each stable release. Technically, there is a brief period(until the subsequent BCR release of nanobind_bazel) during which the documentation contains non-working MODULE.bazel code.

However, this risk is probably acceptable, since I release nanobind_bazel typically in short order after each nanobind release, and the command to override nanobind_bazel from GitHub can be found directly below.

No other changes intended - specifically, no print statements or other logging were added to preserve stdout behavior expected by Meson.

---------------------

As per our latest discussion.

The new script leaves docs unchanged on dev version bumps (i.e. `python src/version.py -w X.Y.Z-dev1`, and updates on stable releases - this is intended because the `MODULE.bazel` snippet I keep sending PRs about is scoped to the latest stable release only.

Happy to hear your thoughts.

### DIFF
--- a/src/version.py
+++ b/src/version.py
@@ -31,7 +31,8 @@ def get_version(root):
     else:
         print(version_core)
 
-# Write the semantic version to nanobind.h, pyproject.toml, and __init__.py.
+# Write the semantic version to nanobind.h, pyproject.toml, __init__.py,
+# and docs/bazel.rst.
 # The semver string must be either 'X.Y.Z' or 'X.Y.Z-devN', where X, Y, Z are
 # non-negative integers and N is a positive integer.
 def write_version(root, semver):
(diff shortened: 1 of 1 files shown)


hpkfft (approved):
LGTM.

Meson does not use the `-w` flag.  It calls `version.py` to obtain the version number of the nanobind installation that it found.  This is printed for informational purposes.
The extension developer can specify a nanobind version, e.g., `>=2.8.0`, and Meson will honor the requirement.

The nanobind docs for Meson do not specify a nanobind version, so no update there is needed when a new version of nanobind is released.

Thanks for offering me the opportunity to review this before it is merged!

--Paul

PR #1168, opened 2025-09-19
Improve ndarray testing

hpkfft:
+ Move JAX tests into the currently empty test_jax.py.  
+ Move TensorFlow tests into the currently empty test_tensorflow.py.
+ Change ret_numpy_half type from `__fp16` to `_Float16`.
+ Enable ret_numpy_half test if `__AVX512FP16__` is defined.

I've starting looking into supporting DLPack's DLManagedTensorVersioned, which has a read-only flag.
That will involve writing more tests, so I thought to do a little cleanup first....

### DIFF
--- a/tests/test_jax.py
+++ b/tests/test_jax.py
@@ -1,0 +1,64 @@
+import test_ndarray_ext as t
+import test_jax_ext as tj
+import pytest
+import warnings
+import importlib
+from common import collect
+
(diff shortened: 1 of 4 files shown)


wjakob:
Perfect, thank you. I am excited about `DLManagedTensorVersioned` :).

PR #1175, opened 2025-09-29
Add support to ndarray for DLPack version 1

hpkfft:
This PR adds support for DLPack version 1 and adds the ndarray framework `nb::arrayapi`, which returns an object that provides the buffer interface and has the two DLPack methods `__dlpack__()` and `__dlpack_device__()`.

Given the following:
```
using array_t = nb::ndarray<float, nb::shape<-1>, nb::c_contig,
                            nb::numpy, nb::device::cpu>;

void set(const array_t& a) {
    const std::size_t n = a.shape(0);
    float* ptr = a.data();
    for (std::size_t i = 0; i < n; ++i) {
        ptr[i] = 1.0f;
    }
}

array_t create(std::size_t n) {
    float* ptr = new float[n];
    nb::capsule deleter(ptr, [](void* p) noexcept { delete[] (float*) p; });
    return array_t(ptr, {n}, std::move(deleter));
}

NB_MODULE(my_extension, m) {
    m.doc() = "nanobind my_extension module";
    m.def("create", &create, "Create a new NumPy array.");
    m.def("set", &set, "Set all elements of NumPy array to one.");
}
```

I measure performance as follows:

|  test   |   old   |   new   |
| ------- | ------- | ------- |
|  set    | 160 ns  |  111 ns |
|  create | 564 ns  |  560 ns |

using Python 3.14 and
```
python -m timeit -n 10000000 -r 10 -s "import numpy as np, my_extension as me; a = np.zeros(8, dtype=np.float32)" "me.set(a)"
python -m timeit -n 1000000 -r 10 -s "import numpy as np, my_extension as me;" "me.create(8)"
```


include/nanobind/ndarray.h
```diff
@@ -86,6 +91,7 @@ NB_FRAMEWORK(tensorflow, 3, "tensorflow.python.framework.ops.EagerTensor");
 NB_FRAMEWORK(jax, 4, "jaxlib.xla_extension.DeviceArray");
 NB_FRAMEWORK(cupy, 5, "cupy.ndarray");
 NB_FRAMEWORK(memview, 6, "memoryview");
+NB_FRAMEWORK(arrayapi, 7, "object");
```
hpkfft, line 94: I'd be happy to change `arrayapi` to something else if you can think of something better.

Also, I'm not sure what to use for the name, so I put `"object"` for now.
Maybe this should be `"ndarray"` and the name for `no_framework` should be `"capsule"`?

src/nb_ndarray.cpp
```diff
@@ -2,37 +2,155 @@
 #include <atomic>
 #include "nb_internals.h"
 
+#define OK_TO_USE_STD_UNIQUE_PTR
```
hpkfft, line 5: For performance, `dlpack_from_buffer_protocol()` no longer returns a PyCapsule.  It was used as an RAII object, but it's expensive because it's allocated on the heap and because `PyCapsule_GetPointer()` does a string compare, etc.
Instead, I'd prefer to use `std::unique_ptr`, which is passed on the stack and is very cheap (maybe even free) to get the pointer.
Note that `<memory>` is not included in the header files used by the extension, only in this file, which is compiled in building `libnanobind-static.a` (or shared, etc.)
If you prefer not to include `<memory>`, a reasonable alternative is to manually call `mt_from_buffer_delete()` before each return statement, as seen in the conditionally compiled code below.

Please advise; then I will remove this macro and update the code below accordingly.

include/nanobind/ndarray.h
hpkfft: In your spare time, please give some thought to changing:
`NB_FRAMEWORK(no_framework, 0, "ndarray")` to `NB_FRAMEWORK(no_framework, 0, "capsule")` and
`NB_FRAMEWORK(arrayapi, 7, "object")` to `NB_FRAMEWORK(arrayapi, 7, "ndarray")`
 or other ideas.
I'm not very familiar with Python typing, and I'm not a good name-commer-upper-wither.   :)
Certainly, I will add documentation before upgrading this draft to a PR.

wjakob: For capsule, I think that the right type would be `types.CapsuleType`.

src/nb_ndarray.cpp
```diff
@@ -765,57 +1020,103 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
...
-                case pytorch::value: pkg_name = "torch.utils.dlpack"; break;
-                case tensorflow::value: pkg_name = "tensorflow.experimental.dlpack"; break;
-                case jax::value: pkg_name = "jax.dlpack"; break;
-                case cupy::value: pkg_name = "cupy"; break;
-                default: pkg_name = nullptr;
-            }
+    if (framework == numpy::value) {
+        try {
+            static PyObject* const array_str = PyUnicode_FromString("array");
+#if PY_VERSION_HEX < 0x03090000
```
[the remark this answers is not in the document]
hpkfft: Yes, using `PyObject_VectorcallMethod()` directly is only done as a performance optimization.  It's faster to customize the call site and use static objects (the pre-made tuple `copy_tpl` for the `kwnames` argument).
I don't see how to improve `operator()` generally since it has to work at any call site.  In other words, it must create a tuple of keyword names at runtime.  I suppose there's a way to do something (similar to JIT compiling), but that's beyond the scope of this PR.
Philosophically, I think it OK to use the low-level Python C-API from within nanobind itself to squeeze that last drop of performance.

docs/ndarray.rst
```diff
@@ -643,26 +657,98 @@ support inter-framework data exchange, custom array types should implement the
...
+by nanobind when specifying :cpp:class:`nb::arrayapi <arrayapi>` as the
+framework template parameter.
+For example:
 
 .. code-block:: cpp
 
-   nb::class_<MyArray>(m, "MyArray")
-      // ...
-      .def("__dlpack__", [](nb::kwargs kwargs) {
-          return nb::ndarray<>( /* ... */);
```
hpkfft: I don't think this works (implementing a Python method as a lambda), or I'm missing something interesting.
I don't see how the `/* ... */` can access the `this` pointer (to get a pointer to the actual data in `MyArray`).
In the new documentation, I added member functions to `MyArray` and used them by name in the binding.

wjakob: You can implement a lambda function that accesses `self`, either by taking a C++ type as first argument, by taking a `nb::handle` as first argument, or by taking `nb::pointer_and_handle<T>` that gives you both.

wjakob:
Is this still a draft PR?

docs/ndarray.rst
```diff
@@ -458,6 +465,13 @@ interpreted as follows:
 - :cpp:enumerator:`rv_policy::move` is unsupported and demoted to
   :cpp:enumerator:`rv_policy::copy`.
 
+Note that when a copy is returned, the copy is made by the framework, not by
+nanobind itself.
```
wjakob: Maybe say: "this is because nanobind does not have a dependence (say, on CUDA) to perform such a copy itself."

hpkfft:
> If I have a project using nb::ndarray, what do I need to benefit from the new interfaces? Can I opt out? What are the implications on compatibility? 

Nothing.  No.  Only goodness.

When nanobind imports a DLPack-capable object, it first tries to call the object's `__dlpack__()` method with keyword argument `"max_version"` set to `(1, 1)`, indicating that nanobind can accept a versioned tensor.  (The minor version is irrelevant.)  The object can return either the old unversioned tensor or a versioned tensor--either way nanobind does the import.  If the object cannot accept the kwarg at all (i.e., raises TypeError), nanobind calls `__dlpack__()` without any kwargs and imports the unversioned tensor.  (In theory, it could be versioned (which would be a bug in their code), but in reality, if the object doesn't even know about `"max_version"`, then it doesn't know about versioned tensors.)
If the object is not DLPack-capable, nanobind tries to import using the buffer protocol.
If that doesn't work, nanobind tries to call `to_dlpack(obj)` on the framework to get an unversioned capsule.  [This is very obsolete, but the code was there, so might as well keep it.]

In the case of a versioned capsule, a flag bit can be set to indicate that the tensor is read-only.   Nanobind honors this and creates a read-only nd-array.
In the case of an unversioned capsule, nanobind assumes it's writable.  As before, it would be the user's responsibility to know if that's not the case and to refrain from actually writing to it.

On export, it depends on the framework.

`no_framework` is unchanged.  It continues to return an unversioned capsule for backward compatibility.

Tensorflow is unchanged.  An unversioned capsule is passed to `tensorflow.experimental.dlpack.from_dlpack()`.  Their online docs show that's the thing to do.

`arrayapi` is new.  It returns an object of type `nanobind.nb_ndarray`, which supports both the buffer protocol and the DLPack `__dlpack__()` and `__dlpack_device__()` methods.  The `__dlpack__()` method accepts and honors the keyword argument `"max_version"` and returns a versioned tensor if and only if the value tuple[int, int] has first component (i.e., major version) of at least 1.  (If the value is None, or the keyword argument is missing, that is equivalent to passing a maximum major version of 0.)

NumPy is unchanged.  It first makes a new `nanobind.nb_ndarray` and then passes it to NumPy, which imports it using the buffer protocol.  I did not see a performance improvement in changing to DLPack.  Also, `numpy.array()` supports a `"copy"` keyword argument, so if a copy is needed, it's done in the same call without having to subsequently call a `copy()` or `clone()` function.

`memview` is unchanged.  It uses the buffer protocol on a new `nanobind.nb_ndarray` object.

PyTorch, JAX, and CuPy: nanobind creates a new  `nanobind.nb_ndarray`  object and then passes that to the framework's `from_dlpack()` function.   That's not different per se, but these frameworks can now call our `__dlpack__()` with a maximum major version of 1 (and any minor version) and get a versioned tensor in return.  They can also pass a maximum major version of 0 and get an unversioned tensor, as before.   Or, pass `max_version=None`, or omit the keyword argument, and get an unversioned tensor, as before.

wjakob:
Beautiful, thank you for this clarification. I guess there could be a performance cost when we try to import a tensor from an older framework that doesn't support versioned capsules (due to calling __dlpack__ multiple times), correct? But I suppose the impact of that should diminish over time.

docs/ndarray.rst
hpkfft: Thanks!  I changed the example to use a lambda, since it's nice to show that these methods can be added in a binding without having to change the C++ class.

docs/api_extra.rst
```diff
@@ -1108,6 +1108,11 @@ convert into an equivalent representation in one of the following frameworks:
 
    Builtin Python ``memoryview`` for CPU-resident data.
 
+.. cpp:class:: arrayapi
```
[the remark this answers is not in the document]
hpkfft: I slightly prefer `arrayapi`, without the underscore, copying the style of `memview`.
In the example code, I think
```
using arrayapi_t = nb::ndarray<nb::arrayapi, double>;
```
looks a bit nicer than
```
using array_api_t = nb::ndarray<nb::array_api, double>;
```
But I'd be happy to change it if you prefer, so don't hesitate to say so.

docs/ndarray.rst
hpkfft: I added a few sentences here.   Let me know if you disagree or want to change something.

hpkfft:
That doesn't seem to be the case.  Using Python 3.11.2 and adding the following to `ndarray_check`:
```
    PyObject* tpfoe = 
      PyUnicode_InternFromString("tensorflow.python.framework.ops.EagerTensor");
    printf("name %s tpfoe\n", (name == tpfoe) ? "is" : "is not");
    int sc = strcmp(tp_name, "tensorflow.python.framework.ops.EagerTensor");
    printf("name %s tpfoe\n", (sc == 0) ? "==" : "!=");
```
I get
```
name is not tpfoe
name == tpfoe
```
when running
```
python3 -m pytest --capture=tee-sys tests/test_tensorflow.py
```

docs/api_extra.rst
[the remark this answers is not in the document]
hpkfft: Done.

docs/ndarray.rst
```diff
@@ -643,26 +665,92 @@ support inter-framework data exchange, custom array types should implement the
...
+    >>> y = np.from_dlpack(ma)
+    >>> y
+    array([0., 1., 2., 3., 4.])
+
+
+The ``kwargs`` argument in the implementation of ``__dlpack__`` above can be
+used to support additional parameters (e.g., to allow the caller to request a
+copy).  Please see the DLPack documentation for details.
+
+The caller may or may not supply the keyword argument ``max_version``.
```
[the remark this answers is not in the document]
hpkfft: You're right; it's better to omit it.  I think an "Internals" subsection is unnecessary because all the info about versioned or unversioned tensors, the fact that `max_version` is optional, etc. is covered in the Python array API standard, the URL to which is now provided.

src/nb_ndarray.cpp
```diff
@@ -765,57 +970,95 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
...
+                return PyMemoryView_FromObject(o.ptr());
+            default:
+                pkg_name = nullptr;
+        }
+        if (pkg_name) {
+#if PY_VERSION_HEX < 0x03090000
+            o = module_::import_(pkg_name)
+                .attr(static_pyobjects[pyobj_name::from_dlpack_str])(o);
+#else
+            PyObject* pkg_mod = module_import(pkg_name);
```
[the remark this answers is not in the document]
hpkfft: Yes, I saw that.  Yes, I think the perf test `create_array_np` in the opening comment to this PR can be made about 100ns faster.  My hypothesis is that making our own PyDict would not be faster than `sys.modules`.  The speedup would be possible by using a C++ cache.  I am thinking this could be done by adding an 8-entry module cache and using the per-module state for storage.  The key could be `void*`, so it could be a C++ `const char*`.  (Literal strings are placed in the `.rodata` section of the ELF shared library.)  So, 128B of storage: 8 key pointers and 8 module `PyObject*`.  Linear search and pointer-equality-compare is probably faster than hashing.

Have to think about locking.  Multiple readers are OK.  Maybe `std::shared_mutex`

I think this should be a separate PR.
It seems like it can/should be a feature of `detail::module_import` and no changes to `nb_ndarray.cpp` would be needed.

Given that Python 3.14 supports subinterpreters in python code (previously, subinterpreters were only available through the C-API), maybe you want nanobind to support it?  The "What's new in Python 3.14" page says, "Regarding extension modules, work is in progress to update some PyPI projects, as well as tools like Cython, pybind11, nanobind, and PyO3."  I mention this because maybe that should be done first so that the design of a module cache can accommodate the restrictions.  Might have to add a pointer to `type_data` to replace `static_pyobjects`....

I haven't given it much thought, but probably my advice/preference would be to announce (in the changelog, etc.) that the upcoming release of nanobind will be the last one to support Python 3.8.
Then, after the release, remove 3.8 support.
Then support subinterpreters.
Then add a C++ cache for imports.

Maybe `nanobind::ndarray_traits<T>` can be removed (deprecated in #742)

PR #1181, opened 2025-10-04
Add test that returns `nb::ndarray<nb::memview>`

hpkfft:
This PR adds a test for returning an array that specifies the `nb::memview` framework.

I had considered changing the label of `nb::no_framework` from `"ndarray"` to `"types.CapsuleType"`, but that would change the signatures of all functions that take an `nb::ndarray<>` as input.  For example, a diff would be:
```
-def inspect_ndarray(arg: NDArray, /) -> None: ...
+def inspect_ndarray(arg: types.CapsuleType[], /) -> None: ...
```


hpkfft:
I marked `test19_return_memview` as `@skip_on_pypy` since `collect()` is not increasing the destruct count.

PR #1182, opened 2025-10-05
Exclude auto-generated enum APIs from the stub

superbobry:
I wasn't sure if I should add a test for this, because none of the current tests seem to run with --include-private.


superbobry:
Ooops, sorry, all done now.

PR #1185, opened 2025-10-09
[stubgen] Added a new --exclude-values flag

superbobry:
The flag forces all values to be rendered as ..., which is usually what you want in a .pyi file.

The motivating use-case in JAX was version attributes, which are currently rendered as

    foo_version: int = 42

so every version bump will be unnecessarily reflected in the .pyis.


superbobry:
@wjakob shall we get these two PRs in? The CI seems happy.

PR #1193, opened 2025-10-20
Allow simple function dispatch if nargs == NB_MAXARGS_SIMPLE

hpkfft:
The complex dispatch loop is needed if `nargs` is strictly greater than `NB_MAXARGS_SIMPLE` (i.e., eight).
This PR allows a function with eight arguments to use simple function dispatch if it is otherwise eligible.


wjakob:
Thanks!

PR #1222, opened 2025-11-23
Enhance FREE_THREADED python detection

cyyever:
This PR enhances free-threaded python detection by querying `Py_GIL_DISABLED` config, which replaces the current behavior of matching `NB_ABI`.
The motivation is that `NB_ABI=cpython-313` on a FreeBSD python 3.13 non-ft build, and it is incorrectly detected as ft version.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -357,13 +357,33 @@ function(nanobind_add_module name)
     set(ARG_STABLE_ABI FALSE)
   endif()
 
-  if (NB_ABI MATCHES "t")
-    # Free-threaded Python interpreters don't support building a nanobind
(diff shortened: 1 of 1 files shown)


wjakob:
This is a very complicated approach. I would prefer if detection of the "t" suffix in NB_ABI could be improved.

hpkfft:
Does the following patch work on FreeBSD:
```
-  if (NB_ABI MATCHES "t")
+  if (NB_ABI MATCHES "[0-9]t")
```
This restricts the match to a `t` that occurs after a digit.

wjakob:
Thanks @cyyever and @hpkfft -- I committed this shorter alternative: acd6a3e31dc83188c1d0f63a6405b19d0eb0cbeb

PR #1224, opened 2025-11-24
Throw TypeError when subclasses forget to call __init__

### LINKED ISSUE #1210: [BUG]: nanobind fails with unhelpful error message when super.__init__ isn't called by overridden constructor
### Problem description

pybind11 does this and it's very useful for reminding users to call `super().__init__()` when inheriting from a bound class. A quick look at nanobind indicates that it's also required and currently nothing performs this check. nanobind doesn't crash (unlike pybind11) but the error is still really difficult to diagnose:

```
>>> f = Foo()
>>> f.getX()
    f.getX()
    ~~~~~~^^
TypeError: getX(): incompatible function arguments. The following argument types are supported:
    1. getX(self) -> int
```

If you're open to bringing it over, I can put together a PR at some point in the near future.

From the original pybind11 issue: https://github.com/pybind/pybind11/issues/2103:

> My users are high school robotics teams, and the RobotPy project is wrapping a C++ library that uses a lot of inheritance. They keep inheriting from C++ objects, add a constructor without calling the super constructor, it crashes, and they don't really know why at first. It would be great if pybind11 could catch this and throw a python exception instead with a useful error message.

pybind11 PR fixing it (but obviously we'd want to use whatever the latest version is): https://github.com/pybind/pybind11/pull/2152

### Reproducible example code

```text
class Foo(SomeNanoBoundClass):
   def __init__(self):
       pass # oops

f = Foo()
f.some_bound_function_that_needs_cpp_instance_data()
```

virtuald:
- Based on https://github.com/pybind/pybind11/pull/2152
- Fixes #1210

### DIFF
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -498,6 +498,33 @@ static int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
     return 0;
 }
 
+/// metaclass `__call__` function that is used to create all nanobind objects
+static PyObject *nb_type_call(PyObject *type, PyObject *args, PyObject *kwargs) {
+
+    // use the default metaclass call to create/initialize the object
+#if defined(Py_LIMITED_API)
+    PyObject *self = ((ternaryfunc)PyType_GetSlot(&PyType_Type, Py_tp_call))(type, args, kwargs);
+#else
+    PyObject *self = PyType_Type.tp_call(type, args, kwargs);
+#endif
+    if (self == nullptr) {
+        return nullptr;
+    }
+
+    // This must be a nanobind instance
+    nb_inst *inst = (nb_inst *) self;
+    if (inst->state == nb_inst::state_uninitialized) {
+        const type_data *t = nb_type_data(Py_TYPE(self));
+        PyErr_Format(PyExc_TypeError,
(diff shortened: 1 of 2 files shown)


src/nb_type.cpp
```diff
@@ -498,6 +498,68 @@ static int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
     return 0;
 }
 
+/// metaclass `__call__` function that is used to create all nanobind objects
+static PyObject *nb_type_call(PyObject *type, PyObject *args, PyObject *kwargs) {
+
+    // use the default metaclass call to create/initialize the object
+#if defined(Py_LIMITED_API)
+    PyObject *self = ((ternaryfunc)PyType_GetSlot(&PyType_Type, Py_tp_call))(type, args, kwargs);
```
virtuald: Caching this would require an ABI bump, but we can do that if you want.

virtuald:
Never mind, my local test environment was stale when I tested it. I'll take a closer look tonight.

virtuald:
Well, I got rid of the MRO stuff so it's a lot simpler now.

I feel like the other performance improvement would be to convert this tp_call to vectorcall, but I haven't been successful at that yet.

PR #1228, opened 2025-11-27
Add bfloat to ndarray_import conversion code

### LINKED ISSUE #1227: [BUG]: ndarray_import silently fails when with convert and bfloat16
### Problem description

I've been using nanobind to handle importing 3rd party library types into ndarray, including dtype conversions and transformations to c_contig. When trying to transform to bfloat16 types, however, a `nullptr` is returned. It does work for other non-bfloat types. I narrowed down the issue to bfloat not being accounted for here:
https://github.com/wjakob/nanobind/blob/dbe8a3c075f628482cfdf0e9fac1d2939794a6ca/src/nb_ndarray.cpp#L697-L713

I'll put together a quick PR.

### Reproducible example code

Cut down example:
```cpp
template <typename... Ts>
auto convert_py_tensor(nb::ndarray<Ts...> py_tensor) {

  // I know, I'm sorry
  nb::detail::ndarray_config config(typename decltype(py_tensor)::Config{});

  // convert to bfloat16
  config.dtype.code = (uint8_t)nb::dlpack::dtype_code::Bfloat;
  config.dtype.bits = 16;

  nb::detail::ndarray_handle* handle = nb::detail::ndarray_import(
        py_tensor.cast().ptr(),
        &config,
        true /*convert*/,
        nullptr /*(cleanup*)*/);

  return nb::ndarray{handle}; // currently returns an empty ndarray
}
```

ThisIsFineTM:
Closes #1227.

`ndarray_import`'s conversion code was missing a case for `bfloat`, causing import conversions to bfloat16 to fail with a `nullptr` and return a `NoneType` back to python. Tested that this behaves as expected in my sandbox and converts to bfloat16 when taking in a `nb::ndarray<nb::array_api>` argument for a `torch` tensor. I'll followup with a test validating the patch soon. Just wanted to get this up before losing it in the holiday rush.

### DIFF
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -704,6 +704,9 @@ ndarray_handle *ndarray_import(PyObject *src, const ndarray_config *c,
                 case (uint8_t) dlpack::dtype_code::Float:
                     prefix = "float";
                     break;
+                case (uint8_t) dlpack::dtype_code::Bfloat:
+                    prefix = "bfloat";
+                    break;
                 case (uint8_t) dlpack::dtype_code::Complex:
                     prefix = "complex";
                     break;


blozano-tt:
Question regarding the preexisting code: Why is the default case returning nullptr? Shouldn’t it throw exception?

wjakob:
Thanks! nullptr is usually used to signal conversion failures.

PR #1249, opened 2025-12-15
Fix func_create to not copy return value when call_guard is used

virtuald:
- Specifying an explicit return value instead of depending on the compiler to autodetect it addresses the problem
- Added a test that shows it (at least, it doesn't compile on macOS clang without this fix)


virtuald:
Good catch, all set.

PR #1251, opened 2025-12-16
Add nb::never_destroy to prevent nanobind from destroying instances

### LINKED ISSUE #1248: Document how to prevent nanobind from ever calling a destructor
If I have a header like this ([real world version](https://github.com/wpilibsuite/allwpilib/blob/1b2f051b4b6d8035ad7a4cf0b85e9aacf26a5441/wpinet/src/main/native/include/wpinet/WebServer.h#L18)):

```c++
#include <memory>

class StaticOnly2 {
  public:
    StaticOnly2(const StaticOnly2&) = delete;
    StaticOnly2& operator=(const StaticOnly2&) = delete;

    int getNumber() { return 1; }

  private:
    StaticOnly2();

    struct SO2Impl;
    std::unique_ptr<SO2Impl> m_impl;

};
```

Nanobind `class_` fails to compile because of errors similar to this:

```
unique_ptr.h:79:19: error: invalid application of 'sizeof' to an incomplete type 'StaticOnly2::SO2Impl'
...
nb_class.h:594:38: note: in instantiation of function template specialization 'nanobind::detail::wrap_destruct<StaticOnly2>' requested here
    594 |                 d.destruct = detail::wrap_destruct<T>;
```

This error can be avoided by specializing the wrap_destruct function in the same compilation unit:

```c++
NAMESPACE_BEGIN(NB_NAMESPACE)
NAMESPACE_BEGIN(detail)

template <> inline void wrap_destruct<StaticOnly2>(void *value) noexcept {
}

NAMESPACE_END(detail)
NAMESPACE_END(NB_NAMESPACE)
```

... but obviously since that's in `nanobind::detail` it's not guaranteed to be stable? I'm happy to just use it and hope it never breaks, but it would be nice to do one of the following:

* Document this and make it a public API similar to type casters
* Add something like `py::nodelete` that does the same thing
* Hide it behind a preprocessor macro like `NB_NEVER_DESTROY(StaticOnly2)`

virtuald:
- This is useful for wrapping singletons
- Fixes #1248

Not very intrusive in nanobind itself, but the tests were a bit intrusive as I needed to add a separate compilation unit to ensure that the incomplete type error occurred when this fix wasn't applied.


virtuald:
I think that should be all set now. I also added a note to the porting section.

Initially I put the documentation in classes.rst, but ownership_adv.rst feels like a better fit so I moved it there.

I chose the name `never_destroy` instead of `nodelete` because in pybind11 the `nodelete` is implemented as part of the holder infrastructure so it's not one-to-one. Open to other names too.

wjakob:
Fair enough. One more thing: what happens if the user tries to destruct a `never_destroy` type. For example if a constructor bound and the user calls it to construct the type from Python (I realize that's a misuse of the feature). Will later destruction then trigger this failure?

```
    if (nbi->destruct) {
        check(t->flags & (uint32_t) type_flags::is_destructible,
              "nanobind::detail::nb_inst_destruct(\"%s\"): attempted to call "
              "the destructor of a non-destructible type!",
              t->name);
        if (t->flags & (uint32_t) type_flags::has_destruct)
            t->destruct(inst_ptr(nbi));
        nbi->destruct = false;
    }
```

if so, it may be helpful to mention in the docs that this will terminate the application. Besides recommending care with return value policies, it is then worth mentioning that a constructor should not be bound.

virtuald:
Yes, you're correct that it will trigger a failure. I've updated the documentation note.

wjakob:
FYI, I decided to rename this to `nb::never_destruct` (motivated by the term "destructor) and did some more edits to the documentation. See commit 4ba51fcf795971c5d603d875ae4184bc0c9bd8e6 for those changes.

PR #1253, opened 2025-12-17
Support casting of datetime with timezone information

lbonn:
Updated (and fixed) version of #1144.

> Before these changes, nanobind would silently accept datetime objects with timezone information and treat them as local time.
>
> In the naive timestamp case (tzinfo=None), the cost should be small.


lbonn:
Updated the changelog and fixed a memory leak in the limited api implementation.

Indeed, the previous behavior was documented but I don't think there was any case where it is actually desirable so it can probably be included in a minor version bump.

PR #1255, opened 2025-12-19
Extend stubgen skip list by two enum attributes

denialhaag:
This PR extends the `stubgen` skip list by the enum attributes `_hashable_values_` and `_unhashable_values_map_` so that they do not appear in stubs.


wjakob:
Thanks!

PR #1261, opened 2026-01-11
feat(ndarray): add fallback support for capsule-backed NumPy arrays

klugman-yanai:
Add fallback mechanism in type_caster<ndarray<Args...>>::from_python() to accept NumPy arrays backed by PyCapsule objects (OWNDATA=False).

When initial ndarray_import() fails, the fallback checks if the array is capsule-backed and retries with relaxed dtype constraints, enabling zero-copy behavior for arrays returned from nanobind bindings.

The fallback uses guard clauses for clean, readable code and has zero overhead for normal arrays (only activates when initial import fails).

### DIFF
--- a/include/nanobind/ndarray.h
+++ b/include/nanobind/ndarray.h
@@ -565,10 +565,54 @@ template <typename... Args> struct type_caster<ndarray<Args...>> {
             (void) shape_buf;
         }
 
+        // Standard path: try normal import
         value = Value(ndarray_import(src.ptr(), &config,
                                      flags & (uint8_t) cast_flags::convert,
                                      cleanup));
 
+        // Early return if import succeeded
+        if (value.is_valid()) {
+            return true;
+        }
+
+        // Fallback path for capsule-backed numpy arrays (OWNDATA=False)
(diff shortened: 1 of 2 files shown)


include/nanobind/ndarray.h
```diff
@@ -565,10 +565,54 @@ template <typename... Args> struct type_caster<ndarray<Args...>> {
...
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+        }
+
+        // Re-attempt import with unchanged config (dtype, shape, order, device, ro)
+        // This treats capsule-backed ownership as equivalent to owning memory
+        // for casting purposes, but enforces all other constraints strictly
+        value = Value(ndarray_import(src.ptr(), &config,
+                                     flags & (uint8_t) cast_flags::convert,
+                                     cleanup));
```
hpkfft: This call looks the same as the one above (line 569).
If the first call failed, why should this be expected to succeed?

klugman-yanai:
I'll take a look and test in my own environment. Closing the PR until then. Thanks for the responsiveness.

PR #1271, opened 2026-01-21
Disable ft build in cmake

awni:
@wjakob wdyt about something like this (codex did it)? It queries the active interpreter for `Py_GIL_DISABLED` and sets a new build flag `NB_PY_GIL_DISABLED` which it uses to add `nb_ft.cpp` to the target or not.


awni:
Done.. much cleaner. I think you can review it.

PR #1285, opened 2026-02-05
CMake: add support for the Realtime Sanitizer (RTSan) 

ahoarau:
Hello,
In this PR we add support the realtime sanitizer (RTSan) from clang 20, which currently failed be te detected.
Also a pseudo-support for generator expressions like `$<$<BOOL:${ENABLE_RTSAN}>:-fsanitize=realtime>`
A full support requires to defer the detection at generation-time, not the focus of this PR.
Best,


ahoarau:
> I don't understand well enough how the realtime sanitizer would be used with Python. That seems inherently "non-realtime" (GC pauses, etc.). Basically I don't think why it would make sense to ship this PR as part of nanobind.

It's a great new tool that appeared in clang 20. The difference with other sanitizers is that it's scoped-based, meaning that only the c++ functions that are instrumented with the `[[clang::nonblocking]]` are checked at runtime, not the whole program.  
So it's definitely compatible with any wrapping environnement like python.

Without this patch, the stub generation fails with an error like this:
```
[build] ==47391==ERROR: Interceptors are not working. This may be because RealtimeSanitizer is loaded too late (e.g. via dlopen). Please launch the executable with:
[build] DYLD_INSERT_LIBRARIES=/opt/homebrew/Cellar/llvm/21.1.8/lib/clang/21/lib/darwin/libclang_rt.rtsan_osx_dynamic.dylib
[build] "interceptors not installed" && 0
```

wjakob:
Can you let me know what this change does? (the addition to `>`)

```
if(opt MATCHES "-fsanitize=([^ >]+)")
```

Issue #957, opened 2025-03-03
[BUG]: memory leaks.


jcelerier:
### Problem description

When running with asan I'm seeing leaks in multiple traces:

https://github.com/celtera/libremidi/actions/runs/13620071032/job/38068205086?pr=140

My code is very basic and only initializes types and enums, it's entirely contained in this file:
https://github.com/celtera/libremidi/blob/master/bindings/python/pylibremidi.cpp#L185
 
```
=================================================================
==6022==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1096 byte(s) in 1 object(s) allocated from:
    #0 0x[7](https://github.com/celtera/libremidi/actions/runs/13620071032/job/38068205086?pr=140#step:9:8)fdde40fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x5[8](https://github.com/celtera/libremidi/actions/runs/13620071032/job/38068205086?pr=140#step:9:9)6d36 in PyObject_Malloc (/usr/bin/python3.12+0x586d36) (BuildId: 37451b37c71cb46f8ccb27cb3cdbb7aa004b[9](https://github.com/celtera/libremidi/actions/runs/13620071032/job/38068205086?pr=140#step:9:10)987)
    #2 0x597fc8  (/usr/bin/python3.12+0x597fc8) (BuildId: 37451b37c71cb46f8ccb27cb3cdbb7aa004b9987)
    #3 0x7fdddef16634 in nanobind::detail::init(char const*) /home/runner/work/libremidi/libremidi/build/_deps/nanobind-src/src/nb_internals.cpp:370
    #4 0x7fdddea20edb in PyInit_pylibremidi /home/runner/work/libremidi/libremidi/bindings/python/pylibremidi.cpp:185
    #5 0x6a9630  (/usr/bin/python3.12+0x6a9630) (BuildId: 37451b37c71cb46f8ccb27cb3cdbb7aa004b9987)
    #6 0x6a8d81  (/usr/bin/python3.12+0x6a8d81) (BuildId: 37451b37c71cb46f8ccb27cb3cdbb7aa004b9987)
    #7 0x582126  (/usr/bin/python3.12+0x582126) (BuildId: 37451b37c71cb46f8ccb27cb3cdbb7aa004b9987)
[...]

PR #1287, opened 2026-02-09
Fix function calls after module deletion (fixes #1283)

### LINKED ISSUE #1283: [BUG]: calls to functions using `ndarray`s cause segmentation fault if their module is not referenced
### Problem description

Calls to functions using ndarrays crash the interpreter (segmentation fault) since nanobind 2.10, if there are not live references to the module containing the functions. In 2.9 and previous versions this works without problems. It looks like the module constants introduced in https://github.com/wjakob/nanobind/pull/1184 are deleted when the module is not longer referenced, which causes the crash when the function is called. @tehrengruber originally experienced the crashes and narrowed down the cause to this issue.

To give a bit of context, we use nanobind to load just-in-time generated C++ extension modules which only contain a single function dealing with CPU or GPU arrays. The generated module is not in the `sys.path`, so we load it using `importlib.util` functions, and we don't keep any references to the module, only to the function.

### Reproducible example code

```text
# A very simple reproducible example can be created from the nanobind example repo
# (https://github.com/wjakob/nanobind_example) by just adding a function using ndarrays
# (`get_ndim` in this case) to the C++ extension module:

# ---- C++ extension module ---- #
#include <nanobind/ndarray.h>

namespace nb = nanobind;

using namespace nb::literals;

NB_MODULE(nanobind_example_ext, m) {
    m.doc() = "This is a \"hello world\" example with nanobind";
    m.def("add", [](int a, int b) { return a + b; }, "a"_a, "b"_a);
    m.def("get_ndim", [](nb::ndarray<> array) { return array.ndim(); });
}
# ------------------------- #

# Now, if we load the extension module in python using `importlib.util` functions and delete
# the reference to the module, calls to the `add` function will still work but calls to `get_ndim`
# will crash the interpreter in nanobind 2.10 or later versions:

# ---- Python script ---- #
#!/usr/bin/env python3
import importlib.util as iutil
import numpy as np

spec = iutil.spec_from_file_location(
    "nanobind_example_ext",
    ".venv/lib/python3.14/site-packages/nanobind_example/nanobind_example_ext.abi3.so"
)
module = iutil.module_from_spec(spec)
spec.loader.exec_module(module)
get_ndim = module.get_ndim
add = module.add

arr = np.array([1, 2, 3])

print(f"{add(1, 2)=}")
print(f"{get_ndim(arr)=}")

del module
print("\n-- Module deleted --\n")

print(f"{add(1, 2)=}")
print(f"{get_ndim(arr)=}")  # This call causes the segmentation fault in nanobind >= 2.10
# ------------------------- #

wjakob:
Nanobind 2.10 introduced a local stash of interned strings to accelerate element access in some common situations (mostly related to ndarrays). The lifetime of this stash was coupled to that of the module.

However, this meant that freeing the module (e.g., by dropping the reference from ``sys.modules``) and subsequently calling functions from that module could cause a use-after-free issue.

This commit fixes this by making the internals state reference-counted. References are held by

- modules
- functions
- types

When the reference count reaches zero, we now not only release the interned strings but also other nanobind-related constructs (nb_func, nb_type, etc.). This should also address the memory leak issue reported in issue #957.

An ABI version bump is required by these changes.

### DIFF
--- a/include/nanobind/nb_defs.h
+++ b/include/nanobind/nb_defs.h
@@ -169,8 +169,6 @@
     X(const X &) = delete;                                                     \
     X &operator=(const X &) = delete;
 
-#define NB_MOD_STATE_SIZE (12 * sizeof(PyObject*))
-
 // Helper macros to ensure macro arguments are expanded before token pasting/stringification
 #define NB_MODULE_IMPL(name, variable) NB_MODULE_IMPL2(name, variable)
 #define NB_MODULE_IMPL2(name, variable)                                        \
@@ -196,9 +194,9 @@
         NB_MODULE_SLOTS_2                                                      \
     };                                                                         \
     static struct PyModuleDef nanobind_##name##_module = {                     \
-        PyModuleDef_HEAD_INIT, #name, nullptr, NB_MOD_STATE_SIZE, nullptr,     \
-        nanobind_##name##_slots, nanobind::detail::nb_module_traverse,         \
-        nanobind::detail::nb_module_clear, nanobind::detail::nb_module_free    \
+        PyModuleDef_HEAD_INIT, #name, nullptr, 0, nullptr,                     \
+        nanobind_##name##_slots, nullptr, nullptr,                             \
+        nanobind::detail::nb_module_free                                       \
     };                                                                         \
     extern "C" [[maybe_unused]] NB_EXPORT PyObject *PyInit_##name(void);       \
     extern "C" PyObject *PyInit_##name(void) {                                 \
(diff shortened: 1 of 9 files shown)


hpkfft:
It's definitely nice to fix both the use-after-free after a module is deleted and also the reported memory leak!
It's also nice not to have both module state and internals; it seems it should be one or the other, not both.
Though, I had hoped that internals could move over to module state instead of the other way around.
I guess I'm thinking about possible future support for subinterpreters, and I have a vague feeling that module state is the way to go for that.  Although, I freely admit that I don't know what I'm talking about.  :)

egparedes:
> @egparedes Can you verify that this fixes the issue you encountered?

Sorry for the late reply. I just ran some tests with the version in `main` after merging this PR and it seems to work fine for us. If we find any other problem we will report it.

Thanks a lot for the quick fix.

PR #1291, opened 2026-02-13
Add `nb::memoryview` wrapper class

nicholasjng:
Adds an `nb::memoryview` class wrapping a `memoryview` Python object. Implementation and tests follow the style of previous wrapper additions (e.g. `frozenset` and `bytearray`).

-------------------

I noticed this was not yet implemented when attempting to port a pybind11 project this week. I'm undecided on whether to add a constructor using `PyMemoryView_FromMemory` (https://docs.python.org/3/c-api/memoryview.html#c.PyMemoryView_FromMemory), since its flags are in the stable ABI starting from Python 3.11 only.


wjakob:
`PyMemoryView_FromMemory` seems kind of dangerous as an API since it doesn't provide a way of specifying an *owner*/keep-alive field. I foresee people running into serious undefined behavior issues when using such an API, so perhaps it's best not to expose it.

PR #1294, opened 2026-02-17
Add explicit casts to avoid compiler warnings

hpkfft:
As seen in https://github.com/wjakob/nanobind/actions/runs/22058320272/job/63731838031
GCC complains, `comparison of constant ‘1’ with boolean expression is always false [-Wbool-compare]`

Also, MinGW complains:
```
D:/a/nanobind/nanobind/src/nb_ndarray.cpp:274:27: warning: comparison of integer expressions of different signedness: 'long int' and 'const uint32_t' {aka 'const unsigned int'} [-Wsign-compare]
  274 |     if (max_major_version >= dlpack::major_version)
      |         ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

This PR makes the compilers happy.


wjakob:
Thanks!

PR #1302, opened 2026-03-06
[stubgen] Added a special-case for classmethods

superbobry:
This ensures that stubgen can correctly recognize the signature of classmethods and produce e.g.

    @classmethod
    def class_method(cls, value: int = 0) -> int:
        ...

instead of

    @classmethod
    def class_method(*args, *kwargs) -> int:
        ...


wjakob:
Looks great. Should something similar be done for `@staticmethod` for consistency/symmetry?

PR #1303, opened 2026-03-06
[stubgen] Wrap static properties into ClassVar[...] and Final[...]

superbobry:


lahwaacz:
@superbobry This change causes some new issues with pyright:

```
/home/user/test_classes_ext.pyi
  /home/user/test_classes_ext.pyi:35:25 - error: "Final" is not allowed in this context (reportInvalidTypeForm)
  /home/user/test_classes_ext.pyi:286:24 - error: "Shared" is not defined (reportUndefinedVariable)
  /home/user/test_classes_ext.pyi:393:17 - error: "Final" is not allowed in this context (reportInvalidTypeForm)
  /home/user/test_classes_ext.pyi:397:5 - error: "x" cannot be redeclared because parent class "StaticPropertyOverride" declares it as Final (reportGeneralTypeIssues)
  /home/user/test_classes_ext.pyi:397:17 - error: "Final" is not allowed in this context (reportInvalidTypeForm)
5 errors, 0 warnings, 0 informations
```

(I've downloaded `test_classes_ext.pyi.ref` from the repo, renamed to `test_classes_ext.pyi` and ran pyright on that.)

wjakob:
Thanks for the report @lahwaacz, I fixed this in commit https://github.com/wjakob/nanobind/commit/b7be2b2e510ee7c56bdca33d7ccb71ed850ddac4.

The remaining "Shared" is not defined error is unrelated to #1303 and is expected. `Shared` is built in a separate binding domain (NB_DOMAIN mydomain) and intentionally invisible to `test_classes_ext`, so the signature falls back to a quoted forward reference.

superbobry:
@wjakob that PEP was actually obsoleted by the Typing spec. See specifically the `ClassVar[Final[...]]` example in https://typing.python.org/en/latest/spec/qualifiers.html#uppercase-final.

wjakob:
@superbobry Would you be willing to make a PR that brings this into the form that keeps PyRight happy? Typing spec aside, it's a problem if nanobind generates stubs that don't typecheck with PyRight.

lahwaacz:
@superbobry I read that as the example `x: ClassVar[Final[int]] = 3` applies to dataclasses. Specifically, this sentence should clarify it:

> In non-dataclasses, combining ClassVar and Final is redundant, and type checkers may choose to warn or error on the redundancy.

wjakob:
@superbobry @lahwaacz The discussion about the state of these type annotations is still unresolved. I am not familiar enough with the details to make a call.

Please note that current state on `master` (including 4a8f4fcfe802174d6351b91b677ffef4fb2f2e5b based on the discussion above) will go out in the next release unless we find an alternative.

wjakob:
(Next being probably in the next few days. Please tell me if you want to jump in.)

PR #1307, opened 2026-03-09
Remove /MP when using Clang in MSVC compat mode

heyer2:
When using Clang in MSVC compatibility mode `/MP` is not supported. This PR ensures that `/MP` is limited to MSVC proper.


wjakob:
Thanks!

PR #1311, opened 2026-03-23
Robust infinite recursion check in StubGen.put()

jbcoe:
When `value` is a type that overrides `__eq__` (like a numpy array), the check `if value in self.stack` may no longer be a simple boolean. Checking using identifity rather than equality avoids this problem.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -183,6 +183,7 @@ set(TEST_FILES
   test_ndarray.py
   test_jax.py
   test_tensorflow.py
+  test_stubgen.py
   test_stubs.py
   test_typing.py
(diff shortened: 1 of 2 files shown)


jbcoe:
@mgorny I think I need some pointers in setting up tests for this change. The logic itself feels sound.

mgorny:
I'm sorry but I really don't know why you're asking me.

jbcoe:
> I'm sorry but I really don't know why you're asking me.

I’d meant to tag @wjakob . Sorry!

jbcoe:
Rebased and added tests using relative paths for imports.

jbcoe:
PR is no longer needed.

PR #1315, opened 2026-03-26
[stubgen] Added a new convenience type `nb::typing_self`

superbobry:
The type is useful for defining factory methods which return an instance of the enclosing class or a subclass.

I was unsure about the best name for this, since `nb::self` is already used for operator definitions. Ideas welcome!


superbobry:
Hey @wjakob, just a friendly ping in case this slipped through.

PR #1327, opened 2026-04-14
tests : change Eigen `find_package()` version test to include Eigen version 5

ManifoldFR:
This small PR ensures that the tests do not reject Eigen3 if the version number is 5.x


wjakob:
Thanks!

PR #1328, opened 2026-04-15
Add xtensor support

keltecc:
[xtensor](https://github.com/xtensor-stack/xtensor) is a header-only C++ library for multi-dimensional arrays and tensors with lazy evaluation semantics. It provides NumPy-like syntax for element-wise operations, broadcasting, and mathematical functions.

This PR adds native casters for xtensor types, enabling seamless interop between NumPy arrays and xtensor containers.

## Owning containers

`xt::xarray<T>` and `xt::xtensor<T, N>` are owning containers backed by contiguous storage. Since they must own their data, the caster copies from the NumPy buffer on input. The copy handles layout conversion transparently — a row-major function accepts Fortran-contiguous input and reorders during the copy. With `nb::arg().noconvert()`, mismatched layouts are rejected instead. On output, returning by value moves the container into a capsule for zero extra copy; `rv_policy::reference_internal` keeps the parent object alive when returning references from bound classes.

```cpp
m.def("compute", [](const xt::xarray<double>& a, const double& s, const double& t) {
    return xt::sin(a) * s + t;
});

m.def("compute", [](const xt::xtensor<double, 2>& a, const double& s, const double& t) {
    return xt::sin(a) * s + t;
});
```

## Zero-copy views

For performance-critical paths, `nb::xarray_view<T>` and `nb::xtensor_view<T, N>` wrap the NumPy buffer directly without copying:

```cpp
m.def("compute", [](const nb::xarray_view<double>& a, const double& s, const double& t) {
    return xt::sin(a) * s + t;
});

m.def("compute", [](const nb::xtensor_view<double, 2>& a, const double& s, const double& t) {
    return xt::sin(a) * s + t;
});
```

Views are layout-aware with a template parameter that defaults to `layout_type::dynamic`, which accepts any array (including non-contiguous slices) and detects the actual memory layout for optimized iteration. For maximum performance, views can be narrowed to `layout_type::row_major` or `layout_type::column_major`, which gives xtensor a compile-time contiguity guarantee and enables fast flat-pointer iteration. Since views point directly into the NumPy buffer, they do not support implicit type conversion and can mutate the source array in-place.

## Expression returns

Returning an xtensor expression (e.g., `xt::sin(a) * s + t`) produces a lazy expression tree. The `xexpression` caster matches any `xt::is_xexpression<T>` type — covering lazy expression templates, `xt::strided_view(...)`, and `xt::adapt(...)` — and evaluates it directly into a freshly allocated NumPy buffer, avoiding intermediate container materialization.

## Vectorization

`nb::xvectorize` wraps a scalar C++ function for element-wise application over arrays, analogous to `np.vectorize` but compiled:

```cpp
m.def("fast_abs", nb::xvectorize([](double x) -> double {
    return std::abs(x);
}));
```

## Scope

The entire implementation is ~500 lines (excluding tests), split across 6 headers under `include/nanobind/xtensor/`.

## Benchmark

A standalone benchmark suite is available at [keltecc/xtensor-nanobind-benchmark](https://github.com/keltecc/xtensor-nanobind-benchmark).

For API details, see the documentation in `xtensor.rst` and the test cases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,21 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
       run: sudo apt-get -y install libeigen3-dev
 
+    # ubuntu packages xtensor-dev and xtl-dev are outdated
+    # so we will install them from sources
+    - name: Install xtensor and xtl
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        mkdir -p /tmp/xtensor-install && cd /tmp/xtensor-install
+        wget https://github.com/xtensor-stack/xtl/archive/refs/tags/0.8.2.zip -O xtl-0.8.2.zip
+        unzip xtl-0.8.2.zip && cd xtl-0.8.2
+        cmake . && sudo make install
+        cd /tmp/xtensor-install
+        wget https://github.com/xtensor-stack/xtensor/archive/refs/tags/0.26.0.zip -O xtensor-0.26.0.zip
+        unzip xtensor-0.26.0.zip && cd xtensor-0.26.0
+        mkdir build && cd build
+        cmake .. && sudo make install
+
     - name: Install PyTest
       run: |
         python -m pip install pytest pytest-github-actions-annotate-failures typing_extensions
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,21 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
       run: sudo apt-get -y install libeigen3-dev
 
+    # ubuntu packages xtensor-dev and xtl-dev are outdated
+    # so we will install them from sources    
+    - name: Install xtensor and xtl
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        mkdir -p /tmp/xtensor-install && cd /tmp/xtensor-install
+        wget https://github.com/xtensor-stack/xtl/archive/refs/tags/0.8.2.zip -O xtl-0.8.2.zip
+        unzip xtl-0.8.2.zip && cd xtl-0.8.2
+        cmake . && sudo make install
+        cd /tmp/xtensor-install
+        wget https://github.com/xtensor-stack/xtensor/archive/refs/tags/0.26.0.zip -O xtensor-0.26.0.zip
+        unzip xtensor-0.26.0.zip && cd xtensor-0.26.0
+        mkdir build && cd build
+        cmake .. && sudo make install
+
     - name: Install PyTest
       run: |
         python -m pip install pytest pytest-github-actions-annotate-failures typing_extensions
--- a/docs/exchanging.rst
+++ b/docs/exchanging.rst
@@ -145,6 +145,8 @@ to external projects that provide further casters:
     - ``#include <nanobind/eigen/sparse.h>``
   * - ``Eigen::Tensor<..>``, ``Eigen::TensorMap<..>``, ``Eigen::TensorRef<..>``
     - ``#include <nanobind/eigen/tensor.h>``
+  * - ``xt::xarray<..>``, ``xt::xarray_container<..>``, ``xt::xtensor<..>``, ``xt::xtensor_container<..>``
+    - ``#include <nanobind/xtensor/xcontainer.h>``
   * - Apache Arrow types
     -  `https://github.com/maximiliank/nanobind_pyarrow <https://github.com/maximiliank/nanobind_pyarrow>`__
   * - ...
--- a/docs/ndarray_index.rst
+++ b/docs/ndarray_index.rst
@@ -11,3 +11,4 @@ Python and C++.
 
    ndarray
    eigen
+   xtensor
--- a/docs/xtensor.rst
+++ b/docs/xtensor.rst
@@ -1,0 +1,401 @@
+.. cpp:namespace:: nanobind
+
+.. _xtensor:
+
+The *xtensor* multi-dimensional array library
+==============================================
+
+`xtensor <https://github.com/xtensor-stack/xtensor>`__ is a header-only C++
+library for multi-dimensional arrays and tensors with lazy evaluation
+semantics. It provides NumPy-like syntax for element-wise operations,
+broadcasting, and mathematical functions.
+
+nanobind includes custom type casters that enable bidirectional conversion
+between xtensor and NumPy arrays. These casters build on the :ref:`n-dimensional array <ndarray_class>` class.
+
+Setup
+-----
+
+Add the following include directive to your binding code:
+
+.. code-block:: cpp
+
+   #include <nanobind/xtensor.h>
+
+This umbrella header pulls in casters for all supported xtensor types. You can
+also include individual headers if you only need a subset:
+
+.. code-block:: cpp
+
+   #include <nanobind/xtensor/xcontainer.h>  // xt::xarray, xt::xtensor
+   #include <nanobind/xtensor/xview.h>       // nb::xarray_view, nb::xtensor_view
+   #include <nanobind/xtensor/xexpression.h> // lazy xexpression return
+   #include <nanobind/xtensor/xvectorize.h>  // nb::xvectorize
+
+Owning containers
+-----------------
+
+Owning containers (``xt::xarray<T>`` and ``xt::xtensor<T, N>``) manage their
+own memory. When converting between Python and C++, data is **always copied**.
+
+Python → C++
+^^^^^^^^^^^^^
+
+When a Python NumPy array is passed to a C++ function expecting an owning
+container, the data is **always copied** into the container. This means the
+container owns its memory independently of the Python object.
+
+.. code-block:: cpp
+
+   // Data is copied into the xarray, safe to modify
+   m.def("process", [](xt::xarray<double>& arr) {
+       arr(0) = 999.0;  // does NOT affect the Python array
+       return arr;
+   });
+
+The owning caster supports implicit conversion at two levels: if the NumPy
+array has a different dtype (e.g., ``float32`` when ``double`` is expected),
+nanobind will convert it automatically. For containers with a strict layout
+(e.g., ``row_major``), a non-matching memory order is also reordered during
+the copy. Both can be disabled with ``nb::arg().noconvert()``, which rejects
+dtype or layout mismatches with a ``TypeError``.
+
+C++ → Python
+^^^^^^^^^^^^^
+
+When a C++ function returns an owning container, the caster creates a NumPy
+array. The behavior depends on the return value policy:
+
+- **Return by value** (``rv_policy::move``): the container is move-constructed
+  into a capsule that owns the data. The resulting NumPy array references this
+  capsule with zero copy.
+
+- **Return by reference**: the NumPy array references the original C++ data.
+  Use ``rv_policy::reference_internal`` when the data is owned by a bound C++
+  object to ensure correct lifetimes.
+
+.. code-block:: cpp
+
+   // Returns by value, data is moved, no copy
+   m.def("make_array", []() {
+       return xt::xarray<double>{1.0, 2.0, 3.0};
+   });
+
+   // Returns by reference from a method, reference_internal keeps self alive
+   nb::class_<MyClass>(m, "MyClass")
+       .def("get_data", &MyClass::get_data, nb::rv_policy::reference_internal);
+
+.. warning::
+
+   ``rv_policy::reference_internal`` must only be used with **methods** (where
+   a ``self`` object exists). It works by preventing the Python wrapper of
+   ``self`` from being garbage-collected while the returned array is alive.
+
+   For **free functions** returning references to static or global data, use
+   ``rv_policy::reference`` instead, there is no ``self`` to keep alive, and
+   using ``reference_internal`` will cause a segmentation fault:
+
+   .. code-block:: cpp
+
+      static xt::xarray<double> global = {1.0, 2.0, 3.0};
+
+      // WRONG, free function has no self → segfault
+      m.def("bad", []() -> xt::xarray<double>& {
+          return global;
+      }, nb::rv_policy::reference_internal);
+
+      // CORRECT, use reference for static/global data
+      m.def("good", []() -> xt::xarray<double>& {
+          return global;
+      }, nb::rv_policy::reference);
+
+Non-owning views
+----------------
+
+Non-owning views (``nb::xarray_view<T>`` and ``nb::xtensor_view<T, N>``)
+provide **zero-copy** access to the underlying NumPy array's data. They do not
+own the memory and instead wrap the Python buffer directly using
+``xt::adapt()`` with ``xt::no_ownership()``.
+
+.. code-block:: cpp
+
+   #include <nanobind/xtensor.h>
+   namespace nb = nanobind;
+
+   // Zero-copy read access
+   m.def("sum_view", [](const nb::xarray_view<double>& a) {
+       return xt::sum(a)();
+   });
+
+   // Zero-copy write, modifies the Python array in-place
+   m.def("mutate", [](nb::xarray_view<double>& a) {
+       a(0) = 999.0;
+   });
+
+Because views wrap existing memory without copying, they **do not support
+implicit type conversion**. If the NumPy array's dtype does not match the
+expected scalar type exactly, nanobind will raise a ``TypeError``. This is
+intentional: converting to a temporary buffer would create a copy that the view
+would reference, leading to a dangling pointer when the temporary is destroyed.
+
+Owning vs. non-owning: when to use which
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
++-------------------+---------------------------+---------------------------+
+|                   | Owning                    | Non-owning (view)         |
+|                   | (``xt::xarray``,          | (``nb::xarray_view``,     |
+|                   | ``xt::xtensor``)          | ``nb::xtensor_view``)     |
++===================+===========================+===========================+
+| Memory            | Copies data               | Wraps existing buffer     |
++-------------------+---------------------------+---------------------------+
+| Type conversion   | Implicit (converts dtype) | Strict (exact dtype only) |
++-------------------+---------------------------+---------------------------+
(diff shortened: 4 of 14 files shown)


hpkfft:
Hi Vladimir,
I'm curious to ask how this relates to [xtensor-python](https://github.com/xtensor-stack/xtensor-python).  Is this PR an independent effort?  Have you benchmarked against that project?
Also, does this PR require C++20?

keltecc:
Hi @hpkfft, this PR is written from scratch, I suppose it would be a successor of xtensor-python (based on pybind11). I don't have a benchmark that compares both implementations yet, but plan to write it. xtensor-python is extremely optimized and quite fast, the target of this PR is to achieve the same performance, but I believe it could be even faster since nanobind function calls have less overhead. I have a benchmark for nanobind implementation only and it shows promising results

It does not require C++20 actually, xtensor 0.26.0 (which is installed in tests) requires C++17, it should work. xtensor 0.27.0 requires C++20, but that depends on which xtensor version user installs

Currently this PR is in progress, I'm still updating the library interface and testing the behavior. I've opened this draft for public testing and benchmarking

wjakob:
This looks cool. I would be excited to have nice xtensor bindings in nanobind 👍

keltecc:
some benchmark results on AMD Ryzen 9 9950X with xsimd enabled

more details in [the benchmark repository](https://github.com/keltecc/xtensor-nanobind-benchmark#sample-output)

```
=== Nanobind vs xtensor-python ===
+-------------------------------------------+----------+----------------+
| Metric                                    | nanobind | xtensor-python |
+-------------------------------------------+----------+----------------+
| Call overhead  [ndarray]                  |    84 ns |         308 ns |
| Call overhead  [view]                     |   109 ns |          86 ns |
| Call overhead  [owning array]             | 110.4 us |       132.1 us |
| Sum 1D  [ndarray]                         | 390.1 us |       394.2 us |
| Sum 1D  [view]                            | 392.9 us |        2.33 ms |
| Sum 1D  [owning array]                    | 520.3 us |       546.5 us |
| Sum 2D  [ndarray]                         | 391.3 us |       389.2 us |
| Sum 2D  [view]                            | 365.8 us |       391.7 us |
| Sum 2D  [owning array]                    | 543.8 us |       522.9 us |
| Compute sin(a)*s+t  [view]                | 811.2 us |       843.4 us |
| Compute sin(a)*s+t  [owning array]        | 997.5 us |        3.01 ms |
| Compute2D xarray  [xarray row_major]      |  1.05 ms |        1.13 ms |
| Compute2D xarray  [xarray dynamic]        |  1.05 ms |        1.11 ms |
| Compute2D xtensor  [xtensor row_major]    |  1.03 ms |        1.11 ms |
| Compute2D xtensor  [xtensor dynamic]      |  1.05 ms |        1.09 ms |
| Compute2D view  [view row_major]          | 813.4 us |       839.6 us |
| Compute2D view  [view dynamic]            | 813.0 us |       848.0 us |
| Compute2D t.view  [tensor_view row_major] | 817.3 us |       794.0 us |
| Compute2D t.view  [tensor_view dynamic]   | 836.4 us |       805.8 us |
| Vectorize abs  [xvectorize]               | 138.4 us |       127.2 us |
+-------------------------------------------+----------+----------------+
```

wjakob:
I guess this one is puzzling: what's happening with views that would cause nanobind to be slower?
```
| Call overhead  [view]                     |   109 ns |          86 ns |
```

keltecc:
> what's happening with views that would cause nanobind to be slower?

good question, this part is tricky! let's look at the full call overhead table:

```
=== Call Overhead (return first element) ===
+---------------+--------------+----------------+
| Backend       |     nanobind | xtensor-python |
+---------------+--------------+----------------+
| array view    |       109 ns |          86 ns |
| tensor view   |        77 ns |          98 ns |
| owning array  |     110.4 us |       132.1 us |
| owning tensor |     106.8 us |       113.6 us |
+---------------+--------------+----------------+
```

it turns out that the results are different for views:

- for xarray views (`nb::xarray_view` and `xt::pyarray`) nanobind is slower
- for xtensor views (`nb::xtensor_view` and `xt::pytensor`) nanobind is faster

the actual implementations of views are slightly different. xtensor-python *does not* copy array metadata (shape and strides), so their views are truly zero-copy. to achieve this it implements custom inner types for iterators and steppers. our implementation *does* copy array metadata to achieve code simplicity, that causes the difference in benchmark:

1. `nb::xarray_view` allocates two `std::vector<T>` storages. since this is a two heap allocations, it costs some time and it's slower
2. `nb::xtensor_view` allocates two `std::array<T, N>` storages. this allocation is stack-based, it does not consume time for heap allocation, so it's faster

basically, there are two possible optimizations: avoid heap allocation and avoid copying of the metadata (pointing directly to ndarray internal buffers). but I suppose this could be done later in the further development

hpkfft:
> `nb::xarray_view` allocates two `std::vector<T>` storages.

If both shape and strides were `int64_t`, it might be possible to have only one vector, `shape_and_strides`, similar to line 212 of nb_ndarray.cpp.

keltecc:
> If both shape and strides were `int64_t`, it might be possible to have only one vector, `shape_and_strides`, similar to line 212 of nb_ndarray.cpp.

interesting idea, but there are integer types mismatch in our case:

1. ndarray interface, `shape(i)` and `stride(i)` return types are different: `size_t` (unsigned) and `int64_t` (signed)
2. xarray adaptor's shape type is `std::vector<std::size_t>` (unsigned) and strides type is `std::vector<std::ptrdiff_t>` (signed)

it seems we can't use the common vector, but can create a custom structure containing both vectors and allocate it only once on the heap. however we still need to copy the metadata values and this could be optimized too (via wrapping the raw ndarray pointers)

I would suggest to make these optimizations as a different research, remaining the current PR as a basic xtensor support

keltecc:
I have tested the implementation with valgrind, it seems that we have no leaks related to the nanobind code:

<details>
<summary>valgrind output</summary>

```text
$ valgrind --leak-check=full python3 -m pytest tests/test_xtensor.py
==10900== Memcheck, a memory error detector
==10900== Copyright (C) 2002-2024, and GNU GPL'd, by Julian Seward et al.
==10900== Using Valgrind-3.26.0 and LibVEX; rerun with -h for copyright info
==10900== Command: python3 -m pytest tests/test_xtensor.py
==10900==
============================================== test session starts ==============================================
platform linux -- Python 3.14.4, pytest-9.0.3, pluggy-1.6.0
rootdir: /nanobind
configfile: pyproject.toml
collected 88 items

tests/test_xtensor.py ................................................................................... [ 94%]
.....                                                                                                     [100%]

============================================== 88 passed in 4.40s ===============================================
==10900==
==10900== HEAP SUMMARY:
==10900==     in use at exit: 2,207,808 bytes in 1,671 blocks
==10900==   total heap usage: 30,687 allocs, 29,016 frees, 111,350,998 bytes allocated
==10900==
==10900== 528 bytes in 1 blocks are possibly lost in loss record 119 of 441
==10900==    at 0x484B858: malloc (vg_replace_malloc.c:447)
==10900==    by 0x513283: _PyObject_GC_NewVar (in /usr/bin/python3.14)
==10900==    by 0x50C6F0: PyTuple_New (in /usr/bin/python3.14)
==10900==    by 0x53A80F: ??? (in /usr/bin/python3.14)
==10900==    by 0x53AAC9: ??? (in /usr/bin/python3.14)
==10900==    by 0x53A8F7: ??? (in /usr/bin/python3.14)
==10900==    by 0x53AAB1: ??? (in /usr/bin/python3.14)
==10900==    by 0x53A8F7: ??? (in /usr/bin/python3.14)
==10900==    by 0x53AAB1: ??? (in /usr/bin/python3.14)
==10900==    by 0x68E7DD: ??? (in /usr/bin/python3.14)
==10900==    by 0x6B8519: ??? (in /usr/bin/python3.14)
==10900==    by 0x54FCE3: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==
==10900== 1,984 bytes in 3 blocks are possibly lost in loss record 339 of 441
==10900==    at 0x484B858: malloc (vg_replace_malloc.c:447)
==10900==    by 0x513283: _PyObject_GC_NewVar (in /usr/bin/python3.14)
==10900==    by 0x50C6F0: PyTuple_New (in /usr/bin/python3.14)
==10900==    by 0x53A80F: ??? (in /usr/bin/python3.14)
==10900==    by 0x53A8F7: ??? (in /usr/bin/python3.14)
==10900==    by 0x53AAB1: ??? (in /usr/bin/python3.14)
==10900==    by 0x53A8F7: ??? (in /usr/bin/python3.14)
==10900==    by 0x53AAB1: ??? (in /usr/bin/python3.14)
==10900==    by 0x68E7DD: ??? (in /usr/bin/python3.14)
==10900==    by 0x6B8519: ??? (in /usr/bin/python3.14)
==10900==    by 0x54FCE3: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x58C803: ??? (in /usr/bin/python3.14)
==10900==
==10900== 2,495 bytes in 1 blocks are possibly lost in loss record 352 of 441
==10900==    at 0x484B858: malloc (vg_replace_malloc.c:447)
==10900==    by 0x5B5B6D: _PyUnicode_JoinArray (in /usr/bin/python3.14)
==10900==    by 0x5E953F: ??? (in /usr/bin/python3.14)
==10900==    by 0x5355B2: PyObject_Vectorcall (in /usr/bin/python3.14)
==10900==    by 0x55023D: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x547BD5: PyEval_EvalCode (in /usr/bin/python3.14)
==10900==    by 0x5E2B5A: ??? (in /usr/bin/python3.14)
==10900==    by 0x55132C: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x58C803: ??? (in /usr/bin/python3.14)
==10900==    by 0x57691E: ??? (in /usr/bin/python3.14)
==10900==    by 0x5CAF66: PyObject_CallMethodObjArgs (in /usr/bin/python3.14)
==10900==    by 0x6A8592: ??? (in /usr/bin/python3.14)
==10900==
==10900== 11,056 bytes in 8 blocks are possibly lost in loss record 410 of 441
==10900==    at 0x484B858: malloc (vg_replace_malloc.c:447)
==10900==    by 0x513283: _PyObject_GC_NewVar (in /usr/bin/python3.14)
==10900==    by 0x512E7E: _PyTuple_FromArray (in /usr/bin/python3.14)
==10900==    by 0x553C29: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x53680E: ??? (in /usr/bin/python3.14)
==10900==    by 0x5955BE: ??? (in /usr/bin/python3.14)
==10900==    by 0x52F311: _PyObject_MakeTpCall (in /usr/bin/python3.14)
==10900==    by 0x54D935: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x58C803: ??? (in /usr/bin/python3.14)
==10900==    by 0x709185: ??? (in /usr/bin/python3.14)
==10900==    by 0x5A53CF: ??? (in /usr/bin/python3.14)
==10900==    by 0x55132C: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==
==10900== 170,216 bytes in 94 blocks are possibly lost in loss record 440 of 441
==10900==    at 0x484B858: malloc (vg_replace_malloc.c:447)
==10900==    by 0x5B5B6D: _PyUnicode_JoinArray (in /usr/bin/python3.14)
==10900==    by 0x5E953F: ??? (in /usr/bin/python3.14)
==10900==    by 0x54EE77: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x547BD5: PyEval_EvalCode (in /usr/bin/python3.14)
==10900==    by 0x5E2B5A: ??? (in /usr/bin/python3.14)
==10900==    by 0x55132C: _PyEval_EvalFrameDefault (in /usr/bin/python3.14)
==10900==    by 0x58C803: ??? (in /usr/bin/python3.14)
==10900==    by 0x57691E: ??? (in /usr/bin/python3.14)
==10900==    by 0x5CAF66: PyObject_CallMethodObjArgs (in /usr/bin/python3.14)
==10900==    by 0x6A8592: ??? (in /usr/bin/python3.14)
==10900==    by 0x5C99DD: PyImport_ImportModuleLevelObject (in /usr/bin/python3.14)
==10900==
==10900== LEAK SUMMARY:
==10900==    definitely lost: 0 bytes in 0 blocks
==10900==    indirectly lost: 0 bytes in 0 blocks
==10900==      possibly lost: 186,279 bytes in 107 blocks
==10900==    still reachable: 2,021,497 bytes in 1,563 blocks
==10900==         suppressed: 32 bytes in 1 blocks
==10900== Reachable blocks (those to which a pointer was found) are not shown.
==10900== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==10900==
==10900== For lists of detected and suppressed errors, rerun with: -s
==10900== ERROR SUMMARY: 5 errors from 5 contexts (suppressed: 0 from 0)
```

</details>

wjakob:
What's the status of this PR? Ready to be merged?

keltecc:
> What's the status of this PR? Ready to be merged?

yes, I believe the PR is ready. the main features (support for owning and non-owning containers) are implemented, covered with tests and benchmarked. the performance is similar to xtensor-python and even better in some cases

there are some things to improve (for example an optimization of metadata usage, see a discussion above), but I think it's not necessary at the moment

about metadata (shape and strides): I think it could be implemented via wrappers around ndarray's shape and strides buffers, it seems it fits the current architecture, but won't make a significant effect to the performance

wjakob:
I am not very familiar with xarray and outsourced to Claude Opus 4.8 for a first review pass asking it to compare in particular if this caster is consistent with the ndarray constraints API and how things work in the (established and widely used) Eigen type caster.

Please see the comments it had. Items 1-4 look legit, though of course take with a grain of salt coming from the big black box.
```
  Correctness concerns

  1. Owning caster rejects read-only inputs — a regression vs. the Eigen reference. xcontainer_caster uses NDArray = ndarray<Scalar, numpy> with a non-const scalar for from_python (xcontainer.h:652). I confirmed
  in src/nb_ndarray.cpp:729-734 that ndarray_import hard-rejects a read-only source whenever a writable view is requested, before any convert fallback (which only handles dtype/order, nb_ndarray.cpp:805). So
  passing a read-only NumPy array (WRITEABLE=False, np.frombuffer(bytes), etc.) to a function taking const xt::xarray<double>& fails with a TypeError, even though the caster always copies. The Eigen plain caster
  deliberately avoids this — dense.h:133-136 uses array_for_eigen_t<T, const Scalar> with the comment "We're in any case making a copy, so non-writable inputs are also okay." The owning xtensor caster should do
  the same: use a const-scalar ndarray in from_python. This is the one issue I'd consider a must-fix.

  2. xexpression.h hardcodes row-major output but adapts with XTENSOR_DEFAULT_LAYOUT. In from_cpp (xexpression.h, ~line 808-838) the buffer is filled via xt::adapt<XTENSOR_DEFAULT_LAYOUT>(...), but the returned
  ndarray is declared c_contig and the strides are computed row-major unconditionally. The comment asserts "XTENSOR_DEFAULT_LAYOUT (row_major)". XTENSOR_DEFAULT_LAYOUT is user-overridable; if someone compiles
  with -DXTENSOR_DEFAULT_LAYOUT=...column_major, the buffer is column-major but labeled C-contiguous → transposed/garbage results. Either pin the adapt to row_major explicitly (decoupling from the macro), or
  derive the strides/order from the actual layout used. The container caster avoids this trap because its dynamic path uses explicit strides regardless of layout.

  3. The documented "segfault" for reference_internal on free functions is inaccurate. docs/xtensor.rst:142-164 warns in strong terms that using reference_internal from a free function "will cause a segmentation
  fault." It won't. Tracing it: with no self, xcontainer.h:753 skips the owner branch, policy stays reference_internal, and ndarray_export (nb_ndarray.cpp:1122-1140) finds cleanup->self() == nullptr and th->owner
  == nullptr, falls through to the automatic case, and copies — which is exactly what docs/ndarray.rst:447-450,461-463 documents. So the behavior is a safe (if non-zero-copy) copy, not a crash. The doc should be
  corrected; as written it gives users a vivid but false mental model.

     Note this also means the Eigen caster is actually more defensive here than the new code: dense.h:189-194 returns handle() (clean failure) when reference_internal lacks a self, rather than silently degrading
  to a copy. Worth deciding which behavior you want and making it consistent.

  4. Returning views is a latent dangling-pointer hazard. xview_caster::from_cpp (xview.h:990-1011) forces automatic → reference and supplies no owner unless reference_internal is used. For reference with no
  owner, ndarray_export (nb_ndarray.cpp:1147-1158) creates a genuine non-owning view. test_xarray_view_zerocopy passes only because the source a is still alive in the test scope — the returned array does not keep
  the source buffer alive. This matches Eigen::Map's behavior (dense.h:287-308 also returns with an empty owner), so it's defensible, but for views the return path is much more likely to be exercised than for
  Map. I'd at least call this out explicitly in the view section of the docs (the table mentions reference_internal but the prose doesn't warn about bare returns).

  Design / consistency

  - Missing license headers. Every existing nanobind header carries the BSD license block (e.g. eigen/dense.h:1-9). None of the six new files do. Should be added for consistency.
  - Move always heap-allocates. Eigen's move path skips the capsule for static-size or sub-1KB arrays and copies into the NumPy buffer instead (dense.h:176-181). The xtensor caster always does new Value(move) +
  capsule (xcontainer.h:748-752). Minor, but a cheap win for small returns.
  - Input signatures under-document constraints. Name is taken from the plain ndarray<Scalar, numpy> caster, so a xt::xtensor<double,2> parameter shows up in stubs/signatures as a generic ndarray[dtype=float64]
  with no rank or order, even though the caster enforces ndim==2 and possibly a layout. Eigen encodes shape and order into the type name. Not blocking, but a docstring/stub quality gap.
  - Comment style. The headers use Doxygen /// comments inside function bodies and some run 4-6 lines (e.g. the xcontainer_caster preamble). nanobind's idiom is brief // inline comments. Also a couple of typos:
  "nonconvert()" / "could be used" in xcontainer.h:647, and a stray "); inside the #error string in version.h:611.
  ```

keltecc:
thank you for the feedback! it seems there are some really good catches, I'm working on them now, will push an update soon

PR #1331, opened 2026-04-24
Add bitset caster

patrickd-galil:
This is my first pass at a simple bitset caster that just maps `std::bitset` to a python int. I am not sure if this is the simplest or most correct way to do this, but it worked for me.


include/nanobind/stl/bitset.h
```diff
@@ -0,0 +1,32 @@
...
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+template <size_t T>
+struct type_caster<std::bitset<T>> {
+    NB_TYPE_CASTER(std::bitset<T>, const_name("int"))
+
+    bool from_python(handle src, uint8_t, cleanup_list *) noexcept {
+        value = std::bitset<T>(src.ptr());
```
wjakob: This makes no sense. You are interpreting the pointer as a bit-set.

Python has a number API -- you'll have to see if the Python number in `src` fits into a `std::bitset` of the right size.

include/nanobind/stl/bitset.h
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+#include <nanobind/nanobind.h>
+#include <bitset>
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+template <size_t T>
```
wjakob: Should be ``N``, ``T`` indicates a type.

include/nanobind/stl/bitset.h
```diff
@@ -0,0 +1,32 @@
...
+    NB_TYPE_CASTER(std::bitset<T>, const_name("int"))
+
+    bool from_python(handle src, uint8_t, cleanup_list *) noexcept {
+        value = std::bitset<T>(src.ptr());
+        return true;
+    }
+
+    static handle from_cpp(const std::bitset<T> &value, rv_policy,
+                           cleanup_list *) noexcept {
+        return cast(value.to_ullong());
```
wjakob: This is only safe if the bitset value fits into a 64 bit number. For the general case, you will have to use the Python number API, which supports arbitrary-length numbers.

wjakob (commented):
Some comments. This will also need test coverage both ways, including bit sets that are too large to fit into 64 bits.

hpkfft:
Have you considered [PyLong_FromUnsignedNativeBytes](https://docs.python.org/3/c-api/long.html#c.PyLong_FromUnsignedNativeBytes) and [PyLong_AsNativeBytes](https://docs.python.org/3/c-api/long.html#c.PyLong_AsNativeBytes)?

On the C++ side, maybe use https://en.cppreference.com/cpp/utility/bitset/operator_ltltgtgt and https://en.cppreference.com/cpp/utility/bitset/operator_logic2 if more bits than fit in an `unsigned long long`, i.e., use multiple calls `bitset(unsigned long long`/`to_ullong()` after shifting and masking so each piece fits in  `unsigned long long`.  So, your C++ code reads/writes the buffer array to export/import to/from `std::bitset`.

patrickd-galil:
> Have you considered [PyLong_FromUnsignedNativeBytes](https://docs.python.org/3/c-api/long.html#c.PyLong_FromUnsignedNativeBytes) and [PyLong_AsNativeBytes](https://docs.python.org/3/c-api/long.html#c.PyLong_AsNativeBytes)?

Is it a problem that these are only in stable ABI as of Python 3.14? From the docs, I think nanobind supports stable ABI starting on Python 3.12.

hpkfft:
Yes.   :(

patrickd-galil:
I've noticed that there's some confusing behavior when constructing `std::bitset` from a string. If the string is too long, the bitset constructor will just stop reading the string once it's seen enough bits. This leads to a bit shift rather than a truncate (which would probably be less surprising).

```c
std::bitset<2>("101").to_string() // "10"
std::bitset<2>("110").to_string() // "11"
```

Since it's impossible to tell why this happens from the Python end of things (since you're passing an int, not a str), I figure we need to fail if the input is too long.

include/nanobind/stl/bitset.h
patrickd-galil: You are 100% right - I had mixed up `from_python` and `from_cpp`, so this direction was completely non-functional. I think I have it working now.

include/nanobind/stl/bitset.h
patrickd-galil: The caster is now using strings, which should play nice with arbitrary-length numbers.

wjakob:
Casters in nanobind just refuse invalid casts. There is no intended mechanism for them to supply additional error messages or justifications.

PR #1332, opened 2026-04-27
Fix Clang implicit conversion warnings

tkhyn:
Hi,
This PR fixes the implicit conversion warnings raised by Clang, mainly regarding size_t to and from Py_ssize_t
After these changes nanobind builds warningless on Clang 22

### DIFF
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -281,8 +281,8 @@ static void nb_ndarray_releasebuffer(PyObject *, Py_buffer *view) {
 
 // This function implements __dlpack__() for a nanobind.nb_ndarray.
 static PyObject *nb_ndarray_dlpack(PyObject *self, PyObject *const *args,
-                                   Py_ssize_t nargsf, PyObject *kwnames) {
-    if (PyVectorcall_NARGS(nargsf) != 0) {
+                                   Py_ssize_t nargs, PyObject *kwnames) {
+    if (nargs != 0) {
         PyErr_SetString(PyExc_TypeError,
                 "__dlpack__() does not accept positional arguments");
         return nullptr;
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -545,11 +545,11 @@ static mt_unique_ptr_t make_mt_from_buffer_protocol(PyObject *o, bool ro) {
     }
 
     static_assert(alignof(managed_dltensor_versioned) >= alignof(int64_t));
-    scoped_pymalloc<managed_dltensor_versioned> mt(1, 2 * sizeof(int64_t)*ndim);
+    scoped_pymalloc<managed_dltensor_versioned> mt(1, 2 * sizeof(int64_t) * (size_t) ndim);
     int64_t* shape = nullptr;
     int64_t* strides = nullptr;
     if (ndim > 0) {
-        shape = new ((void*) (mt.get() + 1)) int64_t[2 * ndim];
+        shape = new ((void*) (mt.get() + 1)) int64_t[2 * (size_t) ndim];
         strides = shape + ndim;
     }
 
@@ -630,7 +630,7 @@ ndarray_handle *ndarray_import(PyObject *src, const ndarray_config *c,
         capsule = borrow(src);
     } else {
         PyObject* args[] = {src, NB_INTERNED(dl_version_tpl)};
-        Py_ssize_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
+        size_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
 
         // Call src.__dlpack__(): max_version_kw requests a versioned capsule
         // nullptr the cheaper unversioned one.
@@ -911,8 +911,9 @@ ndarray_handle *ndarray_import(PyObject *src, const ndarray_config *c,
         result->free_strides = true;
 
         scoped_pymalloc<int64_t> strides((size_t) t.ndim);
-        for (int64_t i = t.ndim - 1, accum = 1; i >= 0; --i) {
-            strides[i] = accum;
+        int64_t accum = 1;
+        for (int32_t i = t.ndim - 1; i >= 0; --i) {
+            strides[(size_t) i] = accum;
             accum *= t.shape[i];
         }
         t.strides = strides.release();
@@ -1199,7 +1200,7 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
             PyObject *export_fn = ndarray_export_fn(
                 internals_, copy ? nd_export_numpy_copy : nd_export_numpy_view);
             PyObject *stack[] = {nullptr, o.ptr()};
-            Py_ssize_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
+            size_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
             return PyObject_Vectorcall(export_fn, stack + 1, nargsf, nullptr);
         } catch (const std::exception &e) {
             PyErr_Format(PyExc_TypeError,
@@ -1224,7 +1225,7 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
         if (slot != nd_export_count) {
             PyObject *export_fn = ndarray_export_fn(internals_, slot);
             PyObject *stack[] = {nullptr, o.ptr()};
-            Py_ssize_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
+            size_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
             o = steal(PyObject_Vectorcall(export_fn, stack + 1, nargsf, nullptr));
             if (!o.is_valid())
                 return nullptr;
--- a/src/nb_type.cpp
+++ b/src/nb_type.cpp
@@ -670,13 +670,13 @@ static int nb_type_init(PyObject *self, PyObject *args, PyObject *kwds) {
 
     *t = *t_b;
     t->flags |=  (uint32_t) type_flags::is_python_type;
-    t->flags &= ~((uint32_t) type_flags::has_implicit_conversions);
+    t->flags &= (~(uint32_t) type_flags::has_implicit_conversions) & 0xFFFFFF;
 
     // A Python subclass is always a GC heap type
-    t->flags |= (uint32_t) type_flags::has_gc;
+    t->flags |= ((uint32_t) type_flags::has_gc) & 0xFFFFFF;
 
     // Sublclasses do not inherit the pooling feature as a consequence
-    t->flags &= ~((uint32_t) type_flags::pooled);
+    t->flags &= ~((uint32_t) type_flags::pooled) & 0xFFFFFF;
     t->pool_capacity = 0;
 #if defined(NB_FREE_THREADED)
     t->pool_index = 0;
@@ -970,7 +970,7 @@ static PyObject *nb_type_from_metaclass(PyTypeObject *meta, PyObject *mod,
 
         if (slot == 0) {
             break;
-        } else if (slot * sizeof(nb_slot) <= (int) sizeof(type_slots)) {
+        } else if ((size_t) slot * sizeof(nb_slot) <= sizeof(type_slots)) {
             *(((void **) ht) + type_slots[slot - 1].direct) = ts->pfunc;
         } else {
             PyErr_Format(PyExc_RuntimeError,
@@ -1173,7 +1173,7 @@ PyTypeObject *nb_type_create_metaclass(nb_internals *p,
     int basicsize = -(int) sizeof(type_data),
         itemsize = 0;
 #else
-    int basicsize = (int) (PyType_Type.tp_basicsize + sizeof(type_data)),
+    int basicsize = (int) PyType_Type.tp_basicsize + (int) sizeof(type_data),
         itemsize = (int) PyType_Type.tp_itemsize;
 #endif
 
@@ -1202,7 +1202,7 @@ PyTypeObject *nb_type_create_metaclass(nb_internals *p,
     };
 
     // Workaround because __vectorcalloffset__ does not support Py_RELATIVE_OFFSET
-    members[0].offset = p->type_data_offset + offsetof(type_data, vectorcall);
+    members[0].offset = p->type_data_offset + (Py_ssize_t) offsetof(type_data, vectorcall);
 
     if (NB_DYNAMIC_VERSION < 0x030E0000) {
         slots[4] = { Py_tp_members, (void *) members };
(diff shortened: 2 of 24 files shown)


src/nb_ndarray.cpp
```diff
@@ -240,7 +240,7 @@ static void nb_ndarray_releasebuffer(PyObject *, Py_buffer *view) {
 // This function implements __dlpack__() for a nanobind.nb_ndarray.
 static PyObject *nb_ndarray_dlpack(PyObject *self, PyObject *const *args,
                                    Py_ssize_t nargsf, PyObject *kwnames) {
-    if (PyVectorcall_NARGS(nargsf) != 0) {
+    if (PyVectorcall_NARGS(static_cast<size_t>(nargsf)) != 0) {
```
hpkfft: Hmmm.  I would say this is a bug in python rather than in nanobind.
See: https://docs.python.org/3/c-api/structures.html#c.PyCFunctionFastWithKeywords
So, why does [PyVectorcall_NARGS](https://docs.python.org/3/c-api/call.html#c.PyVectorcall_NARGS) not take `Py_ssize_t`?

tkhyn: Maybe, but it doesn't look like PyVectorcall_NARGS taking `Py_ssize_t` is inconsistent with its documentation!
You are certainly a lot more familiar with the Python C API than I am though, so maybe you're right about it being a python bug / inconsistency...

hpkfft: Or, is there a bug here in that we should not be using `PyVectorcall_NARGS()` at all?
In other words, the argument to the function should simply be `Py_ssize_t nargs`?
And then the test above should be `if (nargs != 0) {`

tkhyn:
I just found more of them, will push a follow-up commit.

src/nb_ndarray.cpp
tkhyn: I'm not sure I follow everything and this is clearly out of my comfort zone, but after a shallow reading of the API documentation, for `nargsf` to be a `size_t` it looks like the `Py_TPFLAGS_HAVE_VECTORCALL` flag would be needed somewhere? Maybe [line 313](https://github.com/wjakob/nanobind/blob/master/src/nb_ndarray.cpp#L313)?

wjakob:
The nanobind codebase eschews `static_cast` for casting integers. Please use C-style casts.

tkhyn:
Done, I think they are all dealt with.

src/nb_ndarray.cpp
```diff
@@ -1033,10 +1033,10 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
         if (pkg_name) {
             PyObject* pkg_mod = module_import(pkg_name);
             PyObject* args[] = {pkg_mod, o.ptr()};
-            Py_ssize_t nargsf = 2 | PY_VECTORCALL_ARGUMENTS_OFFSET;
```
hpkfft: I think the correct fix is `size_t nargsf = 2 | PY_VECTORCALL_ARGUMENTS_OFFSET;`

src/nb_ndarray.cpp
```diff
@@ -998,9 +998,9 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
             PyObject* pkg_mod = module_import("numpy");
             PyObject* args[] = {pkg_mod, o.ptr(),
                                 (copy) ? Py_True : Py_False};
-            Py_ssize_t nargsf = 2 | PY_VECTORCALL_ARGUMENTS_OFFSET;
```
hpkfft: I think the correct fix is `size_t nargsf = 2 | PY_VECTORCALL_ARGUMENTS_OFFSET;`

src/nb_ndarray.cpp
```diff
@@ -527,10 +527,10 @@ ndarray_handle *ndarray_import(PyObject *src, const ndarray_config *c,
     } else {
         // Try calling src.__dlpack__()
         PyObject* args[] = {src, static_pyobjects[pyobj_name::dl_version_tpl]};
-        Py_ssize_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;
```
hpkfft, line 630: I think the correct fix is `size_t nargsf = 1 | PY_VECTORCALL_ARGUMENTS_OFFSET;`

src/nb_ndarray.cpp
```diff
@@ -240,7 +240,7 @@ static void nb_ndarray_releasebuffer(PyObject *, Py_buffer *view) {
 // This function implements __dlpack__() for a nanobind.nb_ndarray.
 static PyObject *nb_ndarray_dlpack(PyObject *self, PyObject *const *args,
                                    Py_ssize_t nargsf, PyObject *kwnames) {
```
hpkfft: I think the correct fix is EITHER `Py_ssize_t nargs` above (and don't use `PyVectorcall_NARGS()` below) or `size_t nargsf` above.
I don't know which is right.
@wjakob, do you know the answer (or can you advise the best way to get advice)?
I'm confused as to the applicability of https://docs.python.org/3/c-api/structures.html#c.PyCFunctionFastWithKeywords vs. https://docs.python.org/3/c-api/call.html#c.vectorcallfunc

src/nb_ndarray.cpp
```diff
@@ -777,7 +777,7 @@ ndarray_handle *ndarray_import(PyObject *src, const ndarray_config *c,
 
         scoped_pymalloc<int64_t> strides((size_t) t.ndim);
         for (int64_t i = t.ndim - 1, accum = 1; i >= 0; --i) {
-            strides[i] = accum;
+            strides[(size_t) i] = accum;
             accum *= t.shape[i];
         }
```
hpkfft, line 918: Personally, I would prefer:
```
int64_t accum = 1;
for (int32_t i = t.ndim - 1; i >= 0; --i) {
    strides[i] = accum;
    accum *= t.shape[i];
}
```
I don't know why the compiler complains about `strides[i]` but is OK with `shape[i]`....

tkhyn: Indeed, not sure either. Pushing a better fix for this one anyway (`for (size_t i = (size_t) t.ndim - 1; ...`)

hpkfft: Since `size_t` is unsigned, the test `i >= 0` is always true.

I like my suggestion, since `t.ndim` is already `int32_t` and the values of the strides and shape elements are `int64_t`.  (Also, the compiler should not complain about indexing an array with `int32_t`.)

src/nb_ndarray.cpp
hpkfft: Maybe line 340?
https://github.com/wjakob/nanobind/blob/acc88f3f91d09889c919ca5ea05cf604c4fce102/src/nb_ndarray.cpp#L340

src/nb_ndarray.cpp
tkhyn: Ha, yes indeed. I actually remember noticing that earlier but when I came back to it I had forgotten! Will fix with your suggestion.

However I'll keep the cast to `size_t` for `strides[i]` because `stride` is a `scoped_pymalloc<T>` which subscript operator only accepts `size_t`. `dltensor::shape`, on the other hand, is an `int64_t*` and its subscript operator accepts negative indices, and therefore `int32_t`.

tkhyn:
I've just rebuilt with `-Werror -Wall -Wextra -Wconversion` and found quite a few more - pushing the proposed fixes in a 2nd commit.

hpkfft:
Please take a look at this [FAQ](https://nanobind.readthedocs.io/en/latest/faq.html#policy-on-clang-tidy-wpedantic-etc)

src/nb_ndarray.cpp
hpkfft: Thank you for this (and the next two, which are similar).  It was my mistake not to have used `size_t`.  This fix is simple, matches the Python docs, and does not require any casts, so it's only goodness.

tkhyn:
Thanks and well noted.
No `const_cast` nor `#pragmas` added ;)

src/nb_ndarray.cpp
```diff
@@ -337,7 +337,7 @@ static PyTypeObject *nb_ndarray_tp() noexcept {
             /* .name = */ "nanobind.nb_ndarray",
             /* .basicsize = */ (int) sizeof(nb_ndarray),
             /* .itemsize = */ 0,
-            /* .flags = */ Py_TPFLAGS_DEFAULT,
+            /* .flags = */ Py_TPFLAGS_DEFAULT | Py_TPFLAGS_HAVE_VECTORCALL,
```
hpkfft: I don't fully understand what effect this has.
(I'm commenting here to draw attention to this from @wjakob)

tkhyn: Actually the more I look at it the more I think it's the wrong place. Will try to put more thought into that.

hpkfft: Why are you trying to add `Py_TPFLAGS_HAVE_VECTORCALL`?
Do you think it will help performance?

tkhyn: So, it's indeed in the wrong place as it has nothing to see with the `__dlpack__` / `nb_ndarray_dlpack` method.
If there was a row to modify this was indeed line 313:

https://github.com/wjakob/nanobind/blob/acc88f3f91d09889c919ca5ea05cf604c4fce102/src/nb_ndarray.cpp#L313

However, there is no `METH_*` flag in the [PyMethodDef](https://docs.python.org/3/c-api/structures.html#c.PyMethodDef) documentation that allows the function signature with `size_t nargsf` as for the [vectorcall protocol](https://docs.python.org/3/c-api/call.html#c.vectorcallfunc). The closest matching candidate - unless I missed one - for a method is through the [`METH_FASTCALL | METH_KEYWORDS`](https://docs.python.org/3/c-api/structures.html#meth-fastcall-meth-keywords) flags, which declare a method of type [PyCFunctionFastWithKeywords](https://docs.python.org/3/c-api/structures.html#c.PyCFunctionFastWithKeywords), for which `nargs` is a `Py_ssize_t`.

Therefore, I think the correct approach is the initial one: do not modify `nb_ndarray_dlpack`'s signature and use `PyVectorcall_NARGS((size_t) nargsf)`. Having to resort to a cast isn't ideal but I don't see a prettier solution here.

Now that I'm convinced that adding the Py_TPFLAGS_HAVE_VECTORCALL is wrong, I'll push the change.

tkhyn: > Why are you trying to add `Py_TPFLAGS_HAVE_VECTORCALL`?
> Do you think it will help performance?

Nope, that was related to trying to make `nb_ndarray_dlpack` accept a `size_t` instead of `Py_ssize_t` following our earlier discussion. But this was never going to work as per my comment above.

hpkfft: The `f` in `nargsf` indicates that the argument is both `nargs` and flags.
The type of `nargsf` is `size_t`.
The type of `nargs` is `Py_ssize_t`.
The purpose of `PyVectorcall_NARGS()` is to convert `nargsf` to `nargs`.
If you believe the correct function signature is [PyCFunctionFastWithKeywords](https://docs.python.org/3/c-api/structures.html#c.PyCFunctionFastWithKeywords), for which `nargs` is a `Py_ssize_t`, then the test should simply be `if (nargs != 0) {`

tkhyn: OK, it looks like you have the `PyVectorcall_NARGS()` figured out while I was focusing on the PyMethodDef side 😊
Having a closer look it indeed only removes the flag from a `size_t` to produce a `Py_ssize_t`, and calling `PyVectorcall_NARGS()` with a `Py_ssize_t` could lead to an unexpected bug - that's why I like `-Wconversion`, I've been bitten in the past by similar things!

I agree with your conclusion, we should have `if (nargs != 0)`. Amending.

wjakob:
I finally had time to take a closer look. These changes all look good to me, but could I ask you to rebase on top of `master`?

tkhyn:
Thanks @wjakob, it's done!
Builds with `-Werror -Wall -Wextra -Wconversion -Wno-comment` with no error on Clang 22.
Would you consider a further commit that adds these compile options to the cmake config? I'll perfectly understand if you don't but it would ensure these warnings are caught early rather than in projects that consume `nanobind`.

hpkfft:
I don't like `-Wno-comment`.  Instead, is there a comment in nanobind that can be cleaned up?
Note that stricter flags can be used for testing without affecting other projects: https://github.com/wjakob/nanobind/blob/706131bef6771ad3634b1d772f029f65b7ea8bd2/tests/CMakeLists.txt#L27

tkhyn:
@hpkfft no, nanobind comments don't need a cleanup. I always use `-Wno-comment` as I often find myself using `//` in comments to abbreviate 'parallel' - you're right it's unnecessary here.

And yes, adding `-Werror -Wconversion` only in testing would be the sensible thing to do. Let me know if you're happy with it and I'll push it.

hpkfft:
This is entirely up to @wjakob 
However, you might want to add it as a commit (which can easily be reverted) to see what happens on the various platforms in the github CI pipeline....

tkhyn:
Looks like this highlighted that my job wasn't complete - there are quite a few new warnings in sections of code my local config didn't include.

I think it's worth fixing these too, I'll look into doing that, it might take a few attempts.

wjakob:
I'm fine with adding -Werror -Wconversion in testing

src/nb_type.cpp
```diff
@@ -1624,10 +1624,10 @@ PyObject *nb_type_new(const type_init_data *t) noexcept {
     if (to->flags & (uint32_t) type_flags::pooled) {
 #if defined(PYPY_VERSION)
         // PyPy's cpyext object model is incompatible with park/revive step
-        to->flags &= ~(uint32_t) type_flags::pooled;
+        to->flags &= (~(uint32_t) type_flags::pooled) & 0xFFFFFF;
```
wjakob, line 1627: Not sure I understand those. I prefer if this is done via a cast instead of math.

tkhyn: I wanted to do that too, but to my knowledge casting to a 24-bit unsigned integer isn't supported and it's the only way to tell the compilers (only GCC complains about that one, but arguably the others should) that we are aware that we are converting a 32-bit value to 24-bit.

There may be a way to do it with a cast but it would involve creating a custom 24-bit uint wrapper class for `type_data:flags`. Not sure you'd want to go this route as it will make the code a little more convoluted.

EDIT: ... or make `type_data:flags` a full `uint32_t` but this may have other implications about which I'm probably not knowledgable enough

tkhyn:
All right, we got there after a few tries as predicted ... Sorry for the spam it was quicker to exercise github's CI pipeline than setting up VMs for all test configurations!

A few notes:

- The first fixes have been rebased against the new master and there were merge conflicts so a 2nd look might be useful
- I pushed the new fixes (the ones that came from the CI failures) in a separate commit so that you can look at them separately, but intend to squash them with the first one once the changes are validated
- I disabled `-Werror` on CUDA. There were many visibility attribute warnings such as the following that I feel are out of scope (and that would require a macro so that these visibility attributes are not present when building with MSVC)

   ```
   /__w/nanobind/nanobind/tests/test_callbacks.cpp:64:8: error: 'callback_data' declared with greater visibility than the type of its field 'callback_data::subscribers' [-Werror=attributes]
      64 | struct callback_data {
         |        ^~~~~~~~~~~~~
   ```

- As pointed by @wjakob above, I had to use `& 0xFFFFFF` to tell GCC that we were intentionally converting a `uint32_t` to a 24 bit uint. If anyone has a better idea I'll be glad to learn something new!
- Of note, `nb_internals::type_data_offset` is now a `ptrdiff_t` which made more sense to me than casting to `size_t` elsewhere, but I may have missed something

wjakob:
Thank you!

tkhyn:
Thanks for merging, and thank you for your patience!

PR #1334, opened 2026-05-02
stubgen: simplify nested type names inside enclosing class bodies

### LINKED ISSUE #1333: [BUG]: stubgen produces bad results for nested enums
### Problem description

When a class contains a nested type (e.g. an `nb::enum_` inside an `nb::class_`), stubgen previously emitted the fully-qualified name in type annotations within the class body:

```python
    class EnumWrapper:
        Alpha: EnumWrapper.Value = EnumWrapper.Value.Alpha
```

Type checkers cannot resolve `EnumWrapper.Value` inside the EnumWrapper class body because the class isn't fully defined yet. This produces `Unknown` typing

With this fix, stubgen strips the enclosing class prefix when emitting types that refer to nested members of the current class:

```python
    class EnumWrapper:
        Alpha: Value = Value.Alpha
```

The nested name `Value` is directly visible in the class scope, so type checkers resolve it correctly.

### Reproducible example code

```text
Included example in the linked PR which I think shows a good example (this example is pulled straight from the https://github.com/rust-diplomat/diplomat project's use of nanobind
```

markjm:
When a class contains a nested type (e.g. an `nb::enum_` inside an `nb::class_`), stubgen previously emitted the fully-qualified name in type annotations within the class body:

```python
    class EnumWrapper:
        Alpha: EnumWrapper.Value = EnumWrapper.Value.Alpha
```

Type checkers cannot resolve `EnumWrapper.Value` inside the EnumWrapper class body because the class isn't fully defined yet. This produces `Unknown` typing

With this fix, stubgen strips the enclosing class prefix when emitting types that refer to nested members of the current class:

```python
    class EnumWrapper:
        Alpha: Value = Value.Alpha
```

The nested name `Value` is directly visible in the class scope, so type checkers resolve it correctly.

Note if you care - I used AI to support finding, debugigng, and fixing this issue, but I am a human and here to answer questions or work through feedback!

Fixes #1333

### DIFF
--- a/tests/test_enum_ext.pyi.ref
+++ b/tests/test_enum_ext.pyi.ref
@@ -129,3 +129,23 @@ class Item(enum.Enum):
     extra = 2
 
 def item_to_int(item: Item = Item.name) -> int: ...
+
+class EnumWrapper:
+    def __init__(self, arg: Value, /) -> None: ...
+
+    class Value(enum.Enum):
+        Alpha = 0
+
+        Beta = 1
+
+        Gamma = 2
+
+    Alpha: Value = Value.Alpha
+
+    Beta: Value = Value.Beta
+
+    Gamma: Value = Value.Gamma
+
+    def __eq__(self, arg: Value, /) -> bool: ...
+
+    def get_value(self) -> Value: ...
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -740,7 +740,21 @@ class StubGen:
                 return cls_name if cls_name != "NoneType" else "None"
             if full_name.startswith(self.module.__name__ + "."):
                 # Strip away the module prefix for local classes
-                return full_name[len(self.module.__name__) + 1 :]
+                result = full_name[len(self.module.__name__) + 1 :]
+                # When inside a class body, also strip the enclosing class
+                # prefix so that e.g. "ErrorEnum.Value" becomes just "Value"
+                # when we are already inside class ErrorEnum.
+                scope = self.prefix[len(self.module.__name__) + 1 :]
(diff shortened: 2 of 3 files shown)


tests/test_enum_ext.pyi.ref
```diff
@@ -129,3 +129,23 @@ class Item(enum.Enum):
...
+
+        Gamma = 2
+
+    Alpha: Value = Value.Alpha
+
+    Beta: Value = Value.Beta
+
+    Gamma: Value = Value.Gamma
+
+    def __eq__(self, arg: Value, /) -> bool: ...
```
markjm, line 149: p.s. this signature is pre-existing but also produces an issue at least in `ty`

```
Invalid override of
  method `__eq__`: Definition is incompatible with `object.__eq__`

  info: parameter `value` has an incompatible type: `object` is not assignable to `Value`
  info: This violates the Liskov Substitution Principle
  help: It is recommended for `__eq__` to work with arbitrary objects, for example:
  help:
  help:     def __eq__(self, other: object) -> bool:
  help:         if not isinstance(other, EnumWrapper):
  help:             return False
  help:         return <logic to compare two `EnumWrapper` instances>
  help: tyinvalid-method-override
  builtins.pyi(142, 9): `object.__eq__` defined here
```

happy to fix it in a seperate fix/change

markjm:
Hi @wjakob - is this a fix you would be interested in landing?

wjakob:
Thanks, merged with a followup fix: 12c3e6d7867ce29e4d8f3f48369372d28cc5a33a

PR #1338, opened 2026-05-08
Support Metal DLPack ndarray imports and byte offsets

XXXXRT666:
## Summary

This adds test coverage for importing Metal-backed DLPack arrays into `nb::ndarray`, including PyTorch MPS tensors.

It also exposes raw DLPack storage metadata for device-backed arrays:

- `data_handle()`
- `byte_offset()`

`nb::ndarray` exports can now set `DLTensor::byte_offset` explicitly via a new trailing `byte_offset` constructor argument. This lets producers represent views into opaque device allocations, such as Metal `id<MTLBuffer>` handles, without copying when the logical array starts after the beginning of the allocation.

## Tests

- Added a mocked Metal DLPack import test.
- Added a real PyTorch MPS DLPack import test guarded by `needs_torch_mps`.
- Added CPU and Metal DLPack export coverage for nonzero `byte_offset`.


src/nb_ndarray.cpp
```diff
@@ -525,21 +525,84 @@ ndarray_handle *ndarray_import(PyObject *src, const ndarray_config *c,
     if (src_is_pycapsule) {
         capsule = borrow(src);
     } else {
+        object dl_device;
```
hpkfft: This extra code will slow down the typical usage, which is importing from CPU.

Note that calling [__dlpack__()](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#api-specification-generated-array-api-array-dlpack--page-root) without specifying a device has a default of `None`, which means the exported capsule should be on the same device as self is.  This, which is done now, seems like the right thing for nanobind to do by default.

wjakob: I am also curious about the context. what's the reason for all of this extra code on a hot path?

XXXXRT666: Thanks, that makes sense. I had interpreted the explicit device constraint as a reason to request the device via `dl_device`, but the default `__dlpack__()` behavior already preserves the producer's device and nanobind validates `DLTensor.device` afterwards.

I removed the extra logic. The PR now only adds raw DLPack handle/offset accessors plus MPS test coverage.

XXXXRT666:
Added explicit DLPack byte_offset export support: `nb::ndarray` constructors now accept a trailing data_offset, with CPU/Metal tests covering nonzero offsets.

src/nb_ndarray.cpp
hpkfft: Please just remove all of this (including the comment).  It's needlessly confusing.

hpkfft: Please change this to `mt->dltensor.data = view->buf;`

hpkfft: Please change this to `mt->dltensor.byte_offset = 0UL;`

src/nb_ndarray.cpp
```diff
@@ -863,9 +865,8 @@ ndarray_handle *ndarray_create(void *data, size_t ndim, const size_t *shape_in,
     uintptr_t data_uint = (uintptr_t) data;
     data = (void *) (data_uint & ~uintptr_t{255});      // upper bits
     uint64_t data_offset = data_uint & uintptr_t{255};  // lowest 8 bits
-#else
-    constexpr uint64_t data_offset = 0UL;
 #endif
```
hpkfft: Please just remove this code.  It's incorrect and needlessly confusing.

src/nb_ndarray.cpp
hpkfft: I'm sorry this is a bit of a mess.  Probably, it's best to consider the Python array API standard as normative and ignore the 256B alignment comment in `dlpack.h` as erroneous. The nanobind source code is not the right place to give a full explanation.  I would simplify this comment as follows:

```suggestion
    /* A comment in the DLPack header file suggests 256-byte alignment of the
       DLTensor::data field, but this is generally (and necessarily) ignored.
       Note that the pointer dltensor.data must point to allocated memory
       (i.e., memory that can be accessed), so it cannot simply be rounded
       down by zeroing its lowest 8 bits.
       A byte_offset can be used to support array slicing when data is an
       opaque device pointer or handle, on which arithmetic is impossible.
       See also: https://github.com/data-apis/array-api/discussions/779      */
```

include/nanobind/ndarray.h
```diff
@@ -436,6 +438,8 @@ template <typename... Args> class ndarray {
     bool is_valid() const { return m_handle != nullptr; }
     int device_type() const { return (int) m_dltensor.device.device_type; }
     int device_id() const { return (int) m_dltensor.device.device_id; }
+    void *data_handle() const { return m_dltensor.data; }
+    uint64_t data_offset() const { return m_dltensor.byte_offset; }
```
hpkfft: I would suggest renaming `data_offset` to `byte_offset` everywhere.

XXXXRT666:
Addressed the latest review

hpkfft:
Thanks.

I think it's helpful for module developers that this PR now uses the name `byte_offset` since DLPack is inconsistent--strides are measured in terms of the dtype but the initial offset is measured in terms of bytes.
Twenty years ago, the buffer protocol made the wiser choice of measuring strides in bytes.
Note, for example, in C++ (assuming typical IEEE floating-point types):
+ `sizeof(float) == 4`
+ `sizeof(std::complex<float>) == 8`
+ `alignof(std::complex<float>) == 4`

so it's unnecessarily limiting to measure strides in terms of element size.
See `test41_noninteger_stride()` for an example of a slice that doesn't work with DLPack.

wjakob:
This change looks good to me. Does it have your blessing too @hpkfft ?

hpkfft:
Yes, thank you.
One question I have for you is whether the ABI version needs to be increased since the signature of `detail::ndarray_create()` has changed to have the additional parameter.

wjakob:
This is an ABI-compatible change.
ABI-incompatibility occurs when one instance of libnanobind can no longer parse internal data structures used by another instance of libnanobind. This would happen for changes in nb_internals, nb_func, nb_instance, etc.

wjakob:
Thanks to both of you!

zcbenz:
Can you please create a new git tag for this change so we can update the nanobind in MLX?

wjakob:
@zcbenz @awni @hawkinsp Sorry about the delay. A quick request: can you try running the MLX/MLIR/JAX test suite with the top-of-tree nanobind? So many things changed especially after the latest bug sweep (#1371), it'd be good if somebody else with a big project stress-tests it before I make a release.

XXXXRT666:
> @zcbenz @awni @hawkinsp Sorry about the delay. A quick request: can you try running the MLX/MLIR/JAX test suite with the top-of-tree nanobind? So many things changed especially after the latest bug sweep (#1371), it'd be good if somebody else with a big project stress-tests it before I make a release.

I ran the full CPU/GPU test suite for the DLPack branch against the latest version of nanobind, and all tests passed in my local envs.

XXXXRT666:
image
MLX CI passed

zcbenz:
Thanks @XXXXRT666 for running the tests, I confirm that latest nanobind passes MLX's test suite in CI and locally.

PR #1340, opened 2026-05-09
Fix polymorphic deallocation

WeiPhil:
Hi,
Recently, I ran into an allocator mismatch crash during cleanup (specifically during Python garbage collection) when working with Mitsuba:
1. A dynamic subclass like `PrincipledBsdf` is compiled with `alignas(16)` for SIMD operations.
2. But its memory lifetime is handed over to Python as a raw, base `Bsdf*` pointer (so that Python takes over and deallocates it).

Since the current deallocator in nanobind does placement destructors + raw `operator delete(void*)` on base pointers, standard C++ dynamic delete routing seems to be bypassed during cleanup. On cleanup, this unfortunately crashes with a mismatched free in our internal build configuration where the default compiler alignment is 8.

I've worked on a fix that has been working for us internally as it reverts to standard type-safe deallocations when classes have virtual destructors. 

To be honest, I'm not 100% sure this is the best fix for this niche problem, but it was the only way I could find to keep the allocators happy.

Let me know what you think or if you have a better alternative!
Best,
Philippe

### DIFF
--- a/include/nanobind/nb_class.h
+++ b/include/nanobind/nb_class.h
@@ -122,6 +122,7 @@ struct type_data {
     void (*destruct)(void *);
     void (*copy)(void *, const void *);
     void (*move)(void *, void *) noexcept;
+    void (*dealloc)(void *) noexcept = nullptr;
     union {
         implicit_t implicit;  // for C++ type bindings
         enum_tbl_t enum_tbl;  // for enumerations
@@ -248,6 +249,10 @@ template <typename T> void wrap_move(void *dst, void *src) noexcept {
 
 template <typename T> void wrap_destruct(void *value) noexcept {
     ((T *) value)->~T();
+}
+
+template <typename T> void wrap_delete(void *value) noexcept {
+    delete (T *) value;
 }
 
 template <typename, template <typename, typename> typename, typename...>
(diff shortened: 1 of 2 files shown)


wjakob:
Hi Philippe,

a few thoughts on this: Nanobind actually explicitly supports _over-aligned_ types (for which alignment >= `__STDCPP_DEFAULT_NEW_ALIGNMENT__`). 

In such a case:

1. It ensures the correct alignment when allocating objects on the Python heap.
2. When given ownership of an existing C++ allocation made by `operator new`, it will call the matching `operator delete` that takes an `align_val_t`.

This functionality is tested in the nanobind test suite (e.g. `struct alignas(128) BigAligned` in `test_classes.cpp`).

What it does _not_ support are custom `new` / `delete` operators. This is explicitly listed in the statute of things that are out of scope.

It sounds to me like you have hit such a case. If nanobind deletes an instance of a abstract base class and then needs to figure out that the actual polymorphic object (in disagreement with the base) suddenly wants a sized deallocation, then that is too complicated for nanobind. Supporting this would have a cost on the fast path, which this project isn't willing to pay as a general policy. The solution is to make your base class over-aligned too. The general rule for such corner cases is "the application has to adapt to the features of nanobind, instead of the other way around".

WeiPhil:
Hi Wenzel,

This is also what I feared. We'll have to change the base class alignment internally because of our custom new/delete. Thanks for the confirmation!

Best,
Philippe

PR #1347, opened 2026-05-28
stubgen: Allow module imports in patterns with a new directive

nicholasjng:
Introduces the keyword `\import` to import a whole module used in a pattern file. This is necessary when a type hint introduced in the pattern file is the only usage of that particular module, and the module itself has to be imported instead of a member.

---------------------

Hey Wenzel, I'm currently writing bindings for duckDB, and I resorted to using a pattern file to handle the following case:

```python
def foo(self, a, b: bool = False) -> mod.A | mod.B: ...
```

where type `mod.A` is returned if `b is True`, and `mod.B` otherwise. I'm applying the following overload pattern:

```
foo:
    @overload
    def foo(self, a, b: Literal[True]) -> mod.A: ...
    @overload
    def foo(self, a, b: Literal[False] = False) -> mod.B: ...
    def foo(self, a, b: bool = False) -> mod.A | mod.B: ...
        \doc
```

Unfortunately, if `foo` is the only place in the stub where `mod` is used, `mod` will not be imported at the top, since `simplify_types` is not called on patterns. There is also no pre-existing pattern directive to import entire modules, with the  alternative being `\from mod import A, B`. Until now!

```
foo:
    \import mod [as mymod]  # optional as_name.
    @overload
    def foo(self, a, b: Literal[True]) -> mod.A: ...
    @overload
    def foo(self, a, b: Literal[False] = False) -> mod.B: ...
    def foo(self, a, b: bool = False) -> mod.A | mod.B: ...
        \doc
```

If this is acceptable for you, I will add the missing documentation to the pattern file docs. Thanks!

### DIFF
--- a/docs/typing.rst
+++ b/docs/typing.rst
@@ -714,6 +714,18 @@ you may use the special ``\from`` escape code to import them:
        def lookup(array: Array[T], index: Literal[0] = 0) -> _Opt[T]:
            \doc
 
+If a replacement rule instead refers to a *whole module* (for example because
+the only use of that module in the stub is a qualified attribute access),
+use the ``\import`` escape code. It mirrors Python's ``import`` statement and
+accepts a comma-separated list of modules, with an optional ``as`` alias:
+
+.. code-block:: text
+
+   my_ext.import_me:
+       \import logging
+       \import collections.abc as cabc
+       def import_me(arg: logging.Logger) -> cabc.Iterable[int]: ...
+
 You may also add free-form text the beginning or the end of the generated stub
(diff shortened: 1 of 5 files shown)


wjakob:
Hi Nicholas -- sorry for the delay. Yes, this looks useful. I will merge it once documented.

wjakob:
(and tested)

nicholasjng:
Thank you, I just added a test and a note in the documentation.

wjakob:
Thanks!

PR #1351, opened 2026-05-29
cmake: report user error if Python Interpreter is missing

### LINKED ISSUE #1350: [BUG]: add_module names .so after some python version, not the one compiled against if EXACT is specified when searching Python
### Problem description

I'm currently using master branch (2deac96697d1b304f3c973cef7de5f94cbad5a57) with cmake version 4.3.2 and noticed that the .so generated gets named after *a* python version found on the system, not necessarily the one it was compiled against.

### Reproducible example code

Install two versions of Python, in my case 3.13 and 3.14 and use the following CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.19)

project(buggy)

# notice the EXACT and the missing "Interpreter" component

find_package(Python 3.13 EXACT COMPONENTS Development.Module REQUIRED)

add_subdirectory(nanobind)

```

Running CMake results in an output like this

```text
-- Found Python: /usr/include/python3.13 (found suitable exact version "3.13.12") found components: Development.Module
-- Found Python: /home/dennis/.local/bin/python3.14 (found suitable version "3.14.3", minimum required is "3.9") found components: Interpreter Development.Module Development.SABIModule
```
It found 3.14 and 3.13 but chose 3.13 since it's the exact match. (The order in which these messages appear does not seem to matter, in my real project 3.14 gets found first, but the issue remains.)

Compiling the project will correctly compile against Python 3.13, but the filename will be something like `my_ext.cpython-314-x86_64-linux-gnu.so`, the 314 should be 313.

hpkfft:
Fixes #1350, in which the module developer calls `find_package(Python)` without `COMPONENTS Interpreter REQUIRED`.
In this case, we will avoid calling `find_package(Python)` ourselves and instead cause a cmake fatal error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ endif()
 # Find the Python interpreter and development libraries
 # ---------------------------------------------------------------------------
 
-if (NOT TARGET Python::Module OR NOT TARGET Python::Interpreter)
+if (NOT TARGET Python::Module AND NOT TARGET Python::Interpreter)
   set(Python_FIND_FRAMEWORK LAST) # Prefer Brew/Conda to Apple framework python
 
   if (CMAKE_VERSION VERSION_LESS 3.18)
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -1,6 +1,6 @@
 include_guard(GLOBAL)
 
-if (NOT TARGET Python::Module)
+if (NOT TARGET Python::Interpreter OR NOT TARGET Python::Module)
   message(FATAL_ERROR "You must invoke 'find_package(Python COMPONENTS Interpreter Development REQUIRED)' prior to including nanobind.")
 endif()
 


wjakob:
What's the status of this PR following the discussion in https://github.com/wjakob/nanobind/issues/1350? Merge? Close?

hpkfft:
I slightly suggest this should be merged.  It's a little more developer friendly for cmake to cause an error if `find_package(Python ...)` has been called with only one of `Interpreter` or `Development.Module` before calling `add_subdirectory(nanobind)`.

If `find_package(Python ...)` is called after nanobind's cmake files do their thing, then there's nothing nanobind can do to catch a developer error.  For example, the developer can change the version of the `Development.Module` using `find_package()`'s EXACT flag without changing the version of `Interpreter`.  That's the developer's bug.

I think the docs are OK.  The step-by-step repeats the phrase, "append the following lines at the end of CMakeLists.txt".  I don't think it's necessary for nanobind to document what might happen if cmake commands are misordered.

Then I'd say #1350 can be closed as "fixed".

It's possible that someone is purposefully calling `find_package(Python ...)` with only one of `Interpreter` or `Development.Module` and relying on nanobind's current behavior of its calling `find_package(Python ...)` to get the missing component.  This PR will break their build, and they'll have to add the missing component to their `find_package(Python ...)` (or else remove their call entirely).  I'd say that's a good thing for them to do anyway.

wjakob:
Thanks!

PR #1355, opened 2026-06-01
security: pin all GitHub Actions to commit SHAs

andife:
## Summary

Pin every `uses:` reference in `.github/workflows/ci.yml` and
`.github/workflows/nvcc-win.yml` from mutable version tags to immutable
commit SHAs, while keeping the human-readable version as a trailing comment.

## Why mutable tags are a supply-chain risk

GitHub Actions run with the repository's full token permissions. A `uses:` line
that references a mutable ref — a floating tag like `@v4` or `@latest`, or a
branch name — resolves to whatever commit that ref points to **at the time the
workflow runs**, not at the time it was written.

This means an action author (or an attacker who has compromised an action
author's account) can push new code to an existing tag, and every repository
that references that tag will silently execute the new code on the next CI run —
with access to `GITHUB_TOKEN`, repository secrets, and the runner filesystem.

**This attack has happened in production.** The
[tj-actions/changed-files incident (March 2025)](https://github.com/advisories/GHSA-mrrh-fwg8-r2c3)
compromised the `@v35` tag and exfiltrated secrets from thousands of CI runs
before it was detected. The
[reviewdog/action-setup compromise](https://github.com/advisories/GHSA-qmg3-hpqr-gqvc)
followed the same pattern days later.

A commit SHA, by contrast, is cryptographically bound to the exact tree of
code that was reviewed. Pushing new commits to a tag does not change what a
SHA-pinned workflow executes.

## What changed

| Action | Before | After |
|--------|--------|-------|
| `actions/checkout` | `@v4` | `@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4` |
| `actions/setup-python` | `@v5` | `@a26af69be951a213d495a4c3e4e4022e16d87065 # v5` |
| `actions/cache` | `@v4` | `@0057852bfaa89a56745cba8c7296529d2fc39830 # v4` |
| `lukka/get-cmake` | `@latest` | `@591817e96fcad43505fb4eae36172462abb3a42e # v4.3.3` |
| `deadsnakes/action` | `@v3.1.0` | `@6c8b9b82fe0b4344f4b98f2775fcc395df45e494 # v3.1.0` |
| `msys2/setup-msys2` | `@v2` | `@e9898307ac31d1a803454791be09ab9973336e1c # v2` |
| `Jimver/cuda-toolkit` | `@v0.2.17` | `@dc581bec6470cf161025608f13d71b3c00c2c0b7 # v0.2.17` |

`lukka/get-cmake@latest` was the highest-risk reference: `@latest` is a
branch-like pointer that resolves to the tip of whatever the author decides,
with no version semantics at all. It is now pinned to the `v4.3.3` release commit.

## Keeping pins up to date

SHA pins create a maintenance burden if not automated. A companion PR
[#1354](https://github.com/wjakob/nanobind/pull/1354) configures
Dependabot to open monthly PRs that update both the SHA and the version
comment in lockstep, so the pins stay current without manual work.

## No functional change

All SHAs correspond exactly to the same releases previously referenced by the
mutable tags. CI behaviour is unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,18 +27,18 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       with:
         submodules: true
 
     - name: Setup Python ${{ matrix.python }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
       with:
         python-version: ${{ matrix.python }}
         cache: 'pip'
 
     - name: Install the latest CMake
-      uses: lukka/get-cmake@latest
+      uses: lukka/get-cmake@591817e96fcad43505fb4eae36172462abb3a42e # v4.3.3
 
     - name: Install Eigen
       if: matrix.os == 'ubuntu-latest'
@@ -86,7 +86,7 @@ jobs:
     - name: Install dependencies
       run: apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y cmake git python3-dev python3-pytest python3-pip libeigen3-dev python3-typing-extensions python3-numpy
 
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       with:
         submodules: true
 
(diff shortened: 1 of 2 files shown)


wjakob:
I have some objections to this PR:

- Pinning the actions to a commit hash is obviously inconvenient since they then will be stale and eventually cause issues.
- There are no secrets in this repository. nanobind isn't published through GHA.

While this policy may be appropriate in other settings, I don't think it's the right one here.

PR #1358, opened 2026-06-01
Stubgen: Always preserve private typevars

dvicini:
Stubgen drops private variables (as identified by the `_` prefix) by default.

However, I think this is suboptimal when a `TypeVar` is used for a public interface. For example, consider the following:

```python
_InputT = TypeVar('_InputT', int, float)

def scale(x: _InputT) -> _InputT:
  return 2 * x
```

The public signature here uses the private TypeVar to indicate that the input and output type of the `scale` function are identical. This example is inspired by the real use in Dr.Jit: https://github.com/mitsuba-renderer/drjit/blob/master/drjit/nn.py#L32. 

Currently, stubgen will drop `_InputT`, leading to an error for type checking tools that then cannot find `_InputT`.

The current stubgen implementation already ensures that type aliases are always preserved. Therefore, preserving `TypeVar`s also seems to make sense.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -900,12 +900,15 @@ class StubGen:
                 and (value.__name__ != name or value.__module__ != self.module.__name__)
             )
 
+            tp = type(value)
+
             # Ignore private members unless the user requests their inclusion
             if (
                 not self.include_private
                 and name
                 and not is_type_alias
                 and len(name) > 2
+                and not self.is_type_var(tp)
                 and (
                     (name[0] == "_" and name[1] != "_")
                     or (name[-1] == "_" and name[-2] != "_")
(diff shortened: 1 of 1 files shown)


wjakob:
Hi Delio -- this fixes this specific case. However, I am wondering if a more general criterion might be better (e.g. don't suppress a name if used as a type annotation within the same module)

dvicini:
I thought that maybe just adding the check for type vars would make sense, as it's consistent with the existing check for type aliases.

It's in principle possible to generalize this, but I am not entirely convinced it is worth the extra complexity? I've update the PR, but I am not sure this approach is a good idea either.

Maybe it's also ok to treat this as a special case and expect users to just do this with the pattern files.

wjakob:
Hi Delio -- I've thought about this, and the architectural cost of doing it properly is too big. Can you reset your PR to the first commit in the sequence and rebase on master?

dvicini:
Done, removed the second commit and rebased

wjakob:
Thanks!

PR #1361, opened 2026-06-08
Add support for binding C++ enums as `StrEnum`

nicholasjng:
An `nb::enum_` annotated with `is_str_enum()` produces a `enum.StrEnum` on Python 3.11, and an `enum.Enum` with a `(str, Enum)` MRO in Python <3.11.

Adds a new `enum_::str_value()` API that takes a `const char*` name along with the enumeration value to be bound, such that the value is bound under the given name on the Python enum.

Also patch stubgen to put enum values by repr, otherwise enum string values are not quoted in the stub, which is an error.

----------------

As mentioned in https://github.com/wjakob/nanobind/discussions/1360.

Disclosure: I used Claude to help in the implementation, mostly the `enum_append_str` and `enum_from_python` bits.

### DIFF
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -2253,7 +2253,7 @@ The following annotations can be specified using the variable-length
    whether or not the enumeration is also marked to support arithmetic
    operations (see :cpp:class:`is_arithmetic`).
 
-.. cpp:struct:: is_str_enum
+.. cpp:struct:: is_str
 
    Indicate that the enumeration carries a string value per entry. Passing
    this annotation changes the Python enumeration parent class to
@@ -2695,7 +2695,7 @@ Class binding
       Add the entry `value` to a string-valued enumeration using the identifier
       `name` and the Python-side string `str_val`, optionally with a docstring
       provided via `doc`. The enumeration must have been declared with the
-      :cpp:class:`is_str_enum` annotation.
+      :cpp:class:`is_str` annotation.
 
    .. cpp:function:: enum_ &export_values()
 
--- a/docs/api_core.rst
+++ b/docs/api_core.rst
@@ -2253,6 +2253,23 @@ The following annotations can be specified using the variable-length
    whether or not the enumeration is also marked to support arithmetic
    operations (see :cpp:class:`is_arithmetic`).
 
+.. cpp:struct:: is_str_enum
+
+   Indicate that the enumeration carries a string value per entry. Passing
+   this annotation changes the Python enumeration parent class to
+   :py:class:`enum.StrEnum` (Python 3.11+) or an equivalent class derived from
+   ``(str, enum.Enum)`` on older Python versions. Each enumerator must then be
+   registered using :cpp:func:`enum_::str_value` instead of
+   :cpp:func:`enum_::value`, supplying the string that represents the entry in
+   Python.
+
+   Members are instances of :py:class:`str` and compare equal to their string
+   value (e.g., ``Color.Red == "red"``). Conversion of bare strings to the
+   enumeration is supported when implicit conversions are enabled.
+
+   This annotation cannot be combined with :cpp:class:`is_arithmetic` or
+   :cpp:class:`is_flag`.
+
 Function binding
 ----------------
 
@@ -2672,6 +2689,13 @@ Class binding
 
       Add the entry `value` to the enumeration using the identifier `name`,
       potentially with a docstring provided via `doc` (optional).
+
+   .. cpp:function:: enum_ &str_value(const char * name, T value, const char * str_val, const char * doc = nullptr)
+
+      Add the entry `value` to a string-valued enumeration using the identifier
+      `name` and the Python-side string `str_val`, optionally with a docstring
+      provided via `doc`. The enumeration must have been declared with the
+      :cpp:class:`is_str_enum` annotation.
 
    .. cpp:function:: enum_ &export_values()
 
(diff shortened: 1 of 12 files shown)


include/nanobind/nb_attr.h
```diff
@@ -119,6 +119,7 @@ struct is_implicit {};
 struct is_operator {};
 struct is_arithmetic {};
 struct is_flag {};
+struct is_str_enum {};
```
hpkfft: The other tags don't have the suffix `_enum`.
Perhaps consider `is_str` or `is_string`?

nicholasjng: I considered `is_str` too, but wasn't convinced, since the name is not very expressive. But maybe I was too critical. Happy to go with either (or `is_str_valued`?).

hpkfft: Naming is difficult.  I'm happy with whatever you and Wenzel decide.  ;)

wjakob: I think I would prefer `is_str`. I don't think there is a big potential for clashes here.

docs/api_core.rst
```diff
@@ -2658,6 +2675,13 @@ Class binding
       Add the entry `value` to the enumeration using the identifier `name`,
       potentially with a docstring provided via `doc` (optional).
 
+   .. cpp:function:: enum_ &str_value(const char * name, T value, const char * str_val, const char * doc = nullptr)
```
wjakob, line 2693: Why can't we use the same .value() API and have it handle strings correctly here? Seems weird to introduce an altogether different API.

src/nb_enum.cpp
```diff
@@ -62,18 +67,39 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
...
+    }
 
     scope.attr(name) = result;
     result.attr("__doc__") = ed->docstr ? str(ed->docstr) : none();
 
-    result.attr("__str__") = enum_mod.attr(is_flag ? factory_name : "Enum").attr("__str__");
-    result.attr("__repr__") = result.attr("__str__");
+    if (is_str_enum) {
+        // Match stdlib StrEnum: str(member) returns the member's string value, not Enum.NAME.
+        result.attr("__str__") = str_tp.attr("__str__");
```
wjakob: nanobind enums deviate from base enum behavior by implementing `str(Enum.Item) == "Enum.Item"`. Actually even Python hasn't been very consistent across versions. But at this point this is how it is, and we should not carve out exceptions for `StrEnum`.

src/stubgen.py
```diff
@@ -640,7 +640,7 @@ def put_value(self, value: object, name: str, parent: Optional[object] = None, a
 
         if isinstance(parent, type) and issubclass(tp, parent):
             # This is an entry of an enumeration
-            self.write_ln(f"{name} = {typing.cast(enum.Enum, value)._value_}")
+            self.write_ln(f"{name} = {typing.cast(enum.Enum, value)._value_!r}")
```
wjakob, line 643: Why is this needed?

docs/api_core.rst
wjakob: (Never mind.)

src/nb_enum.cpp
```diff
@@ -184,6 +210,58 @@ void enum_append(PyObject *tp_, const char *name_, int64_t value_,
     rev->emplace((int64_t) (uintptr_t) el.ptr(), value_);
 }
 
+void enum_append_str(PyObject *tp_, const char *name_, int64_t value_,
```
wjakob: the duplication is unfortunate. Would it be possible to somehow merge with the regular enum_append?

src/stubgen.py
nicholasjng: Without requesting the repr, stubgen would write a `StrEnum` member value (i.e., a string) without quotes, which Python would then interpret as a variable name - that's not intended. The repr outputs the string with quotes, such that it is written to the stub as a string value. It boils down to this:

```python
➜ python3
Python 3.14.5 (main, May 10 2026, 10:21:34) [Clang 21.0.0 (clang-2100.0.123.102)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = 0
>>> repr(a)
'0'   # <- results in A = 0, same as before.
>>> b = "0"
>>> repr(b)
"'0'"  # <- on a StrEnum, becomes A = "0"
```

It was fine like this because until now, all enum values were written as int literals, which cannot be identifiers by definition.

include/nanobind/nb_attr.h
nicholasjng: Renamed to `is_str`.

src/nb_enum.cpp
nicholasjng: Reverted, and updated tests.

src/nb_enum.cpp
nicholasjng: Done.

nicholasjng:
I rebased on your Fable 5 sweep (where I was, embarrassingly, responsible for the "headliner" in-place accessor ref leaks story, sorry about that!), let me know if there's anything else I can do.

src/stubgen.py
wjakob: gotcha!

src/nb_enum.cpp
wjakob: thanks!

wjakob:
Merged, thank you! No worries about the bug, I was responsible for most of the other ones 😅

PR #1366, opened 2026-06-09
``nb::pooled()``: per-type instance pool to accelerate object creation

wjakob:
Some nanobind workloads create large numbers of short-lived objects and bottleneck on object allocation, which involves:

1. Allocating the object via ``PyObject_Malloc()``
2. Increasing the reference count of both the object and its heap type
3. Registering the instance in nanobind's instance hash table
4. Calling the C++ constructor

Releasing the instance involves the inverse of each step in reverse order. In free-threaded builds, the data structure in step 3 is furthermore guarded by a sharded lock that is relatively costly to acquire/release.

This commit introduces an optional object pool to reduce these costs. To use it, pass the new ``nb::pooled()`` class binding attribute:

```cpp
nb::class_<MyClass>(m, "MyClass", nb::pooled(128));
```

The implementation then stashes released objects in per-type pool where they can be cheaply reused, which removes steps 1-3 from the critical path (only the placement constructor/destructor calls remain). Incidentally, CPython internally also pools dicts/lists/tuple instances to improve performance.

In free-threaded builds, each thread gets its own pool, which makes object allocation lock-free.

A microbenchmark that exercises object construction becomes 1.42x faster on Python 3.14. On free-threaded Python, the improvement ranges from 1.32x (uncontended) to 3.2x when using 32 threads a 32-core machine.


wjakob:
PS: This feature is helpful for bindings that are constructed at a very high frequency. My motivating use case is a library with tensor objects that are merely pointers into a traced computation graph (i.e., whose construction doesn't do any real computation at that point in time). This speeds up tracing and helps and keep the GPU fed with work. 

cc @hawkinsp @awni in case this applies to JAX/MLX/..

hawkinsp:
> PS: This feature is helpful for bindings that are constructed at a very high frequency. My motivating use case is a library with tensor objects that are merely pointers into a traced computation graph (i.e., whose construction doesn't do any real computation at that point in time). This speeds up tracing and helps and keep the GPU fed with work.
> 
> cc @hawkinsp @awni in case this applies to JAX/MLX/..

This is very interesting, given that I'm spending considerable amounts of my personal time optimizing JAX trace times, and with Python, I agree that half the game is "allocate less". Thanks for the pointer!

wjakob:
> This is very interesting, given that I'm spending considerable amounts of my personal time optimizing JAX trace times, and with Python, I agree that half the game is "allocate less". Thanks for the pointer!

That's great to hear, please let me know how it goes. I will push a few more touchups to this PR, it should be ready in a few hours.

wjakob:
All done!

PR #1367, opened 2026-06-09
Skip refcounting of immortal objects on free-threaded builds

wjakob:
On free-threaded builds, nanobind immortalizes type objects, enums, and function objects. Reference counting operations on immortal objects are already cheap and uncontended, but they are not entirely free. Since the target is known to be immortal, these operations can be skipped outright.

Introduce `NB_{INC,DEC}REF_{TYPE,ENUM,FUNC}` and `NB_CLEAR_FUNC` macros that compile to no-ops when `Py_GIL_DISABLED` is set, and forward to the usual `Py_{INC,DEC}REF/Py_CLEAR` otherwise. Replace the relevant refcount calls on type/enum/function objects in `nb_type.cpp`, `nb_enum.cpp`, and `nb_func.cpp` with these macros.

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -30,7 +30,7 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
                              "nanobind: type '%s' was already registered!\n",
                              ed->name);
             PyObject *tp = (PyObject *) it->second->type_py;
-            Py_INCREF(tp);
+            NB_INCREF_ENUM(tp);
             return tp;
         }
     }
--- a/src/nb_ft.h
+++ b/src/nb_ft.h
@@ -9,6 +9,28 @@
 
 #pragma once
 
+/* Nanobind immortalizes type objects, enums, and function objects on FT builds.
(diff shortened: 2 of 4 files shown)


hpkfft:
Thanks for this work!

Just FYI, I've been applying #1199 locally.  If there are changes I can make such that the PR would be suitable for merging, I'm happy to do so.

wjakob:
@hpkfft I took another look at this commit. I am not opposed to the core idea. If `Py_RETURN_TRUE` is more optimized on some Python versions then we should just use that. But I don't like hardcoding such Python implementation details within nanobind:
```
#if PY_VERSION_HEX < 0x030C0000
        : object(value ? Py_True : Py_False, detail::borrow_t{}) { }
#else
```
In the case of #1367 that made sense to do because we created those objects ourselves.

wjakob:
Maybe you could make a new version of this commit that relies on internal inline wrapper functions that call Py_RETURN_{TRUE,FALSE,NONE}?

hpkfft:
Thanks for the feedback!  I'll work on it this coming weekend.

PR #1368, opened 2026-06-09
Skip refcounting of immortal objects on free-threaded builds

wjakob:
On free-threaded builds, nanobind immortalizes type objects, enums, and function objects. Reference counting operations on immortal objects are already cheap and uncontended, but they are not entirely free. Since the target is known to be immortal, these operations can be skipped outright.

Introduce NB_{INC,DEC}REF_{TYPE,ENUM,FUNC} and NB_CLEAR_FUNC macros that compile to no-ops when Py_GIL_DISABLED is set, and forward to the usual Py_{INC,DEC}REF/Py_CLEAR otherwise. Replace the relevant refcount calls on type/enum/function objects in nb_type.cpp, nb_enum.cpp, and nb_func.cpp with these macros.

### DIFF
--- a/src/nb_enum.cpp
+++ b/src/nb_enum.cpp
@@ -30,7 +30,7 @@ PyObject *enum_create(enum_init_data *ed) noexcept {
                              "nanobind: type '%s' was already registered!\n",
                              ed->name);
             PyObject *tp = (PyObject *) it->second->type_py;
-            Py_INCREF(tp);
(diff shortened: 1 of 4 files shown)


hpkfft:
In case it's useful (and I don't know whether that's true), Python 3.15 has added [PyUnstable_SetImmortal](https://docs.python.org/3.15/c-api/object.html#c.PyUnstable_SetImmortal)

PR #1370, opened 2026-06-12
Add a 'medium' function dispatcher for named/default arguments

wjakob:
nanobind previously selected between two call dispatchers: a fast `simple` one (positional-only signatures with up to `NB_MAXARGS_SIMPLE` arguments) and a general `complex` one. Annotating even a single argument with a name, default value, or `nb::none` flag demoted the function to the complex dispatcher, and every subsequent call paid for machinery (keyword matching, `alloca`-based scratch space, etc.) that a plain positional call never uses.

This commit inserts a `medium` tier between the two: it handles overload chains whose arguments may be named, may carry default values or accept `None`, and whose annotations may mutate the argument list (`nb::keep_alive`, `nb::call_policy`), as long as no overload uses `nb::args`/`nb::kwargs` or exceeds `NB_MAXARGS_SIMPLE` arguments. Positional arguments are copied into a small stack array, missing trailing/keyword-only parameters are filled from defaults, and per-argument flags are honored. Calls that actually pass keyword arguments are forwarded to the complex dispatcher through a frameless wrapper that compiles into two tail calls.

A microbenchmark (Python 3.13, AppleClang 17 `-O3`, M-series arm64, 1M-iteration `timeit`, best of 25) measuring a bound `lambda (int, int)` shows that calls now run ~32% faster, essentially matching the simple dispatcher, while kwargs-passing calls are unaffected:

```
add_simple(1, 2)   [simple tier]   12.6 ns   12.2 ns
add_named(1, 2)                    20.2 ns   13.8 ns
add_default(1, 2)                  19.9 ns   13.7 ns
add_default(1)                     18.8 ns   12.5 ns
add_over(1, 2)     [2 overloads]   20.1 ns   13.7 ns
add_named(1, b=2)  [kwargs call]   21.1 ns   21.9 ns
```

Functions using `nb::keep_alive`/`nb::call_policy` benefit as well: they previously forced the complex dispatcher only because it copies the argument vector, which the medium dispatcher also does.

### DIFF
--- a/src/nb_func.cpp
+++ b/src/nb_func.cpp
@@ -37,6 +37,8 @@ static PyObject *nb_func_vectorcall_simple_2(PyObject *, PyObject *const *,
                                              size_t, PyObject *) noexcept;
 static PyObject *nb_func_vectorcall_simple(PyObject *, PyObject *const *,
                                            size_t, PyObject *) noexcept;
+static PyObject *nb_func_vectorcall_medium(PyObject *, PyObject *const *,
+                                           size_t, PyObject *) noexcept;
 static PyObject *nb_func_vectorcall_complex(PyObject *, PyObject *const *,
                                             size_t, PyObject *) noexcept;
 static uint32_t nb_func_render_signature(const func_data *f,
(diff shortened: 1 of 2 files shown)


wjakob:
cc @hpkfft, I think you had wanted a similar feature a while back.

hpkfft:
Yes, thank you.  I had suggested #1207
My use case is accepting ndarrays.  The parameters are always marked `noconvert()` and sometimes there is a default value `none()`.
A classical example from NumPy would be: https://numpy.org/doc/stable/reference/generated/numpy.dot.html

wjakob:
Right. I am assuming my caster subsumes the use cases of yours, correct? IIRC yours had not accepted named arguments, which seemed restrictive.

hpkfft:
Yes, this is great!  I verified the performance for the case important to me (and using IntelLLVM compiler 2026.0.0).
Yeah, my suggested PR had not accepted named arguments because I don't happen to use them here (since naming ndarray parameters as "a" and "b" seems silly).  I'm glad your code accepts them without hurting the performance of functions that are positional-only.  This is awesome!  Thanks again.

wjakob:
@hpkfft Thanks for confirming. I have a few more perf.-related commits incoming. (including the Py_RETURN_NONE change). So no need to make a PR for it.

PR #1371, opened 2026-06-12
Fishing expedition with Claude Fable 5

wjakob:
Anthropic kindly provided me with several months of free access to their models as an OSS maintainer. I've done various bug-fixing passes with their models in the past (leading to occasional true positives) and decided to try the newly released Claude Fable 5 on a whim.

Well, holy shit. It found a lot. Most of it is in the "nice to have" category (error-path hygiene, reference leaks on failure paths, free-threading hardening). But some of it is like really, really bad.

My personal (**non**-)favorites:

---

### Reference leak in accessor in-place operators

Every in-place operation on an accessor (`obj.attr("x") += v`, `obj[k] -= v`, …) leaked the value object that was just stored back.

```cpp
void f(nb::object o) {
    o.attr("x") += nb::int_(1);   // leaks one reference per call
}
```

ehm, WHAT? How come nobody noticed that?

### Dangling shared_ptr from implicit conversion

When an argument was implicitly convertible to a `std::shared_ptr<T>` parameter, the conversion produced a temporary owned by the cleanup list, but the resulting `shared_ptr`'s deleter held the *original* (pre-conversion) object. Once the cleanup list released the temporary at the end of the call, any `shared_ptr` that outlived the function call pointed to freed memory.

```cpp
nb::class_<A>(m, "A").def(nb::init_implicit<B>());
m.def("store", [](std::shared_ptr<A> a) { stored = a; });
```

```python
store(B())   # 'stored' now references a freed temporary -> use-after-free
```

(Such calls now cleanly fail overload resolution, matching the existing behavior of the `unique_ptr` caster.)

### Dangling/non-owning key in `.attr(handle)` and `operator[](handle)`

Object-keyed accessors stored their key as a non-owning `handle`, while the factory functions passed `borrow(key)`, which vanished at the end of the full expression. The "borrow" was therefore completely pointless, and an accessor outliving its key expression referenced a freed object:

```cpp
auto acc = obj[nb::int_(1000)];  // key temporary is destroyed here
nb::object v = acc;              // reads a dangling key
```

### Count locked arguments with default values toward `nargs_locked`

On free-threaded builds, the `nb::arg("x").lock()` annotation asks nanobind to lock an argument while the bound function runs. Due to a missing template specialization, the combination of `.lock()` with a default argument value didn't count as a locked argument — so a function whose locked arguments all carried defaults silently ran **without any locking at all**:

```cpp
m.def("f", &func, nb::arg("x").lock() = 42);
// free-threaded build: 'x' was not locked, despite the annotation
```

### Fix crash when deleting a static property (`del Cls.attr`)

CPython signals attribute deletion by invoking `tp_setattro` with `value == NULL`. The static-property branch of nanobind's metaclass `setattro` dereferenced `value` without a NULL check:

```python
del MyClass.static_prop   # segfault; now raises AttributeError
```

### Fix NULL deref when constructor 'self' is passed as a keyword argument

The dispatcher derived a constructor's implicit `self` from the first *positional* argument only. Since `self` is an ordinary named parameter, it can also be passed by keyword, in which case the post-construction fixup ran on a NULL pointer:

```python
inst = MyType.__new__(MyType)
MyType.__init__(self=inst, value=42)   # segfault
```

---

## Everything else

The remaining ~50 commits fall into a few buckets:

- **Use-after-free / lifetime**: `std::pair`/`std::tuple`/`set` casters reading from already-released temporaries when given generic sequences/iterables; self-aliasing `v.extend(v)` / `m.update(m)` in `bind_vector`/`bind_map`; exporting ownerless ndarrays that would alias freed C++ storage (now refused); Eigen `Map`/`Ref` edge cases.
- **Reference leaks**: type objects of Python subclasses on free-threaded builds, `nb::list` accessors, the framework module during ndarray export, partially-constructed containers when `bind_vector`/`bind_map` constructors throw mid-fill.
- **Free-threading correctness**: torn reads of exception translators, data races on `enable_shared_from_this`'s `weak_this`, lost updates in `intrusive_counter::set_self_py`, a critical section outliving `PyGILState_Release()` in the trampoline, borrowed-reference hazards in `nb::dict::get`, and several paths that ran arbitrary Python while holding the (non-reentrant) internals lock.
- **Error-path robustness**: `python_error::what()` on exceptions with a throwing `__str__` or non-UTF-8 traceback names, formatting un-encodable keyword arguments, types with missing/non-string `__module__`/`__name__` attributes, warnings-as-errors at `PyErr_WarnFormat` call sites, out-of-memory checks.
- **API contract**: `__dlpack__()` silently ignored the Array API `copy`/`dl_device` keyword arguments (a consumer asking for `copy=True` received an alias of the original buffer); buffer-protocol request flags; `nb::cast<T>()` raising `cast_error` instead of leaking the internal `next_overload` exception (which silently re-dispatched to other overloads).
- **Interpreter shutdown**: GIL acquisition after `Py_Finalize` in the `std::function`, `unique_ptr`, and ndarray cleanup paths (now leak instead of aborting the process at exit).
- **Performance**: quadratic `__repr__` for bound containers (~620 ms → ~5 ms at 80k elements), pre-interned strings for hot-path dunder lookups.

Each commit message describes the defect, the fix, and (where the behavior is observable) how it was reproduced; regression tests accompany nearly every change. The findings were model-generated but human-reviewed: a number of proposed fixes did not survive review and were dropped or substantially simplified before landing here.


bjodah:
Thank you for this write-up, and your continued work on nanobind!

hpkfft:
I hope it's OK to comment here on commit 28040d

> copy/dl_device were matched only by pointer identity ... so an equal-but-distinct key spelling bypassed the checks.

> The max_version branches ended the loop with break, so a trailing copy/dl_device kwarg would have gone unexamined.

Yeah, I had a comment that these keyword arguments were ignored.  Equal-but-distinct key spellings were also ignored.  Quite efficiently, one might say.  :)

Of course, I agree it's nicer to throw `BufferError`.  But now I think there's a performance regression in what I believe is the common path, which I'd like to fix.  Consider:
```
>>> a.__dlpack__(max_version=(1,0))
```
[In my opinion, DLPack is poorly designed.  Why a tuple as a kwarg when the minor version is ignored anyway since it's supposed not to matter for compatibility?  But I digress....]

Anyway, the nanobind code will now do the following:
1. Check pointer equality `key == "copy"`
2. Call `RichCompareBool(key, "copy", Py_EQ)`
3. Check pointer equality `key == "dl_device"`
4. Call `RichCompareBool(key, "dl_device", Py_EQ)`
5. Check pointer equality `key == max_version`

This last check will succeed if, for example, it's NumPy making the call to us, since NumPy has interned `max_version`.

I suggest two loops: first do all the pointer equality tests, then (if and only if some key has not been identified) do the `RichCompareBool()` calls.   Also, I think currently a misspelled kwarg like "copie" or "mux_version" is silently ignored.  I guess anything we don't recognize should be `BufferError`.  Please let me know what you think.

wjakob:
> I suggest two loops: first do all the pointer equality tests, then (if and only if some key has not been identified) do the RichCompareBool() calls.

That sounds fine to me—If you can make a PR about that, I will merge it before the next release.

PR #1372, opened 2026-06-12
Avoid ADL when invoking the internal 'concat' helper

Cfretz244:
### Problem

`nanobind::detail::descr<N, Ts...>` carries the bound C++ types `Ts...` as template arguments, so the associated namespaces of a `descr` value (for argument-dependent lookup) include the namespaces of those types. The variadic `detail::concat` was written as a self-recursive *unqualified* call:

```cpp
template <size_t N, typename... Ts, typename... Args>
constexpr auto concat(const descr<N, Ts...> &d, const Args &...args)
    -> decltype(std::declval<descr<N + 2, Ts...>>() + concat(args...)) {
    return d + const_name(", ") + concat(args...);
}
```

The recursive call is resolved at instantiation time using ADL. If a bound type lives in a namespace that *also* declares a function named `concat`, that declaration enters the candidate set. Worse, the arguments at nanobind's internal call sites are typically prvalues (e.g. the results of `type_descr(...)`), so a foreign forwarding-reference overload such as

```cpp
template <typename OutStringType = std::string, typename... Args>
OutStringType concat(Args &&...args);
```

binds rvalue references directly and is a *strictly better match* than nanobind's `const descr<...> &` overloads — partial ordering never gets a chance to prefer the more specialized overload. The result is a hard compile error deep inside nanobind's headers, far from the user code that triggered it:

```
nb_func.h:178: error: invalid operands to binary expression
('descr<2UL - 1>' and 'std::string')
```

This is not hypothetical: the declaration above is the exact shape of `nlohmann::detail::concat` in nlohmann/json. Users binding types declared in `nlohmann::detail` (iterators, SAX interfaces, the `value_t` enum, exception types, ...) — or in any other namespace that declares a `concat` — hit this.

### Fix

Two parts, no behavior change (generated signatures are identical):

1. **`nb_descr.h`**: rewrite the variadic `concat` as a fold over `descr::operator+`, which lives in `nanobind::detail` and cannot be hijacked. This removes the recursive unqualified call (and with it the ADL exposure) entirely:

   ```cpp
   return (d + ... + (const_name(", ") + args));
   ```

2. **Call-site hardening**: parenthesize the remaining internal unqualified calls — `(concat)(...)` / `(concat_maybe)(...)` — in `nb_cast.h`, `nb_func.h`, `ndarray.h`, `stl/function.h`, `stl/pair.h`, and `stl/tuple.h`. A parenthesized callee suppresses ADL, so only nanobind's own overloads are ever considered.

### Test

`tests/test_functions.cpp` now declares a namespace containing a small struct and a `concat` function template mirroring the nlohmann/json declaration, then binds the struct plus a two-parameter function returning `std::pair`, which instantiates the variadic concatenation paths in both `nb_func.h` and `stl/pair.h`. Without the fix, the test extension fails to compile with the error shown above. The Python side checks the bound function round-trips values and renders the expected docstring/signature.

Verified on macOS arm64 (Apple Clang 17, Python 3.12): the full test suite passes, and reverting the header changes while keeping the test reproduces the compile failure.


wjakob:
Looking at the +100 - 30 LOC counter and details, my gut reaction is: no way. It also has the AI slop cardinal sin of verbosely documenting the delta compared to some previous state that is irrelevant after the commit.

Cfretz244:
I can minimize it manually if you want. It was indeed discovered during a session with Claude, as the commit shows; I just want to get the issue resolved. It is a real conflict with `nlohmann::json` 😄

Cfretz244:
The issue this PR was raised to resolve is an argument-dependent-lookup bug for [concat](https://github.com/wjakob/nanobind/blob/master/include/nanobind/nb_descr.h#L125) that conflicts with [this parallel definition inside of `nlohmann::json`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/string_concat.hpp#L137). I didn't actually know this before I hit this bug, but even though the calls to `concat` are `descr` type-wrappers, and not the `nlohmann::json` types themselves, it is _still_ eligible for ADL because [the standard says that template _arguments_ are also part of the ADL set](https://eel.is/c++draft/basic.lookup.argdep#3.3).

> if T is a class template specialization, its associated entities also include ... the types of the template arguments provided for template type parameters

The declaration of `concat` inside of `nlohmann::json` takes its argument by unconstrained forwarding reference, and therefore become the best match, superseding the intended definition inside of `nanobind`. This allows unqualified call-sites such as [this](https://github.com/wjakob/nanobind/blob/master/include/nanobind/nb_func.h#L179), and others, to be hijacked into calling the nlohmann function instead. In a manually verified run in GCC 16.1.0, it spews out:

```text
nanobind/include/nanobind/nb_func.h:178:25: error: no match for 'operator+' (operand types are 'nanobind::detail::descr<1>' and 'std::__cxx11::basic_string<char>')
  178 |         const_name("(") +
      |         ~~~~~~~~~~~~~~~~^
  179 |         concat(type_descr(
      |         ~~~~~~~~~~~~~~~~~~
  180 |             make_caster<remove_opt_mono_t<intrinsic_t<Args>>>::Name)...) +
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
It is trying to find a match between `nanobind::detail::descr<1>` and `std::string` because the `nlohmann::json` function returns `OutStringType`.

I've now manually minimized stuff (I've left the comments inside of the tests as I figured they are useful. I will also remove them as well if you'd like), let me know if there's anything additional you'd like to see in order to get this merged. Thank you! 😄

Cfretz244:
@wjakob is there anything else you'd like to see on this PR? I believe I've manually cleaned it up as requested. If you'd truly prefer to leave a latent ADL bug in the codebase I can close this PR and go away, I was just trying to report an issue I hit while working with a real-world codebase. I don't think the diff can be made much more minimal while still accomplishing its purpose.

It's also possible/likely that there are more bugs of this same shape littered throughout the codebase, I was just fixing the point issue I saw. If you'd like I can try to do a larger investigation and make this more of a holistic patch for the category of ADL hijacking internal functions. Or I can just go away and we close out this PR. Was just trying to contribute and improve usability, did not mean to give offense.

PR #1373, opened 2026-06-14
Improve kwarg parsing in nb_ndarray's __dlpack__()

hpkfft:
This avoids calling `PyObject_RichCompareBool()` on keyword names that will later be recognized by pointer equality.

Using Python 3.15.0b2, NumPy 2.4.4, and
```
python -m timeit -n 10000000 -r 10 -s "import numpy as np, test_ndarray_ext as t; nda = t.ret_array_api()" "np.from_dlpack(nda)"
```
I see over 8% speedup (231 nsec -> 213 nsec).

See also: https://github.com/wjakob/nanobind/pull/1371#issuecomment-4701159570

### DIFF
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -289,25 +289,16 @@ static PyObject *nb_ndarray_dlpack(PyObject *self, PyObject *const *args,
     }
     Py_ssize_t nkwargs = (kwnames) ? NB_TUPLE_GET_SIZE(kwnames) : 0;
 
+    // Match a keyword name against an interned reference.
+    auto key_is = [](PyObject *key, PyObject *r) -> bool {
+        return key == r;
+    };
+
     // Match a keyword name against an interned reference, falling back to a
     // string comparison since kwnames passed via f(**d) are not guaranteed to
     // be identical to the interned objects.
-    auto key_is = [](PyObject *key, PyObject *r) -> bool {
+    auto key_equals = [](PyObject *key, PyObject *r) -> bool {
         return key == r || PyObject_RichCompareBool(key, r, Py_EQ) == 1;
-    };
-
-    // Extract a 2-tuple of integers; returns false (with no error set) for
-    // any other input
-    auto get_int_pair = [](PyObject *value, long *a, long *b) -> bool {
-        if (!PyTuple_Check(value) || NB_TUPLE_GET_SIZE(value) != 2)
-            return false;
-        *a = PyLong_AsLong(NB_TUPLE_GET_ITEM(value, 0));
-        *b = PyLong_AsLong(NB_TUPLE_GET_ITEM(value, 1));
-        if (PyErr_Occurred()) {
-            PyErr_Clear();
-            return false;
-        }
-        return true;
     };
 
     ndarray_handle *th = ((nb_ndarray *) self)->th;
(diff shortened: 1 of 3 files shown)


hpkfft:
+ Accept `stream=None`, but reject other values for `stream`.
+ Raise TypeError on unrecognized keyword arguments.

wjakob:
Thank you!

src/nb_ndarray.cpp
```diff
@@ -289,63 +289,97 @@ static PyObject *nb_ndarray_dlpack(PyObject *self, PyObject *const *args,
...
+                                "max_version must be None or tuple[int, int]");
+                        return -1;
+                    }
+                    max_major_version = a;
                 }
-                max_major_version = a;
+            } else if (compare(key, NB_INTERNED(stream))) {
+                if (value != Py_None) {
+                    PyErr_SetString(PyExc_RuntimeError,
+                            "nanobind only supports stream=None.");
```
Ahajha, line 360: What is the reasoning for this? This broke some code in MAX. I think I can work around it by "swallowing" the `stream` argument (i.e. accepting it but not passing it), but I don't know if that's a good solution.

wjakob: What do you provide as a stream here? IIRC this is the CPU import path

Ahajha: This is PyTorch calling into our `__dlpack__` from https://github.com/pytorch/pytorch/blob/4b0ca472f3e7991ae9fafb629054c6dc00752a66/torch/utils/dlpack.py#L179. Seeing this on both CUDA and ROCm paths.

Ahajha: Specifically the call is here: https://github.com/pytorch/pytorch/blob/4b0ca472f3e7991ae9fafb629054c6dc00752a66/torch/utils/dlpack.py#L186, line 179 is where I assume it's getting the stream arg.

hpkfft: Wouldn't it be best for pytorch not to pass `stream` at all?
I believe the comment in pytorch:
```
    # The array API specify that the default legacy stream must be passed
    # with a value of 1 for CUDA
```
is incorrect.
https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html
If the kwarg is omitted,  it defaults to `None`, which is correct for both CUDA and ROCm (being equivalent to 1 for CUDA and equivalent to 0 for ROCm).
It would be a bit faster to not pass `stream` as a kwarg.

The truth is that nanobind ignores `stream`, so I think it's best we're upfront about only accepting `None`.
It is true that we could add logic that, for example, accepts `stream=1` if the device is CUDA, because that's another way of saying `stream=None`.   But I'd rather not have to keep nanobind up to date with every device's alternate way of specifying the default stream.

Since the data-api specification states

> Support for a `stream` value other than `None` is optional and implementation-dependent.

I would suggest pytorch should prefer to pass `None` (or omit the `stream` kwarg).  What do you think?

hpkfft: In pytorch line 178, if `stream.cuda_stream` is neither 0 nor 1, then there's a bug in pytorch in that pytorch is assuming that nanobind will honor a non-default stream.  That is not a feature that nanobind supports.

Also, I don't understand that line of code for ROCm.  It looks like `stream_ptr` is set to `stream.cuda_stream`, which doesn't seem right for ROCm.

wjakob: Please see https://github.com/wjakob/nanobind/pull/1383 for a proposal from me. Would this be acceptable to everyone?

Ahajha: I agree the comment is out of date. But `stream=1` could be supported by other implementations. I wonder though if a hard error is necessary, would it be sufficient to just ignore it entirely?

Ahajha: @wjakob It sounds like we came to the same conclusion, so +1 from me!

Ahajha: Thanks for getting this resolved!

PR #1375, opened 2026-06-15
Optimize the critical path of ndarray

wjakob:
This commit bundles several performance improvements targeting the critical path of the ndarray implementation.

The following table shows performance improvements for a pair of microbenchmarks that repeatedly take (**Consume**) or return (**Produce**) a NumPy array. Most improvements are not specific to NumPy and also benefit other array frameworks. The time values are shown individually (nanoseconds per conversion) and as cumulative percentage relative to the baseline.

  | Commit | consume | produce |
  |---|---|---|
  | Cache export callable | ~same | 430.4→218.6 (−49%) |
  | Export via `asarray` | ~same | 430.4→187.7 (−56%) |
  | Maybe skip GIL acquire on free | 109.1→106.4 (−2%) | 430.4→178.8 (−58%) |
  | Don't zero moved-from descriptor | 109.1→100.2 (−8%) | ~same |
  | Populate caster in place | 109.1→98.8 (−9%) | ~same |
  | Unversioned `__dlpack__` | 109.1→89.7 (−18%) | ~same |
  | Skip capsule rename | 109.1→85.7 (−21%) | ~same |

cc @hpkfft

### DIFF
--- a/src/nb_ndarray.cpp
+++ b/src/nb_ndarray.cpp
@@ -1043,31 +1043,34 @@ ndarray_handle *ndarray_create(void *data, size_t ndim, const size_t *shape_in,
     return result.release();
 }
 
-/// Resolve (and cache) the callable that exports an ndarray a given framework.
-static PyObject *ndarray_export_fn(nb_internals *internals_, int framework) {
-    PyObject *fn = internals_->ndarray_export[framework].load_acquire();
+/// Module + attribute of export callables, indexed by `ndarray_export_slot`.
+static constexpr struct { const char *pkg, *attr; }
+    ndarray_export_spec[nd_export_count] = {
+        { "numpy",                          "asarray"     },
+        { "numpy",                          "copy"        },
+        { "torch.utils.dlpack",             "from_dlpack" },
+        { "tensorflow.experimental.dlpack", "from_dlpack" },
+        { "jax.dlpack",                     "from_dlpack" },
+        { "cupy",                           "from_dlpack" },
+    };
+
+/// Resolve (and cache) the callable for an ``ndarray_export`` cache slot.
+static PyObject *ndarray_export_fn(nb_internals *internals_,
+                                   ndarray_export_slot slot) {
+    PyObject *fn = internals_->ndarray_export[slot].load_acquire();
     if (NB_LIKELY(fn))
         return fn;
 
     lock_internals guard(internals_);
-    fn = internals_->ndarray_export[framework].load_relaxed();
+    fn = internals_->ndarray_export[slot].load_relaxed();
     if (fn)
         return fn;
 
-    const char *pkg_name, *attr;
-    switch (framework) {
-        case numpy::value:      pkg_name = "numpy";                          attr = "array";       break;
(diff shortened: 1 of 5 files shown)


wjakob:
I will merge this for now as it passes the test suite. I am happy to still make changes if there are any objections :-).

wjakob:
@hpkfft Does this commit sequence have your blessing? (as heavy user of the `ndarray` interface and somebody who understands its internals extremely well :-)).

src/nb_ndarray.cpp
```diff
@@ -1176,12 +1237,14 @@ PyObject *ndarray_export(ndarray_handle *th, int framework,
     }
 
     if (copy) {
-        PyObject* copy_function_name = NB_INTERNED(copy);
+        PyObject* copy_fn_name = NB_INTERNED(copy);
```
hpkfft, line 1240:
```suggestion
        PyObject* copy_fn_name;
```

hpkfft:
This LGTM.  The results are impressive!
I just have the one minor cleanup above.  I can make a PR or you can just sneak it into your next commit.  :wink:

src/nb_ndarray.cpp
```diff
@@ -1044,6 +1081,37 @@ ndarray_handle *ndarray_create(void *data, size_t ndim, const size_t *shape_in,
...
+    PyObject *fn = internals_->ndarray_export[slot].load_acquire();
+    if (NB_LIKELY(fn))
+        return fn;
+
+    lock_internals guard(internals_);
+    fn = internals_->ndarray_export[slot].load_relaxed();
+    if (fn)
+        return fn;
+
+    object obj = steal(module_import(ndarray_export_spec[slot].pkg))
```
hpkfft, line 1107: If I understand correctly, we have incremented the refcount of the module, which is good.
However, I think we'll leak it after nanobind exits.  Should it be put on the lifeline?

----

Actually, `steal()` returns an object, which will be destructed soon, which will decrement the refcount.
So, I think we have the opposite problem, but anyway, I think we ought to `new_object()` the module to put it on the lifeline.

wjakob: new_object in the lifeline calls `Py_DECREF` to transfer ownership to the lifeline.

hpkfft: No, I meant the PyObject returned by `module_import()`, not the result of `.attr()`.  [Sorry, you have to look below this comment to see it.]
I not at all sure if anything needs to be done here.  In general, if module A imports module B, and will continue to use functions from module B, then is A required to hold a reference to B?
Or, is the `object obj` above (from module B) itself required to hold a reference to module B?  So, it's a bug in module B if module B is garbage collected while we have `fn` on our lifeline?

wjakob: I deliberately rely on sys.modules to keep the module alive and only cache the callable.

However, regarding "Is it a bug in B if B is collected while we hold `fn?`" Even in the pathological case where someone does `del sys.modules['numpy']` and drops every other reference, our function still keeps its the function's module dict  alive through the function object, so fn(...) continues to resolve names correctly. The module wrapper object could be torn down, but the function stays callable.

PR #1376, opened 2026-06-15
perf: leverage immortal types (part 2)

hpkfft:
Also, this fixes a doc typo.


wjakob:
Awesome, thank you. The failure is due to a mistake in a (now amended) commit on `master`.

PR #1382, opened 2026-06-17
Support ``nb::pooled()`` for garbage-collected types

wjakob:
Instance pooling previously rejected any GC-tracked type (those using ``nb::dynamic_attr()``, ``nb::is_weak_referenceable()``, or a custom ``tp_traverse``). Such types can now be pooled too. The overhead of this is below the noise floor, and it brings the benefit of pooling to many more situations.


wjakob:
@zcbenz @awni @hawkinsp @XXXXRT666 Following this change, you can add ``nb::pooled()`` to the array class of MLX. Upgrading nanobind and applying this change makes symbolic tracing ~15% faster on my MBP.
```diff
diff --git a/python/src/array.cpp b/python/src/array.cpp
index 3cef838b..7e1f3967 100644
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -299,7 +299,8 @@ void init_array(nb::module_& m) {
       "array",
       R"pbdoc(An N-dimensional array object.)pbdoc",
       nb::type_slots(array_slots),
-      nb::is_weak_referenceable())
+      nb::is_weak_referenceable(),
+      nb::pooled(/* capacity = */ 128))
       .def(
           "__init__",
           [](mx::array* aptr, nb::object v, std::optional<mx::Dtype> t) {
```

wjakob:
(on a silly microbenchmark that just does some simple arithmetic in a loop to build a graph without evaluating it).

PR #1383, opened 2026-06-17
ndarray: accept and ignore the ``__dlpack__`` stream argument

wjakob:
This commit adapts ``nb_ndarray`` so that its ``__dlpack__()`` method ignores the ``stream`` argument. The documentation explains the rationale and explains how to deal with situations requiring synchronization.


hpkfft:
I'm good with this.

Ahajha:
+1 from me as well

wjakob:
Great, thanks to you both. @hpkfft sorry for spamming you, but it would be great to have your feedback on PR #1375.

PR #1389, opened 2026-06-23
Allow use of nanobind/eigen/sparse.h without nanobind/eigen/dense.h

jwnimmer-tri:
This allows projects to use the sparse caster while writing their own dense caster.

### DIFF
--- a/include/nanobind/eigen/sparse.h
+++ b/include/nanobind/eigen/sparse.h
@@ -10,7 +10,6 @@
 #pragma once
 
 #include <nanobind/ndarray.h>
-#include <nanobind/eigen/dense.h>
 #include <Eigen/SparseCore>
 
 #include <memory>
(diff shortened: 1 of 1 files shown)


wjakob:
I am curious for the need to write a custom dense caster. Is there an issue with the one provided by nanobind?

jwnimmer-tri:
For context, I am porting my project from pybind11 to nanobind (https://github.com/RobotLocomotion/drake/issues/21572).

The problem is converting Python lists (and lists-of-lists) to `Eigen::Vector` and `Eigen::Matrix`, just like https://github.com/wjakob/nanobind/discussions/327.

We have many bindings of C++ functions that accept an `Eigen` argument.  Because we created the bindings years ago with pybind11, our users have grown accustomed to passing a list and have it accepted.

So for my project, the dense caster needs to call `numpy.asarray` as a conversion fallback, so that lists are accepted.

wjakob:
Very cool (re: the drake port).

PR #1392, opened 2026-07-07
Improve compatibility of `nanobind.stubgen` with `scikit-build-core`

### LINKED ISSUE #1391: [BUG]: `nanobind.stubgen` is not compatible with `scikit-build-core` 1.0.0
### Problem description

Since https://github.com/scikit-build/scikit-build-core/pull/1403, the `__loader__` of a module built by `scikit-core-build` is no longer an `importlib.machinery.ExtensionFileLoader`. This change breaks this check in `nanobind.stubgen`:

https://github.com/wjakob/nanobind/blob/1a9dc48b8401122f6e4191ce6a347ead0b78394e/src/stubgen.py#L1553-L1561

As a result, `nanonind.stubgen` may create stubs with invalid file names (in particular, the `abi3` shouldn't be there):

```
Module "mqt.core.dd" ..
  - importing ..
  - analyzing ..
  - writing stub "/home/runner/work/core/core/python/mqt/core/dd.abi3.pyi" ..
```

### Reproducible example code

I hope the description above suffices in this case.

denialhaag:
This PR fixes #1391 by weakening a check that affects the file name of the generated stubs. I'm not sure whether this is _the_ solution, but it works for our case.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -1568,8 +1568,7 @@ def main(args: Optional[List[str]] = None) -> None:
                 )
             file = Path(str(file_s))
 
-            ext_loader = importlib.machinery.ExtensionFileLoader
-            if isinstance(mod_imported.__loader__, ext_loader):
+            if file.suffix in importlib.machinery.EXTENSION_SUFFIXES:
                 # Splitting on "." (module nesting qualifier) handles the case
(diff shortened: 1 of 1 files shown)


wjakob:
Looks fine to me. I may revert this change if it causes issues elsewhere.

PR #1394, opened 2026-07-08
Initial AIX port

KV2773:
This patch resolves some macro redefinition errors and linker flag errors on AIX.

### DIFF
--- a/src/nb_internals.h
+++ b/src/nb_internals.h
@@ -24,6 +24,10 @@
 #include <functional>
 #include "hash.h"
 
+#if defined(_AIX) && defined(func_data)
+# undef func_data
+#endif
+
 #if TSL_RH_VERSION_MAJOR != 1 || TSL_RH_VERSION_MINOR < 3
 #  error nanobind depends on tsl::robin_map, in particular version >= 1.3.0, <2.0.0
 #endif
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -230,7 +230,7 @@ function (nanobind_build_library TARGET_NAME)
     nanobind_lto(${TARGET_NAME})
 
     nanobind_strip(${TARGET_NAME})
-  elseif(NOT WIN32 AND NOT APPLE)
+  elseif(NOT WIN32 AND NOT APPLE AND NOT AIX)
     target_compile_options(${TARGET_NAME} PUBLIC $<${NB_OPT_SIZE}:-ffunction-sections -fdata-sections>)
     target_link_options(${TARGET_NAME} PUBLIC $<${NB_OPT_SIZE}:-Wl,--gc-sections>)
   endif()


tests/test_ndarray.cpp
```diff
@@ -474,8 +474,10 @@ NB_MODULE(test_ndarray_ext, m) {
     m.def("noop_2d_f_contig",
           [](nb::ndarray<float, nb::ndim<2>, nb::f_contig>) { return; });
 
-    m.def("accept_rw", [](nb::ndarray<float, nb::shape<2>> a) { return a(0); });
-    m.def("accept_ro", [](nb::ndarray<const float, nb::shape<2>> a) { return a(0); });
+    m.def("accept_rw",
+          [](nb::ndarray<float, nb::shape<2>, nb::device::cpu> a) { return a(0); });
+    m.def("accept_ro",
+          [](nb::ndarray<const float, nb::shape<2>, nb::device::cpu> a) { return a(0); });
```
hpkfft: Why are these two changes needed?  These functions ought to accept ndarrays that are on the cpu without having to specify `nb::device::cpu`.

KV2773: I guess some old changes have been pushed which I didn't make ,
lemme revert them back to include just my changes.

Sry for the confusion..

src/nb_internals.h
```diff
@@ -24,6 +24,10 @@
 #include <functional>
 #include "hash.h"
 
+#if defined(_AIX) && defined(func_data)
+# undef func_data
+#endif
```
hpkfft, line 29: According to Google, an alternative would be adding `-D_LINUX_SOURCE_COMPAT` to the compiler command line.  I'm fine either way.  Actually, maybe it's best done here so that it works for CMake, Meson, Bazel, etc.

hpkfft (approved):

src/nb_internals.h
KV2773: Yeah, I think going ahead this way will work better for different build systems..

KV2773: @hpkfft  if everything is fine may you merge this PR into the main branch as there is a CVE for onnx which I have to resolve and which is using this nanobind , so I have raised a PR there and the  maintainers are waiting for this PR of mine to be merged ..
PR for [onnx](https://github.com/onnx/onnx/pull/8077)

hpkfft: @wjakob is the owner of nanobind; only he can merge.

KV2773: > @wjakob is the owner of nanobind; only he can merge.

ohkk got it !!

KV2773: Hii @wjakob  may you review it once and can merge the PR  if everything is fine..

KV2773: Hey @wjakob  , may you merge the PR.. if everything is fine..

KV2773: Hey @wjakob  when will we be releasing the new version of nanobind as the conda-forge's nanobind there still we need this patch to be applied as the onnx they are asking to make changes there also..  
[cbourjau](https://github.com/cbourjau) commented on my [PR ](https://github.com/onnx/onnx/pull/8077)
`The Pixi-CI Jobs use conda-forge's nanobind so the patch is never applied.`

PR #1395, opened 2026-07-09
Make TypeError message more helpful when subclasses forget to call __init__

virtuald:
Alternative implementation of a fix for #1210. Created via AI, but I tested it manually and it works.
```python
import sys
sys.path += ["./build/tests"]

import test_classes_ext as t

class MissingInit(t.Struct):
    def __init__(self):
        pass

s = MissingInit()
s.value()
```

```
# python3 t.py 
/home/virtuald/nanobind/t.py:11: RuntimeWarning: nanobind: attempted to access an uninitialized instance of type '__main__.MissingInit'!

  s.value()
  File "/home/virtuald/nanobind/t.py", line 11, in <module>
    s.value()
    ~~~~~~~^^
TypeError: value(): incompatible function arguments. The following argument types are supported:
    1. value(self) -> int

Invoked with types: __main__.MissingInit

Hint: instance is not initialized (of type __main__.MissingInit). This can happen when a Python subclass overrides `__init__` but does not call `super().__init__()`.
```

One thing that's different from #1224 and pybind11 is that it doesn't attempt to tell the user they messed up immediately since that adds a performance hit for each constructor and I couldn't find a way to avoid that. Instead, since there's already an (unhelpful) error message that throws a type error, this just adds the detection there so the performance hit is only in potential error cases.

... I don't actually like how it implemented this (nor do I like the hint message..), but posting it for @wjakob to see what you think of the approach. If this seems like an acceptable approach, I can iterate on it some more.

### DIFF
--- a/src/nb_func.cpp
+++ b/src/nb_func.cpp
@@ -520,24 +520,14 @@ PyObject *nb_func_new(const func_data_prelim_base *f) noexcept {
     }
 }
 
-static void nb_func_put_uninitialized_arg_hint(PyObject *const *args,
-                                                size_t nargs,
-                                                size_t start) noexcept {
-    for (size_t i = start; i < nargs; ++i) {
-        PyObject *arg = args[i];
-        if (!nb_type_check((PyObject *) Py_TYPE(arg)))
-            continue;
+static void nb_func_put_arg_type(PyObject *arg) noexcept {
+    str name = steal<str>(nb_inst_name(arg));
+    buf.put_dstr(name.c_str());
 
+    if (nb_type_check((PyObject *) Py_TYPE(arg))) {
         nb_inst *inst = (nb_inst *) arg;
-        if (inst->state.state != nb_inst_state::state_uninitialized)
-            continue;
-
-        str name = steal<str>(nb_inst_name(arg));
-        buf.put("\n\nHint: instance is not initialized (of type ");
-        buf.put_dstr(name.c_str());
-        buf.put("). This can happen when a Python subclass overrides ");
-        buf.put("`__init__` but does not call `super().__init__()`.");
-        return;
+        if (inst->state.state == nb_inst_state::state_uninitialized)
+            buf.put(" (__init__() not called)");
     }
 }
 
@@ -575,8 +565,7 @@ nb_func_error_overload(PyObject *self, PyObject *const *args_in,
 
     buf.put("\nInvoked with types: ");
     for (size_t i = 0; i < nargs_in; ++i) {
-        str name = steal<str>(nb_inst_name(args_in[i]));
-        buf.put_dstr(name.c_str());
+        nb_func_put_arg_type(args_in[i]);
         if (i + 1 < nargs_in)
             buf.put(", ");
     }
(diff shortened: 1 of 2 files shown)


src/nb_func.cpp
```diff
@@ -585,6 +606,12 @@ nb_func_error_overload(PyObject *self, PyObject *const *args_in,
         buf.put(" }");
     }
 
+    size_t nargs_total = nargs_in;
+    if (kwargs_in)
+        nargs_total += (size_t) NB_TUPLE_GET_SIZE(kwargs_in);
+    size_t hint_start = (f->flags & (uint32_t) func_flags::is_constructor) ? 1 : 0;
+    nb_func_put_uninitialized_arg_hint(args_in, nargs_total, hint_start);
```
virtuald: Probably should also detect for kwargs too.

src/nb_func.cpp
```diff
@@ -520,6 +520,27 @@ PyObject *nb_func_new(const func_data_prelim_base *f) noexcept {
...
+        PyObject *arg = args[i];
+        if (!nb_type_check((PyObject *) Py_TYPE(arg)))
+            continue;
+
+        nb_inst *inst = (nb_inst *) arg;
+        if (inst->state.state != nb_inst_state::state_uninitialized)
+            continue;
+
+        str name = steal<str>(nb_inst_name(arg));
+        buf.put("\n\nHint: instance is not initialized (of type ");
```
virtuald: Maybe should add this to an exception note instead? But that would be more complicated.

virtuald:
This commit changes the hint to be inline, I like this a little better (though I don't like the phrasing?):

```
# python t.py
/home/virtuald/nanobind/t.py:11: RuntimeWarning: nanobind: attempted to access an uninitialized instance of type '__main__.MissingInit'!

  s.value()
  File "/home/virtuald/nanobind/t.py", line 11, in <module>
    s.value()
    ~~~~~~~^^
TypeError: value(): incompatible function arguments. The following argument types are supported:
    1. value(self) -> int

Invoked with types: __main__.MissingInit (__init__() not called)
```

wjakob:
How about "uninitialized" instead of "__init__() not called"?

virtuald:
Well, the problem with just that phrasing is that a user unfamiliar with nanobind might not understand what action they need to take (or wasn't taken). I can imagine a user thinking "uninitialized? I set up my object, what could that possibly mean?".

How about:

```
Invoked with types: __main__.MissingInit (uninitialized: __init__() not called)
```

---

I asked GPT it's opinion, the conversation is at https://chatgpt.com/share/6a5e2bad-03f4-83ea-95d8-82b28f7ee397

TL;DR: it made a separate suggestion which I'll need to add, but it thinks a combination of the inline hint and an explicit hint afterwards... and I like this too:

```
Invoked with types: __main__.MissingInit (uninitialized)

Hint: If __main__.MissingInit defines __init__(), it must call
super().__init__() to initialize the base class.
```

wjakob:
I do prefer `uninitialized`. I don't think the role of nanobind should be to teach Python concepts.

PR #1268, opened 2026-01-19
Fix conversion of size-zero vectors to `Eigen::Map`/`Eigen::Ref` on Numpy 2.4

ManifoldFR:
This implements the fix discussed in #1259

Resolves #1259

PR #1398, opened 2026-07-15
Fix type_caster failure for ndim=2 mutable eigen row vectors

### LINKED ISSUE #1397: [BUG]: Failure to convert shape=(1, N) array into mutable Eigen::Ref<Eigen::MatrixX...>
### Problem description

The following new unit test demonstrates the problem:

```py
def test05b_mutable_matrix_ref_given_row_vector_at_runtime():
    A = np.array([[1, 2, 3]], dtype=np.uint32)
    assert A.flags['F_CONTIGUOUS']
    assert_array_equal(t.addRefMXuCC(A, A), A + A)
```

We have `A.shape = (1, 3)` -- a row vector.  It has an `F_CONTIGUOUS` storage order.  (It is also `C_CONTIGUOUS`.)

The `addRefMXuCC` function takes arguments of type `const Eigen::Ref<Eigen::Matrix<uint32_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::ColMajor>>&`.  NumPy data with `F_CONTIGUOUS` storage is compatible with that mutable Ref.

However, on current `master` we get `TypeError`:
```
>       assert_array_equal(t.addRefMXuCC(A, A), A + A)
E       TypeError: addRefMXuCC(): incompatible function arguments. The following argument types are supported:
E           1. addRefMXuCC(arg0: numpy.ndarray[dtype=uint32, shape=(*, *)], arg1: numpy.ndarray[dtype=uint32, shape=(*, *)], /) -> numpy.ndarray[dtype=uint32, shape=(*, *), order='F']
E       
E       Invoked with types: ndarray, ndarray
```

I imagine that when NumPy provides a ndim=2 matrix with both storage orders, the type_caster is choosing one of them (i.e., assuming it's *only* `C_CONTIGUOUS`) instead of exploring both choices.

### Reproducible example code

See https://github.com/wjakob/nanobind/pull/1398.

jwnimmer-tri:
When casting ndarray to Eigen::Map, the numpy runtime stride in a given direction need not be compatible with the Eigen compile-time stride when the runtime size in that direction is zero or one, because the stride will never be used to index the memory.

This is relevant when a `shape=(1, *)` ndarray is passed into a `shape=(*, *, order='F')` map.

Closes #1397.

### DIFF
--- a/include/nanobind/eigen/dense.h
+++ b/include/nanobind/eigen/dense.h
@@ -261,29 +261,32 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
         if (!caster.from_python(src, flags & ~cast_flags::accepts_none, cleanup))
             return false;
 
-        // Check for memory layout compatibility of non-contiguous 'Map' types
-        if constexpr (!is_contiguous_v<Map>)  {
-            // Dynamic inner strides support any input, check the fixed case
-            if constexpr (StrideType::InnerStrideAtCompileTime != Eigen::Dynamic) {
-                // A compile-time stride of 0 implies "contiguous" ..
-                int64_t is_expected = StrideType::InnerStrideAtCompileTime == 0
-                                      ? 1 // .. and equals 1 for the inner stride
-                                      : StrideType::InnerStrideAtCompileTime,
-                        is_actual = caster.value.stride(
-                            (ndim_v<T> != 1 && T::IsRowMajor) ? 1 : 0);
-
-                if (is_expected != is_actual)
+        // The `caster` already validated the shape, and for is_contiguous_v it
+        // also validated the memory layout, but if not then we need to validate
+        // against StrideType. Dynamic strides are compatible with any input, so
+        // only compile-time strides need checking (and can be ignored when the
+        // size in their dimension is <= 1).
+        if constexpr (!is_contiguous_v<Map>) {
+            constexpr size_t inner_dim = (ndim_v<T> == 2 && T::IsRowMajor) ? 1 : 0;
+            constexpr size_t outer_dim = 1 - inner_dim;
+            constexpr int64_t IS = StrideType::InnerStrideAtCompileTime;
+            if constexpr (IS != Eigen::Dynamic) {
+                int64_t actual_stride = caster.value.stride(inner_dim);
+                // A compile-time stride of 0 implies "contiguous".
+                constexpr int64_t expected_stride = IS == 0 ? 1 : IS;
+                if (actual_stride != expected_stride &&
+                        caster.value.shape(inner_dim) > 1)
                     return false;
             }
-
-            // Analogous check for the outer strides
-            if constexpr (ndim_v<T> == 2 && StrideType::OuterStrideAtCompileTime != Eigen::Dynamic) {
-                int64_t os_expected = StrideType::OuterStrideAtCompileTime == 0
-                                        ? caster.value.shape(T::IsRowMajor ? 1 : 0)
-                                        : StrideType::OuterStrideAtCompileTime,
-                        os_actual   = caster.value.stride(T::IsRowMajor ? 0 : 1);
-
-                if (os_expected != os_actual)
+            constexpr int64_t OS = StrideType::OuterStrideAtCompileTime;
+            if constexpr (ndim_v<T> == 2 && OS != Eigen::Dynamic) {
+                int64_t actual_stride = caster.value.stride(outer_dim);
+                // A compile-time stride of 0 implies "contiguous".
+                int64_t expected_stride = OS == 0
+                                            ? (int64_t) caster.value.shape(inner_dim)
+                                            : OS;
+                if (actual_stride != expected_stride &&
+                        caster.value.shape(outer_dim) > 1)
                     return false;
             }
         }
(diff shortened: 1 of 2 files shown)


wjakob:
Would you mind turning this into a more constructive PR that also fixes the issue?

jwnimmer-tri:
Sure!  I didn't jump right into that in case there was a context where this was "working as intended" or some subtlety.

I'll ping back here when I have a candidate fix.  (It won't be right away.)

jwnimmer-tri:
Okay, it took me a bit of study to convince myself how to spell the fix, but I think the latest push is good.

jwnimmer-tri:
FYI @ManifoldFR from #1268 -- this re-implements the map-stride logic in a way that replaces your patch.  I think my new way is still correct (and more clear), and I have tested vs numpy 2.5, but FYI in case you want to check for yourself.

wjakob:
@ManifoldFR Could I ask you to review this PR?

jwnimmer-tri:
At the risk of overstepping, let me also gently ping @jorisv in case @ManifoldFR is busy or no longer working on Eigen bindings.

ManifoldFR:
Hi @wjakob @jwnimmer-tri 

I'm rather busy during weeks now, but I'll have time to look at this PR this weekend.

include/nanobind/eigen/dense.h
```diff
@@ -261,29 +261,32 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
...
+        if constexpr (!is_contiguous_v<Map>) {
+            constexpr size_t inner_dim = (ndim_v<T> == 2 && T::IsRowMajor) ? 1 : 0;
+            constexpr size_t outer_dim = 1 - inner_dim;
+            constexpr int64_t IS = StrideType::InnerStrideAtCompileTime;
+            if constexpr (IS != Eigen::Dynamic) {
+                int64_t actual_stride = caster.value.stride(inner_dim);
+                // A compile-time stride of 0 implies "contiguous".
+                constexpr int64_t expected_stride = IS == 0 ? 1 : IS;
+                if (actual_stride != expected_stride &&
+                        caster.value.shape(inner_dim) > 1)
```
ManifoldFR, line 278: Okay I see, this is the most important change along with L289

ManifoldFR: Also good job on `constexpr`-ing as much as possible here

include/nanobind/eigen/dense.h
```diff
@@ -314,44 +317,14 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
...
-        if constexpr (ndim_v<T> == 1) {
-            if (caster.value.shape(0) == 0)
-                inner = IS == Eigen::Dynamic ? 1 : IS;
-        }
-
-        if constexpr (OS == 0)
-            outer = 0;
-
+        constexpr int64_t IS = StrideType::InnerStrideAtCompileTime,
+                          OS = StrideType::OuterStrideAtCompileTime;
```
ManifoldFR: The only change here is going from `int` (32-bit) to `int64_t`, is this justified? IIRC Eigen uses `int` for all of these size-related template params and class-scope constants

include/nanobind/eigen/dense.h
```diff
@@ -314,44 +317,14 @@ struct type_caster<Eigen::Map<T, Options, StrideType>,
...
         if constexpr (std::is_same_v<StrideType, Eigen::InnerStride<IS>>)
-            return StrideType(inner);
+            return StrideType(inner_stride());
         else if constexpr (std::is_same_v<StrideType, Eigen::OuterStride<OS>>)
-            return StrideType(outer);
+            return StrideType(outer_stride());
         else
-            return StrideType(outer, inner);
```
ManifoldFR: This is mostly a refactor out to `inner_stride()/outer_stride()` right? It's much cleaner than before, thanks!

ManifoldFR (approved):
Looks good, thanks a lot!

I was thinking, do you think a test case where the user passes a callback to C++ which modifies an Eigen array as follows:

```python
def py_callback(x: np.ndarray):
   n = x.shape[1]
   v = np.random.randn(n)
   if x.shape[0] > 0:
     x[0, :] * = v
```

and where the call site passes a Ref to a F contig matrix, would be relevant?

I remember this being a failure case in eigenpy, because `x` would get coerced to an actual 1D numpy array

include/nanobind/eigen/dense.h
jwnimmer-tri: > Okay I see, this is the most important change along with L289.

Yes, exactly.

include/nanobind/eigen/dense.h
jwnimmer-tri: Changed back to `int`.

include/nanobind/eigen/dense.h
jwnimmer-tri: Yes, mostly a refactor for clarity and cleanliness.

The one functional changes is that when a compile-time stride is set, it always wins.  Previously there was special case logic for 0 and 1, but now any non-dynamic FooStrideAtCompileTime is immediately used without looking at the ndarray.  Since we already validated it during the cast, the simple logic is to just use it.

ManifoldFR: Oh yes that's better. If the cast is accepted no need to check, which is probably faster.

wjakob:
All good to merge? The changes look good to me.

ManifoldFR:
Good to merge in my view 👍

wjakob:
Done, thanks @jwnimmer-tri and @ManifoldFR for the review.

jwnimmer-tri:
Thanks both!

FYI I have a follow-up on my TODO list to see about adding more test cases, but with other commitments and vacations coming up it will take some time before I can submit that.

PR #1399, opened 2026-07-18
Support CPython 3.15

take-cheeze:
CIs tested on https://github.com/take-cheeze/nanobind/pull/1

Also fixes 3.14t finding issues from CMake 4.4.0 with `Python_FIND_ABI` flag

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
         submodules: true
 
     - name: Setup Python 3.14t (free-threaded)
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.14t'
 
@@ -139,6 +139,7 @@ jobs:
     - name: Check ABI tag
       run: >
         cd build/tests;
+        ls -alh ;
         python -c 'import test_functions_ext as t; print(f"ABI tag is \"{ t.abi_tag() }\"")'
 
     - name: Run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'windows-2022', 'macos-15']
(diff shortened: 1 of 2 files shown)


hpkfft:
Would you be kind enough to update https://nanobind.readthedocs.io/en/latest/api_cmake.html#high-level-interface since you understand Python's build choices and naming rules?

Please see also: https://discuss.python.org/t/can-we-consider-an-uncoditional-change-of-abi3t-so-to-abi3t-soabi-platform-so-for-3-15-last-minute/107919

Personally, I prefer, for example, `hpk.abi3t-x86_64-linux-gnu.so` over `hpk.abi3t.so`.
[Maybe, that's done by scikit_build_core?  I don't know how it all works.]

take-cheeze:
@hpkfft 
Thank you for reviewing PR.
This PR currently have lack of support of abi3t. I've forgot to add CI of 3.15 with stable abi test.
I'm currently working on it locally and solving symbol hiding by `_Py_OPAQUE_PYOBJECT` seems to be quite hard.

It seems there is no document update since `FREE_THREADED` and `STABLE_ABI` is already documented.
Maybe I need to add how the combination of two options behave.

`.abi3t-x86_64-linux-gnu.so` seems to be better but `Python_SOSABI` which is based on following seems to use `.abi3t.so` :
```
>>> import importlib
>>> importlib.machinery.EXTENSION_SUFFIXES 
['.cpython-315t-x86_64-linux-gnu.so', '.abi3t.so', '.so']
```

Also I'd like to run `abi3audit` but abi3t support seems to be on going( https://github.com/pypa/abi3audit/issues/215 ) so maybe it would be a future task

wjakob:
nanobind does not support abi3t for Python 3.15. It _may_ support it for Python 3.16. Therefore I am surprised to see abi3t-related changes in the PR.

wjakob:
(The fact that you got the Python test suite to compile for abi3t means that something is deeply broken here.). api3t makes PyVarObject opaque that nanobind needs to be non-opaque.

take-cheeze:
Sorry for being hasty.
I've read https://github.com/wjakob/nanobind/compare/master...abi3t and realized the status.
Opaque object seems to be hard part of abi3t support

Issue #1346, opened 2026-05-25
stubgen backward-compatible or platform-dependent?


nilsnolde:
for quick context: I use `nanobind_add_stub` for a project and checked the .pyi files into git for docs building purposes.

it seems that another dev was getting local diffs on the .pyi files without touching any python bindings code, e.g. whitespace diff. I _think_ he's on macos and/or has a different version of nanobind. I honestly didn't go investigate, but maybe someone has experience with this? we wouldn't want git churn because of the stubs.

has someone else experienced this? ([GH comment](https://github.com/Project-OSRM/osrm-backend/pull/7571#issuecomment-4527689848) for reference)

wjakob:
Nanobind may generate different output depending on the Python version. This is expected. Stubs are a build output, it is not recommended that you commit them to your repository.

nilsnolde:
argh, ok, thanks! thought I was being smart by committing them so I don't have to rebuild a multi-100k codebase only to produce a few docs pages 😅

wjakob:
I guess that _can_ be an argument to commit them :-). But then you have to be careful about using a consistent compilation environment.

nilsnolde:
yeah not gonna happen realistically for our projects 😅 

I do still wonder if there's a way to make them more reproducible across python versions? in the hope the docs use case might make it a valuable (yet reasonable) path. note: I have no idea how the stubsgen currently works! a short "no" works for me too :D

PR #1400, opened 2026-07-19
Make stubgen independent of minor python version

nilsnolde:
I figured I'd see what this looks like: https://github.com/wjakob/nanobind/issues/1346#issuecomment-4728046059.

main reasoning:
- I'd like to optionally check in the output stubs to my repo, to e.g. build docs via e.g. `mkdocs` without cmake, so:
- any (supported) python interpreter version should generate byte-identical stubs

was surprisingly few code changes to surprisingly sane python logic:) 

> [!NOTE]
> - this is intentionally only the src changes; I also have tests, most importantly to fail "new" `typing` definitions in >= 3.15 so they can be manually added
> - it does work nicely according to tests; still, of course feel free to reject, not much work lost
> - this is mostly claude, though I did understand, tweak and approve the changes

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -98,6 +98,39 @@ SKIP_LIST = [
     "_unhashable_values_map_", "_value_repr_",
 ]
 
+# Python version in which various ``typing.*`` members were first introduced
+TYPING_INTRODUCED: Dict[str, Tuple[int, int]] = {
+    "Annotated": (3, 9),
+    "Concatenate": (3, 10),
+    "ParamSpec": (3, 10),
+    "ParamSpecArgs": (3, 10),
+    "ParamSpecKwargs": (3, 10),
+    "TypeAlias": (3, 10),
+    "TypeGuard": (3, 10),
+    "is_typeddict": (3, 10),
+    "LiteralString": (3, 11),
+    "Never": (3, 11),
+    "NotRequired": (3, 11),
+    "Required": (3, 11),
+    "Self": (3, 11),
+    "TypeVarTuple": (3, 11),
+    "Unpack": (3, 11),
+    "assert_never": (3, 11),
+    "assert_type": (3, 11),
+    "clear_overloads": (3, 11),
+    "dataclass_transform": (3, 11),
+    "get_overloads": (3, 11),
+    "reveal_type": (3, 11),
+    "TypeAliasType": (3, 12),
+    "override": (3, 12),
+    "NoDefault": (3, 13),
+    "ReadOnly": (3, 13),
+    "TypeIs": (3, 13),
+    "get_protocol_members": (3, 13),
+    "is_protocol": (3, 13),
+    "evaluate_forward_ref": (3, 14),
+}
+
 # Interpreter-internal types.
 TYPES_TYPES = {
     getattr(types, name): name for name in [
@@ -212,7 +245,8 @@ class StubGen:
         max_expr_length: int = 50,
         patterns: List[ReplacePattern] = [],
         quiet: bool = True,
-        output_file: Optional[Path] = None
+        output_file: Optional[Path] = None,
+        min_python: Optional[Tuple[int, int]] = None
     ) -> None:
         # Module to check for name conflicts when adding helper imports
         self.module = module
@@ -243,6 +277,9 @@ class StubGen:
 
         # Target filename, only needed for recursive stub generation
         self.output_file = output_file
+
+        # Oldest target Python version (e.g. ``(3, 9)``)
+        self.min_python = min_python
 
         # ---------- Internal fields ----------
 
@@ -583,10 +620,7 @@ class StubGen:
 
             if same_module:
                 # This is an alias of a type in the same module or same top-level module
-                if sys.version_info >= (3, 10, 0):
-                    alias_tp = self.import_object("typing", "TypeAlias")
-                else:
-                    alias_tp = self.import_object("typing_extensions", "TypeAlias")
+                alias_tp = self.import_object(self.typing_module((3, 10)), "TypeAlias")
                 self.write_ln(f"{name}: {alias_tp} = {tp.__qualname__}\n")
             elif self.include_external_imports or (same_toplevel_module and self.include_internal_imports):
                 # Import from a different module
@@ -608,6 +642,14 @@ class StubGen:
                     if tp_bases is None:
                         tp_bases = tp.__bases__
                     tp_bases = [self.type_str(base) for base in tp_bases]
+
+                # 'enum.StrEnum' only exists on Python 3.11+. When targeting
+                # older versions, emit the equivalent bases that nanobind
+                # itself uses there for string-valued enumerations
+                if self.min_python and self.min_python < (3, 11) \
+                        and "enum.StrEnum" in tp_bases:
+                    pos = tp_bases.index("enum.StrEnum")
+                    tp_bases[pos:pos + 1] = ["str", "enum.Enum"]
 
                 if tp_bases != ["object"]:
                     self._replace_tail(2, "(")
@@ -690,10 +732,7 @@ class StubGen:
             if self.is_type_var(tp):
                 types = ""
             elif typing.get_origin(value):
-                if sys.version_info >= (3, 10, 0):
-                    types = ": " + self.import_object("typing", "TypeAlias")
-                else:
-                    types = ": " + self.import_object("typing_extensions", "TypeAlias")
+                types = ": " + self.import_object(self.typing_module((3, 10)), "TypeAlias")
             else:
                 types = f": {self.type_str(tp)}"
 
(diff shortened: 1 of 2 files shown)


src/stubgen.py
```diff
@@ -98,6 +98,39 @@ class and repeatedly call ``.put()`` to register modules or contents within the
     "_unhashable_values_map_", "_value_repr_",
 ]
 
+# Python version in which various ``typing.*`` members were first introduced
+TYPING_INTRODUCED: Dict[str, Tuple[int, int]] = {
```
nilsnolde, line 102: this would also be the basis for the tests: together with `typing.__all__` from the min supported python version (3.9 currently) and this list, we get test failures on new `typing` additions in the future

nilsnolde: so far I'm just having a bit of trouble importing `stub_gen.py` to the test file. in case you have a recommendation or precedent here, I'd appreciate it. I usually then symlink the source file into the test dir, but I just found out that apparently breaks windows.

src/stubgen.py
```diff
@@ -609,6 +643,14 @@ def put_type(self, tp: NbType, name: Optional[str]):
                         tp_bases = tp.__bases__
                     tp_bases = [self.type_str(base) for base in tp_bases]
 
+                # 'enum.StrEnum' only exists on Python 3.11+. When targeting
```
nilsnolde, line 649: this was a bit of a dance.. if we generate on e.g. 3.12 for min_version < 3.11 and there's `enum.StrEnum` being used, we replace it with a plain `enum.Enum`. it's a bit of a wart maybe? 

if we don't do this, then those cases would still emit different stubs.

src/stubgen.py
```diff
@@ -1023,6 +1071,15 @@ def put(self, value: object, name: Optional[str] = None, parent: Optional[object
             self.stack.pop()
             self.prefix = old_prefix
 
+    def typing_module(self, introduced: Tuple[int, int]) -> str:
```
nilsnolde, line 1077: convenience function to decide importing `typing` or `typing_extensions`

src/stubgen.py
```diff
@@ -1036,6 +1093,28 @@ def import_object(
         if module == "builtins" and name and (not as_name or name == as_name):
             return name
 
+        # When targeting an explicit minimum Python version, import members
```
nilsnolde, line 1099: mostly deduping `typing` & `typing_extensions`

src/stubgen.py
```diff
@@ -1469,6 +1565,11 @@ def parse_options(args: List[str]) -> argparse.Namespace:
         parser.error(
             "The -o option is not compatible with recursive stub generation (-r)."
         )
+    if opt.min_python is not None:
```
wjakob: Why not parse both digits? There will be a v4 Python version at some point.

src/stubgen.py
```diff
@@ -1595,7 +1696,8 @@ def main(args: Optional[List[str]] = None) -> None:
             include_private=opt.include_private,
             max_expr_length=0 if opt.exclude_values else 50,
             patterns=patterns,
-            output_file=file
+            output_file=file,
+            min_python=opt.min_python
```
wjakob: The name is inconsistent wrt. the command line flag. (--min-python-version)?

src/stubgen.py
```diff
@@ -1023,6 +1071,15 @@ def put(self, value: object, name: Optional[str] = None, parent: Optional[object
             self.stack.pop()
             self.prefix = old_prefix
 
+    def typing_module(self, introduced: Tuple[int, int]) -> str:
+        """
+        Return ``typing`` or ``typing_extensions`` depending on the minimum 
+        Python version (defaults to currently executing Python version).
+
+        """
+        version = self.min_python or sys.version_info[:2]
```
wjakob: How about initializing `min_python` properly so that `sys.*` doesn't need to be used as a fallback here?

src/stubgen.py
```diff
@@ -748,6 +787,15 @@ def simplify_types(self, s: str) -> str:
 
         s = self.abc_re.sub(r'collections.abc.\1', s)
 
+        # Normalize 'CapsuleType', it's baked into extension signatures
+        if self.min_python:
```
wjakob: remove `None` guard

src/stubgen.py
```diff
@@ -609,6 +643,14 @@ def put_type(self, tp: NbType, name: Optional[str]):
                         tp_bases = tp.__bases__
                     tp_bases = [self.type_str(base) for base in tp_bases]
 
+                # 'enum.StrEnum' only exists on Python 3.11+. When targeting
+                # older versions, emit the equivalent bases that nanobind
+                # itself uses there for string-valued enumerations
+                if self.min_python and self.min_python < (3, 11) \
```
wjakob: remove None guard.

src/stubgen.py
wjakob: This solution is OK for me.

src/stubgen.py
wjakob: How about renaming this to TYPING_MINVER?

wjakob: Can you let me know what the issue is on Windows?

src/stubgen.py
```diff
@@ -1036,6 +1093,28 @@ def import_object(
         if module == "builtins" and name and (not as_name or name == as_name):
             return name
 
+        # When targeting an explicit minimum Python version, import members
+        # that ``typing`` does not provide from ``typing_extensions``
+        if self.min_python and module == "typing" and name:
+            introduced = TYPING_INTRODUCED.get(name)
+            module = module if not introduced else self.typing_module(introduced)
+
+        # ``typing_extensions`` re-exports the ``typing`` API. Avoid conflicts by
```
wjakob: This seems kind of complicated. Would it be better to do this as some kind of post cleanup at the end before the stubs are generated?

wjakob (commented):
Some comments

src/stubgen.py
nilsnolde: apparently:

- on windows, creating symlinks may require developer mode or elevated privileges
- git’s symlink handling differs depending on configuration

src/stubgen.py
nilsnolde: it's aliased to `min_python` actually, the cli flag is `--min-python-version`. I didn't like that split either and agree.

src/stubgen.py
wjakob: Can't we set the python path or import from an installed nanobind module?

nilsnolde:
thanks @wjakob for the quick review! I wanted to gauge interest first before going over it in detail. will do tmrw or later and see what can be done about https://github.com/wjakob/nanobind/pull/1400#discussion_r3611000794

src/stubgen.py
nilsnolde: yeah PYTHONPATH was also my first reaction. was hoping for some precedent. I don't even know yet, why the source is not just simply importable. i.e. I have more homework to do on this PR :sweat_smile:

nilsnolde: commonly I have this problem when trying to test-import from an executable python script, which isn't part of the actual library API.

nilsnolde: argh haha, this _is_ an executable python script. nvm

src/stubgen.py
henryiii: Not before at least 3.33, however. :P

nilsnolde: now doing it, it feels strange tbh. now people can put in "5.2" and it won't bark. I can make it bark, but then we'd be back to hard-coding 3. I'd actually leave it as-is. who knows what'll be in year 2044 post-3.33 when 4.0 might be getting real :sweat_smile: wdyt @wjakob ?

src/stubgen.py
```diff
@@ -98,6 +98,42 @@ class and repeatedly call ``.put()`` to register modules or contents within the
...
+    "NoDefault": (3, 13),
+    "ReadOnly": (3, 13),
+    "TypeIs": (3, 13),
+    "get_protocol_members": (3, 13),
+    "is_protocol": (3, 13),
+    "evaluate_forward_ref": (3, 14),
+}
+
+# Oldest Python version that stubgen supports targeting.
+MIN_PYTHON_VERSION = (3, 9)
```
nilsnolde, line 135: I introduced this, so it's easier to keep track of all the min python usage for version checking

src/stubgen.py
```diff
@@ -1308,6 +1382,17 @@ def get(self) -> str:
...
+        # ``typing_extensions`` re-exports the ``typing`` API. Keep  copy,
+        # import names are identical
+        typing_imports = self.imports.get("typing")
+        te_imports = self.imports.get("typing_extensions")
+        if typing_imports and te_imports:
+            for key in list(typing_imports):
+                name, as_name = key
+                # keep "as" import names in ``typing``
+                if name and not as_name and key in te_imports:
+                    del typing_imports[key]
```
nilsnolde, line 1394: you're right, this is much nicer to do in the end

src/stubgen.py
```diff
@@ -1450,6 +1538,15 @@ def parse_options(args: List[str]) -> argparse.Namespace:
         help="force the use of ... for values",
     )
 
+    parser.add_argument(
+        "--min-python-version",
+        metavar="VERSION",
+        default=format_version(MIN_PYTHON_VERSION),
```
nilsnolde, line 1544: initializing the min version properly

src/stubgen.py
```diff
@@ -1469,6 +1566,22 @@ def parse_options(args: List[str]) -> argparse.Namespace:
         parser.error(
             "The -o option is not compatible with recursive stub generation (-r)."
         )
+
+    # determine if it's a supported python version
```
nilsnolde, line 1570: and error out on > 3 or < 3.9

src/stubgen.py
nilsnolde: nvm, I landed on https://github.com/wjakob/nanobind/pull/1400/changes#r3647418284 for now

nilsnolde:
I'll add the rest soon (test etc). those were uncommitted on another computer.

Issue #1155, opened 2025-09-09
[OTHER]: Generated stubs for nb::ndarray are too strict about dtype (e.g., NDArray[float64]) when nanobind converts inputs implicitly


nurpax:
### Please don't use.

# Generated stubs for `nb::ndarray` are too strict about dtype (e.g., `NDArray[float64]`) when nanobind converts inputs implicitly

**Summary**

Recent nanobind generates `.pyi` where parameters typed as `nb::ndarray<const double, ...>` become `Annotated[NDArray[float64], ...]`. Static checkers (e.g., VS Code Pylance) then reject valid calls that pass non-`float64` arrays -- even though nanobind converts these to `double` at runtime. Older nanobind emitted `ArrayLike`, which didn’t trigger false positives. 

Request: a way to generate looser stubs to reflect accepted inputs when dtype conversion is allowed.

---

### Environment

* nanobind @ 10cc5109bdb55f7e35583e0fd6a5e3ed06c95683
* Affected checker: VS Code Pylance (strict about `NDArray[...]` vs `ArrayLike`)

### Background / behavior change

Example C++ binding:

```cpp
using ndarray_1d = nb::ndarray<const double, nb::ndim<1>, nb::device::cpu, nb::c_contig>;

m.def("plot_line", [](const char* label_id, ndarray_1d& xs, ndarray_1d& ys, ImPlotLineFlags_ flags) {
    ImPlot::PlotLine(label_id, (const double*)xs.data(), (const double*)ys.data(), xs.shape(0), flags);
}, "label_id"_a, "xs"_a, "ys"_a, "flags"_a.sig("LineFlags.NONE") = ImPlotLineFlags_None);
```

Old stubs:

```py
xs: Annotated[ArrayLike, dict(dtype='float64', shape=(None), order='C', device='cpu', writable=False)]
```

New stubs that issue a type error on valid inputs:

```py
xs: Annotated[NDArray[float64], dict(shape=(None,), order='C', device='cpu', writable=False)]
```

### Problem (false positive type errors)

A call like this is valid at runtime (nanobind converts to `double`) but flagged by Pylance:

```py
import numpy as np

data_x = np.arange(0, 10, 0.25)      # NDArray[floating[Any]]
data_y = np.sin(data_x)
implot.plot_line("Sine Wave", data_x, data_y)
```

Error (abridged):

```
Argument of type "_Array1D[floating[Any]]" cannot be assigned to parameter "xs" of type "NDArray[float64]" ...
[...]

PR #1409, opened 2026-08-04
Correctly import nested array modules, process annotations like for N…

nicholasjng:
…umPy

Until now, nanobind took care to create Python-legal annotations for `nb::ndarray`s resulting in numpy arrays on the Python side. Other frameworks did not get this treatment, which resulted in extras being added in square brackets behind the array type (e.g. `mlx.core.array[dtype=float32, shape=(2, 4)]`), which is not legal Python syntax, and causes tools like ruff and ty to choke on the stub files.

The fixes: Extend the array regex in `StubGen` to pick up other frameworks as well, and add branches in _format_ndarray to output legal annotations for each of these frameworks as part of the processing done in `simplify_types`.

Regenerates the reference stubs for a few cases, including those for JAX and TensorFlow, which are not run in CI, and contained array imports that are not in use anymore today, to the best of my knowledge.

---------------

Hey! I ran into this when using stubgen in https://github.com/nicholasjng/metal-runtime, where I'm returning `nb::ndarray`s without any annotations, resulting in empty brackets like `mlx.core.array[]`, which is an error. Of course, `nb::sig` comes to the rescue, but I figured it was worth fixing upstream.

Reference stubs were updated because they were not valid Python, and could thus not be type-checked or reformatted e.g. in CMake post-build steps using ruff or ty.

The reason why I changed the `is_valid_module` check again is because it returned `True` for any module `foo` that's not even installed - the error branches fire only on dotted accesses, i.e. attempted imports of submodules.

Someone with JAX/TF knowledge could fact-check my changes in array imports and annotations, but I'm reasonably confident I picked the correct names here.


nicholasjng:
Another question here is whether the import pattern should be `import array_namespace` up top, and type as `array_namespace.ArrayT` (like `jax.Array` in this PR), or the pre-existing NumPy case (`from numpy.typing import NDArray` up top, `NDArray` in annotations everywhere).

hpkfft:
At the risk of appearing greedy, I'd like to ask whether the stubs are generated correctly for PyTorch.
Also, I've been hoping someone knowledgeable about typing would consider #1155.  :wink:

nicholasjng:
heh, forgot about Pytorch entirely, thanks for that. Updated the lookup table just now.

hpkfft:
Nicholas, thank you

wjakob:
Hi Nicolas,

I am happy with this, but note that there were significant changes to the stub generator that will require adaptation in this PR. Could I ask you to rebase and fix it onto master?

nicholasjng:
Thanks! Just rebased.

tests/test_ndarray_ext.pyi.ref
```diff
@@ -127,25 +128,25 @@ def ret_numpy_const_ref_f() -> Annotated[NDArray[numpy.float32], dict(shape=(2,
...
 def ret_noframework_noowner() -> Annotated[NDArray[numpy.float64], dict(shape=(2, 4))]: ...
 
-def ret_array_api() -> ArrayLike[dtype=float64, shape=(2, 4)]: ...
+def ret_array_api() -> Annotated[Any, dict(dtype=float64, shape=(2, 4))]: ...
 
-def ret_array_api_byte_offset() -> ArrayLike[dtype=float64, shape=(2, 4)]: ...
+def ret_array_api_byte_offset() -> Annotated[Any, dict(dtype=float64, shape=(2, 4))]: ...
 
-def ret_array_api_metal() -> ArrayLike[dtype=float32, shape=(2, 4), order='C', device='metal']: ...
+def ret_array_api_metal() -> Annotated[Any, dict(dtype=float32, shape=(2, 4), order='C', device='metal')]: ...
```
wjakob: If the goal is to generate valid stubs, then one remaining issue is that dtype stays as-is, which leads to output like `dict(dtype=float32, ...)`. (Except for NumPy, where it gets turned into `np.float32`). How about quoting it in the other cases?

src/stubgen.py
```diff
@@ -108,6 +108,15 @@ class and repeatedly call ``.put()`` to register modules or contents within the
     ]
 }
 
+# ndarray framework types as 2-tuples of (module, array type).
+_FRAMEWORK_STUB_TYPES = {
```
wjakob: Kind of redundant wrt. `NDARRAY_PAT`. I would suggest that we maintain only one of these lists.

src/stubgen.py
wjakob, line 756: The docstring now is no longer accurate.

wjakob (commented):
Thanks! A few more comments/requests.

nicholasjng:
Thanks for the review, made another pass. A few of the frameworks have dtype exports (like jax.numpy), but I would leave that post-processing to a follow-up.

wjakob:
One more quick request: given how strongly `tensorflow.python.framework.ops.EagerTensor` was rewritten, I think the better thing is to also modify `ndarray.h` so that we don't generate this internal name.

nicholasjng:
Done. Also removed a stray dot from the TF module prefix in `import_data`, since it was not there on the other entries, either.

wjakob:
Great, thank you!

PR #1413, opened 2026-08-12
update docs: call destructor before placement new in "Pickling"

ricrogz:
This amends the documentation section about pickling and unpickling to recommend calling the destructor before the placement `new`. The reason is explained in the comment I added: the placement `new` will overwrite the old object with the newly created one, which won´t call the objects destructor, causing any potential non-POD members to leak.

By explicitly calling the destructor before the placement  `new`, the non-POD members are destructed before they are overwritten with the new object.

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -1008,6 +1008,12 @@ pickling. The example below, e.g., does this using a tuple.
 The ``__setstate__`` method should construct the object in-place analogous to
 custom ``__init__``-style constructors.
 
+Note that by calling ``new (&pet)`` we allocate a new ``Pet`` object at the
+same memory position the old one was. This means that, since the destructor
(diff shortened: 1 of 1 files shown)


wjakob:
I think that this is incorrect: `__setstate__` is handled as a constructor. You can't double-call the C++ constructor. You can verify this experimentally with the test suite bindings.

```
>>> s = Struct(5); s.__setstate__(7)
  RuntimeWarning: nanobind: attempted to initialize an already-initialized instance of type 'test_classes_ext.Struct'!
  TypeError: __setstate__(): incompatible function arguments.
```

ricrogz:
My bad. I didn't quite understand what you said about double-calling a constructor, but I now see that I was mistaken about the  `pet` argument that the `__setstate__` lamba receives.

PR #1414, opened 2026-08-16
Fix conversion warning with 32-bit Eigen::Index

sertonix:
```
/tmp/src/nanobind-3.0.1/include/nanobind/eigen/dense.h:343:43: error: conversion from 'int64_t' {aka 'long long int'} to 'Eigen::Index' {aka 'int'} may change value [-Werror=conversion]
  343 |                 return caster.value.stride(0);
      |                        ~~~~~~~~~~~~~~~~~~~^~~
```


jwnimmer-tri:
This function is being rewritten in #1398.  Please check if that new code solves your problem.

wjakob:
ping @sertonix

sertonix:
There is still a conversion issue in 3.0.1

Edit: updated the patch

wjakob:
Thanks!

PR #1420, opened 2026-08-23
Gate the chrono float-to-seconds rule on 'cast_flags::convert'

yosh-matsuda:
# Gate the chrono float-to-seconds rule on `cast_flags::convert`

`duration_caster::from_python` ignores its `flags` and reads any float as a number of seconds, including in the first pass of overload resolution. Nothing that matches a float without a conversion can win against it. The caster has discarded `flags` since it was ported from pybind11 in 1.3.0 (#175).

```cpp
m.def("f", [](std::chrono::milliseconds) { return 1; });
m.def("f", [](double) { return 2; });
```

```python
>>> f(2.0)
1     # expected 2: 'double' matches a float, while the duration only converts it
```

Both casters accept the float in the first pass, so the overload that is declared first wins.

`nb::arg().noconvert()` does not stop the rule either:

```cpp
m.def("g", [](std::chrono::milliseconds d) { return d.count(); }, nb::arg().noconvert());
```

```python
>>> g(2.0)
2000  # expected TypeError
```

The same happens inside a `std::variant`, where a float subclass such as `numpy.float64` never reaches the alternative that wants the number:

```cpp
m.def("h", [](std::variant<double, std::chrono::milliseconds> v) { return v.index(); }); // <nanobind/stl/variant.h>
```

```python
>>> h(numpy.float64(2.0))
1     # expected 0, i.e. the value became a duration of two seconds
```

## Fix

Test `cast_flags::convert` before the rule. The float is still accepted in the second pass, so a call changes behavior only in two cases: the argument is `noconvert`, or another candidate matches the float in the first pass, as in an overload set or a `std::variant`.

## The input name

The change also drops `float` from the input name of the caster:

```diff
-    NB_TYPE_CASTER(type, io_name("datetime.timedelta | float",
-                                 "datetime.timedelta"))
+    NB_TYPE_CASTER(type, const_name("datetime.timedelta"))
```

The input name lists what a caster accepts without a conversion, as `type_caster<double>` names `float` and not `int`. The signature of an existing function therefore changes from `arg: datetime.timedelta | float` to `arg: datetime.timedelta`.

## Tests

`tests/test_chrono.{cpp,py}` gain two tests, one for the overload set and one for `noconvert`. 

`docs/api_extra.rst` now marks both float entries of the chrono table as implicit conversions.

### DIFF
--- a/docs/api_extra.rst
+++ b/docs/api_extra.rst
@@ -1325,14 +1325,17 @@ converting to Python.
 - ``float`` → ``std::chrono::duration``
     A floating-point value can be converted into a duration. The input is
     treated as a number of seconds, and fractional seconds are supported
-    to the extent representable.
+    to the extent representable. This is an :ref:`implicit conversion
+    <noconvert>`: it does not take place in the first pass of overload
+    resolution, nor when an argument is marked ``noconvert``.
 
 - ``float`` → ``std::chrono::[other_clock]::time_point``
     A floating-point value can be converted into a timepoint on a
     clock other than the system clock. The input is treated as a
     number of seconds, and fractional seconds are supported to the
     extent representable. The resulting timepoint will be that many
-    seconds after the target clock's epoch time.
+    seconds after the target clock's epoch time. This is an :ref:`implicit
+    conversion <noconvert>`, as above.
 
 
 Evaluating Python expressions from strings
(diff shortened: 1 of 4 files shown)


wjakob:
That is very bad, thanks for the fix!

AnsgarKasselmann:
Is it intentional, that the stubs now always show `datetime.timedelta` instead of `datetime.timedelta | float` even if `noconvert` is not set?

PR #1421, opened 2026-08-23
Take the address of a dllimport slot at run time

yosh-matsuda:
# Take the address of a dllimport slot at run time

`NB_SLOT_ALIAS` initializes a `constexpr` pointer with the address of a CPython function (`nb_backend.h:615`):

```cpp
#define NB_SLOT_ALIAS(ret, name, args, target) \
    inline constexpr ret (*name) args = &target;
```

Windows declares a CPython function with `__declspec(dllimport)`, so its address is the content of an import slot that the loader fills, and not a constant expression. Clang rejects the initializer:

```cpp
// The declaration that Python.h produces on Windows
extern "C" __declspec(dllimport) int PyObject_Vectorcall(int);
inline constexpr int (*vectorcall)(int) = &PyObject_Vectorcall;
```

```
> clang-cl /std:c++17 /c repro.cpp
repro.cpp(2,24): error: constexpr variable 'vectorcall' must be initialized by a constant expression
```

Every Windows target of Clang fails, both `--target=x86_64-pc-windows-msvc` and `--target=x86_64-w64-windows-gnu`, so no extension builds with clang-cl. MSVC accepts the initializer as an extension, which is why the `windows-2022` job of the CI stays green.

The macro arrived with the vector call slots in 9219167 and reached its first release in 3.0.0.

## Fix

Make the pointer `const` instead of `constexpr`:

```diff
-    inline constexpr ret (*name) args = &target;
+    inline ret (*const name) args = &target;
```

## Tests

The error appears at compile time on one platform, so the test is a CI job. `.github/workflows/ci.yml` gains a `clang-cl` job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -299,3 +299,46 @@ jobs:
         source /opt/intel/oneapi/setvars.sh
         cd build
         python -m pytest
+
+  clang-cl:
+    runs-on: windows-2022
+    name: "Python ${{ matrix.python }} / clang-cl"
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ['3.12']
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+
+    - name: Setup Python ${{ matrix.python }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python }}
+        cache: 'pip'
+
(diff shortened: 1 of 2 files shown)


wjakob:
Thank you. I will trim back some of the comments of these changes, I think they are overkill. The commit history and PR description is enough to look into the detailed rationale.

PR #1422, opened 2026-08-23
Keep the qualified name of a nested class in a stub

yosh-matsuda:
Inside a class body, `stubgen` strips the prefix of the enclosing class and writes the short name of a nested class (`stubgen.py:844`). The two type checkers do not agree on what that name means:

```python
# m.pyi
class Sibling:
    top: int

class MyClass:
    class Sibling:
        nested: str
    def g(self, arg: Sibling, /) -> None: ...
```

```python
MyClass().g(MyClass.Sibling())
```

```
> mypy       Success: no issues found
> pyright    error: Argument of type "Sibling" cannot be assigned to parameter "arg" of type "Sibling"
             "m.MyClass.Sibling" is not assignable to "m.Sibling"
```

mypy reads the short name in the scope of the class, and pyright reads it as the module level name. Nothing in the message points at the stub, and the type is silently wrong whenever the module holds a class of that name.

A qualified name has no such ambiguity, and both checkers accept it:

```python
    def g(self, arg: MyClass.Sibling, /) -> None: ...
```

## Relation to #1333

The rule came from #1333, which states that a type checker cannot resolve `EnumWrapper.Value` inside the body of `EnumWrapper` and reports `Unknown`. A stub is never executed, so neither checker has that problem today:

```python
class EnumWrapper:
    class Value(enum.Enum):
        Alpha = 0
    Alpha: EnumWrapper.Value = EnumWrapper.Value.Alpha
    def get_value(self) -> EnumWrapper.Value: ...
```

```
> pyright    Type of "EnumWrapper.Alpha" is "Value"
> mypy       Revealed type is "q.EnumWrapper.Value"
```

The generated `test_enum_ext.pyi` gives the same result: pyright reports the identical set of 8 diagnostics with the short and the qualified name, and mypy the same 4. Both sets are unrelated to the names of the nested classes.

### DIFF
--- a/src/stubgen.py
+++ b/src/stubgen.py
@@ -841,14 +841,6 @@ class StubGen:
             mod_prefix = self.module.__name__ + "."
             if full_name.startswith(mod_prefix) and \
                     (result := full_name[len(mod_prefix) :]).split(".")[0] in self.module.__dict__:
-                # Inside a class body, also strip the immediate enclosing class
-                # prefix (e.g. "ErrorEnum.Value" -> "Value"). Only that scope
-                # is stripped: class scopes don't nest, so an outer class's
-                # members aren't visible by short name in a nested class body.
-                scope = self.prefix[len(mod_prefix) :]
-                enclosing = scope.rpartition(".")[0]
-                if enclosing and result.startswith(enclosing + "."):
-                    result = result[len(enclosing) + 1 :]
                 return result
             elif mod_name in ("typing", "typing_extensions", "collections.abc"):
                 # Import frequently-occurring typing classes and ABCs directly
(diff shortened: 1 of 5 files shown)


wjakob:
Ugh, looks like you are right. At the time, I tried the offending example code from #1333 in PyRight, but I think I did so after saving it into a `.py` file and not a `.pyi` file.

Turns out both PyRight and MyPy are perfectly happy with the nested type names when they occur in a `.pyi` stub. So the premise of this change was incorrect.

markjm:
apologies for the churn - fwiw this does look to be a problem for my case still, but is perhaps an issue/disparate behaviour specifically with `ty` - I should have checked across all the common checkers. My bad!

https://play.ty.dev/257e3e12-0593-41b4-8187-d7b7a85a374e

PR #1423, opened 2026-08-25
Add missing <cstdio> includes for C-stdio symbols for pyoxide(emscripten)

take-cheeze:
## Summary

nanobind's own `.cpp` files and, as of a recent upstream header reorganization, three public headers (`nb_backend.h`, `nb_types.h`, `ndarray.h`) use `fprintf`/`stderr`/etc. without including `<cstdio>`. They build only because `<Python.h>` transitively pulls in `<stdio.h>` — but CPython excludes that transitive include once `Py_LIMITED_API` targets Python ≥ 3.11, which is exactly what this project's stable-ABI CMake support (`STABLE_ABI`) sets. A stable-ABI build — needed for a portable Pyodide wasm wheel — therefore fails with `use of undeclared identifier 'stderr'`/`'fprintf'`.

Found while cross-compiling [onnxsim](https://github.com/onnxsim/onnxsim) for `wasm32-emscripten` via Pyodide.

**Fix:** add `#include <cstdio>` to the affected files:
- `src/common.cpp`, `src/nb_internals.cpp`, `src/nb_ndarray.cpp`, `src/nb_type.cpp`
- `include/nanobind/nb_backend.h`, `include/nanobind/nb_types.h`, `include/nanobind/ndarray.h` (public headers — this also affects any user extension targeting the stable ABI, not just nanobind's own build)

**Regression test:** a new `wasm32-emscripten (Pyodide)` CI job (`tests/wasm/`, `.github/workflows/pyodide.yml`) cross-compiles a minimal stable-ABI extension for wasm32-emscripten, `pip install`s it into a `pyodide venv`, and runs it with `pytest` — an ordinary Python packaging workflow, just targeting Pyodide. See `tests/wasm/README.md`.

**Aside, documented but not fixed:** while checking whether the wasm test could reuse the existing `nanobind-backend` package via split mode, I found split mode genuinely crashes on Emscripten (an uncaught, `catch (...)`-proof exception across the `dlopen(RTLD_LOCAL)` module boundary) even though it isn't ELF. Left the existing `nb_defs.h` guard as-is and expanded its comment with this finding so a future attempt to narrow it doesn't have to rediscover it.

## Verification

Reproduced the exact compiler error (`-DPy_LIMITED_API=0x030b0000` against real CPython 3.11 headers, and separately under a real Emscripten/Pyodide toolchain) on the pre-fix tree, confirmed `#include <cstdio>` resolves it on the fixed tree, and confirmed no regression in the normal (non-limited-API) desktop build. The new CI job passes end-to-end on GitHub Actions, with `pytest` reporting `platform emscripten`.

### DIFF
--- a/.github/workflows/pyodide.yml
+++ b/.github/workflows/pyodide.yml
@@ -1,0 +1,73 @@
+name: pyodide
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
+      - stable
+      - v*
+
+concurrency:
+  group: pyodide-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  # See tests/wasm/README.md.
+  wasm32-emscripten:
+    name: "wasm32-emscripten (Pyodide) smoke test"
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+
+    - name: Setup host Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+        cache: 'pip'
+
+    - name: Install pyodide-build
+      run: |
+        # wheel>=0.46 dropped 'wheel.cli', which auditwheel_emscripten needs.
+        python -m pip install pyodide-build "wheel<0.46"
+
+    - name: Install the Pyodide cross-build environment
+      run: |
+        pyodide xbuildenv install
+        pyodide config list
+
+    - name: Install the matching (Pyodide-patched) Emscripten SDK
+      run: |
+        pyodide xbuildenv install-emscripten
+
+    - name: Install the latest CMake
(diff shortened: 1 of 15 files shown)


hpkfft:
> public headers — this also affects any user extension targeting the stable ABI, not just nanobind's own build

Maybe a function for a critical nanobind error can be provided in nanobind's own build, and the public headers can call that function instead of calling `fprintf(stderr, ...); abort();`.

take-cheeze:
Great idea!
I couldn't come up with that since I was only targeting pyoxide build.
nanobind isn't header only library so that's more better

wjakob:
Sorry, this PR is too AI-generated (commit messge, AI slop comments). Direct output from Opus 5 is below the quality threshold of this project. Covering PyIodide is neat, but I'd want it to run as much of the full test site as possible. Also, this PR is mixing a few completely unrelated things.

Issue #1424, opened 2026-08-25
[BUG]: Nanobind 3.0.0 doesn't declare `requires-python = ">=3.10"`


godlygeek:
### Problem description

Nanobind 3.0.0 has dropped support for Python 3.9, but hasn't declared that its lower bound is now Python 3.10 in its packaging metadata. As a result, nanobind 3.0.0 successfully installs for Python 3.9, but fails at runtime with an error saying:
```py
      File "/tmp/build-env-e4l3c86y/lib/python3.9/site-packages/nanobind/__init__.py", line 5, in <module>
        raise ImportError("nanobind does not support Python < 3.10.")
```

If the metadata correctly reflected the minimum required Python version, projects building Python 3.9 code would automatically fall back to nanobind 2.x instead of failing to build.

Please yank nanobind 3.0.0 and cut a 3.0.1 release that correctly declares the required version for the 3.x series.

### Reproducible example code

```text

```

wjakob:
Fixed via https://github.com/wjakob/nanobind/pull/1425

godlygeek:
Thanks! Note that this isn't really fixed [unless you yank](https://docs.pypi.org/project-management/yanking/) 3.0.0, though. It's not enough that the next release will declare that it requires Python 3.10 or higher, we _also_ need the most recent non-yanked release that didn't declare that it needs 3.10 to actually work for 3.9 - otherwise a `python3.9 -m pip install --no-binary pystack pystack` will see that it needs nanobind, see that nanobind 3.0.1 is incompatible with Python 3.9, see that the next newest version was nanobind 3.0.0 which doesn't declare that it's incompatible with Python 3.9, and `pip install` will fall back to the incompatible version rather than falling back to a compatible version.

wjakob:
I will not yank 3.0.0 for that, this is too drastic of an action to such a minor issue. nanobind 3.0.0 is an API breaking change, and Python 3.9 is not even a supported Python release anymore.

godlygeek:
Why do you see it as drastic? Yanking 3.0.0 once you have published the next version (3.0.1 or 3.1.0 or whatever) will not have any negative effects:

1. Users who have pinned to `==3.0.0` will continue using it despite it having been yanked
2. Users who have not pinned it and who are running Python 3.10 or newer will pick up the new release
3. Users who have not pinned to it and who are running Python 3.9 or older will fall back to 2.x where their build can succeed, instead of falling back to 3.0.0 and failing.

[...]

PR #1425, opened 2026-08-25
Declare that nanobind now needs Python 3.10+

godlygeek:
This packaging metadata is used by package installers like `pip` to avoid installing an incompatible version of the package.

Relates-to #1424


wjakob:
Thanks!

PR #1427, opened 2026-08-31
cmake: only require Python::Interpreter for non-cross builds

### LINKED ISSUE #1412: [OTHER]: Hard requirement for `Python::Interpreter` (since 2.13.0 / #1351) breaks cross-compilation setups that never had a runnable interpreter to begin with
## Context

We build C++/Python extensions with nanobind inside conda-forge-style cross-compilation jobs (via `rattler-build`/`conda-build`), e.g. building on a `linux-64` machine for a `linux-aarch64` target. In this setup:

- `$PREFIX` (the "host" prefix, matching the *target* platform) contains a Python interpreter and headers/libs built **for the target architecture** (aarch64).
- `$BUILD_PREFIX` (matching the *build* platform, x86_64) is the only environment whose binaries can actually execute on the machine doing the compiling.

Before nanobind 2.13.0, our CMake helper deliberately called:

```cmake
find_package(Python $ENV{PY_VER} EXACT COMPONENTS Development.Module REQUIRED)
```
without Interpreter, and this was enough — nanobind only checked for TARGET Python::Module.

## Why we don't request Interpreter
We can't just add Interpreter to the COMPONENTS list, because the only Python interpreter find_package(Python) can locate that matches our target version lives under `$PREFIX` and it's an aarch64 binary. It "finds" it fine (the file exists, Python_EXECUTABLE gets set, the Python::Interpreter target gets created), but the moment anything actually tries to execute it (e.g. execute_process(COMMAND "${Python_EXECUTABLE}" ...)), it fails, because that interpreter is not for the platform running the build. Requesting Interpreter doesn't fail the CMake configure step — it produces a target that is silently unusable at build time.

Our cross-compiling branch works around this entirely by not needing an executable interpreter at all: Development.Module resolves headers/libs from the target sysroot (via CMAKE_FIND_ROOT_PATH), and the Python version/ABI info we need is fully known from $ENV{PY_VER} and the sysroot — no interpreter execution required for this step.

Here is the relevant part of our helper macro (from our internal phoenixcmake package), showing the distinction we make between the cross-compiling path (no Interpreter) and the native path (Interpreter requested and actually usable):
```cmake
macro(phoenix_configure_nanobind)
	# Cross-compilation using conda environment
	if(CMAKE_CROSSCOMPILING AND DEFINED ENV{PREFIX})
		message(STATUS "Cross-compiling, detecting Python version from filesystem...")

		set(nanobind_ROOT "$ENV{PREFIX}/lib/python$ENV{PY_VER}/site-packages/nanobind/cmake" CACHE STRING "Path to nanobind CMake configuration")
		set(PYTHON_SITE_PACKAGES "$ENV{PREFIX}/lib/python$ENV{PY_VER}/site-packages" CACHE STRING "Path to Python site-packages")

		# Deliberately NOT requesting Interpreter here: the only Python matching
		# $ENV{PY_VER} that find_package(Python) can locate is the one under
		# $ENV{PREFIX}, which is built for the *target* architecture and cannot
		# be executed on the machine performing the cross-build.
		find_package(Python $ENV{PY_VER} EXACT COMPONENTS Development.Module REQUIRED)

	# Native build using conda-build (PREFIX is set by rattler-build/conda-build)
	else()
		# Here Interpreter is safe to request: build platform == target platform,
		# so the interpreter found is actually runnable.
		find_package(Python $ENV{PY_VER} EXACT COMPONENTS Interpreter Development.Module REQUIRED)
		execute_process(
			COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
			OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE nanobind_ROOT_temp
		)
		# ...
	endif()

	find_package(nanobind CONFIG REQUIRED)
endmacro()
```

## The problem
Since #1351, cmake/nanobind-config.cmake starts with:

```cmake
if (NOT TARGET Python::Interpreter OR NOT TARGET Python::Module)
  message(FATAL_ERROR "You must invoke 'find_package(Python COMPONENTS Interpreter Development REQUIRED)' prior to including nanobind.")
endif()
```

This broke our cross-compiling branch above, since we never request Interpreter there.

We found a workaround: create an empty stand-in target purely so the check passes, without providing anything actually runnable behind it:

```cmake
if(NOT TARGET Python::Interpreter)
    add_library(Python::Interpreter INTERFACE IMPORTED)
endif()
```

This works for us because our Python_SOABI/Python_SOSABI are already fully resolved via Development.Module (from the sysroot headers/config, not from running the interpreter), so nanobind's configure step never actually tries to execute ${Python_EXECUTABLE} in our case. 

Still, we wanted to report this situation, since it highlights a case where requiring a runnable Python::Interpreter isn't actually necessary — and isn't always possible, as in our cross-compilation setup.

toprinse:
Fixes #1412.

# Problem

`cmake/nanobind-config.cmake` has unconditionally required a `Python::Interpreter` target since #1351. That change was meant to catch an incomplete `find_package(Python)` call in native builds (#1350), but it also blocks legitimate cross-compilation setups that never have a runnable interpreter for the target architecture — forcing users to fabricate a fake `Python::Interpreter INTERFACE` target just to satisfy the check (see the workaround described in #1412).

# Why this is safe

Looking at CMake's own `FindPython/Support.cmake`: when only the `Development.Module` component is requested (no Interpreter), CMake computes `Python_SOABI` via `python-config --extension-suffix` — a text script that runs fine on the build host even for a foreign-architecture Python — rather than by executing the target interpreter. So the ABI/version information nanobind needs at configure time (Python_SOABI, Python_VERSION) is already obtainable without a runnable interpreter when cross-compiling. The blanket check doesn't add real safety in that case; it only helps catch mistakes in ordinary (non-cross) builds, which is what #1350 was actually about.

# Changes
- Split the top-level check in `cmake/nanobind-config.cmake`: `Python::Module` is still always required. `Python::Interpreter` is only required when `NOT CMAKE_CROSSCOMPILING`, preserving the original #1350 protection for native builds unchanged.
- Added a targeted error in the `EXT_SUFFIX` fallback query (only reached when Python_SOABI/SKBUILD_SOABI aren't already known) when no runnable interpreter (`Python::Interpreter` target and a valid `Python_EXECUTABLE`) is available, explaining how to resolve it.
- Added the same targeted error in `nanobind_add_stub()`, which genuinely needs to execute Python to generate stubs and has no way around that. Also checks `Python_EXECUTABLE`, not just the target, so a leftover/fake `Python::Interpreter` target (e.g. from the workaround people currently use) still gets a clear message instead of a cryptic internal CMake error `(file(TO_CMAKE_PATH ...) argument-count failure)`.
- Documented the behavior in `docs/building.rst`.

# Testing
- Real end-to-end native build: configured, compiled, and ran a nanobind extension against the system Python (3.12), including nanobind_add_stub — confirms zero regression on the ordinary path.
- Native build missing `Python::Interpreter`: still fails with the original error message — confirms the #1350 protection is untouched.
- Simulated cross-compiling configure (`Python::Module` + Python_SOABI/Python_SOSABI set, no `Python::Interpreter`): configures successfully, correct NB_SUFFIX/NB_SUFFIX_S computed.
- Simulated cross-compiling + `nanobind_add_stub()` with no interpreter, and with a fake `Python::Interpreter` target (the existing workaround): both now report a clear, actionable error instead of failing deep inside CMake internals.

### DIFF
--- a/cmake/nanobind-config.cmake
+++ b/cmake/nanobind-config.cmake
@@ -1,6 +1,15 @@
 include_guard(GLOBAL)
 
-if (NOT TARGET Python::Interpreter OR NOT TARGET Python::Module)
+if (NOT TARGET Python::Module)
+  message(FATAL_ERROR "You must invoke 'find_package(Python COMPONENTS Interpreter Development REQUIRED)' prior to including nanobind.")
+endif()
+
+# A runnable interpreter is not required to configure nanobind: the ABI/SOABI
+# information used below comes from 'python-config' (a text script that runs
+# fine even for a foreign-architecture Python) rather than from executing the
+# target interpreter, so this is only enforced for non-cross builds, where its
+# purpose is catching an incomplete find_package(Python) call (see #1350).
+if (NOT TARGET Python::Interpreter AND NOT CMAKE_CROSSCOMPILING)
   message(FATAL_ERROR "You must invoke 'find_package(Python COMPONENTS Interpreter Development REQUIRED)' prior to including nanobind.")
 endif()
 
@@ -63,6 +72,15 @@ endif()
 
 # If either suffix is missing, call Python to compute it
 if(NOT DEFINED NB_SUFFIX OR NOT DEFINED NB_SUFFIX_S)
+  if(NOT TARGET Python::Interpreter OR NOT Python_EXECUTABLE)
+    message(FATAL_ERROR "nanobind: could not determine the extension module "
+      "suffix because no runnable Python interpreter is available (this can "
+      "happen when cross-compiling). Either arrange for 'Python_SOABI' (or "
+      "'SKBUILD_SOABI', if using scikit-build-core) to be set before "
+      "including nanobind, or provide a working 'Python::Interpreter' target "
+      "(with a valid 'Python_EXECUTABLE').")
+  endif()
+
   # Query Python directly to get the right suffix.
   execute_process(
     COMMAND "${Python_EXECUTABLE}" "-c"
@@ -942,6 +960,16 @@ function (nanobind_add_stub name)
     endif()
     list(APPEND NB_STUBGEN_OUTPUTS "${OUTPUT}")
   endforeach()
+
+  if (NOT TARGET Python::Interpreter OR NOT Python_EXECUTABLE)
+    message(FATAL_ERROR "nanobind_add_stub(): generating stubs requires "
+      "running the Python interpreter, but no runnable 'Python::Interpreter' "
+      "target (with a valid 'Python_EXECUTABLE') is available (this can "
+      "happen when cross-compiling). Re-run find_package(Python COMPONENTS "
+      "Interpreter ...) with an interpreter that can execute on the build "
+      "machine, e.g. via 'CMAKE_CROSSCOMPILING_EMULATOR', before calling "
+      "nanobind_add_stub().")
+  endif()
 
   file(TO_CMAKE_PATH ${Python_EXECUTABLE} NB_Python_EXECUTABLE)
 
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -37,6 +37,20 @@ conditional check.
 
     find_package(Python 3.10 COMPONENTS Interpreter ${DEV_MODULE} REQUIRED)
 
+.. note::
+
+    Cross-compilation is not officially supported, but nanobind does not
+    require a *runnable* ``Interpreter`` component when
+    ``CMAKE_CROSSCOMPILING`` is set: querying the extension module suffix
+    only relies on information that CMake's ``FindPython`` module can obtain
+    without executing the target Python (e.g. via a ``python-config``
+    script), so it is enough to request the ``Development.Module`` component
+    for the target Python. If that information cannot be determined this way,
+    set ``Python_SOABI`` (or ``SKBUILD_SOABI`` when using scikit-build-core)
+    manually before including nanobind. Generating stubs with
+    :cmake:command:`nanobind_add_stub` still requires an interpreter that can
+    run on the build machine.
+
 Add the following lines below. They configure CMake to perform an optimized
 *release* build by default unless another build type is specified. Without this
 addition, binding code may run slowly and produce large binaries.


hpkfft:
Can you provide a link to the `python-config` text script?  Also, to its documentation?

toprinse:
`python-config` isn't something nanobind uses directly — it's CMake's own `FindPython` module that falls back to it when no `Interpreter` component is available:
https://cmake.org/cmake/help/latest/module/FindPython.html#result-variables

Its **Result Variables** section documents this explicitly:

> `Python_SOABI` — Extension suffix for modules. Information computed from `sysconfig.get_config_var('EXT_SUFFIX')` or `sysconfig.get_config_var('SOABI')` or `python3-config --extension-suffix`.

nanobind's CMake config only ever reads the resulting `Python_SOABI` variable; it doesn't care which of these three methods CMake used to obtain it.

`python-config` itself is a small script shipped as part of a CPython installation (not part of the Python language spec — other implementations don't necessarily ship an equivalent), generated at CPython's own build time from:
https://github.com/python/cpython/blob/main/Misc/python-config.in

Its `valid_opts` list shows the flags it supports, including the one that matters here:
```python
valid_opts = ['prefix', 'exec-prefix', 'includes', 'libs', 'cflags',
              'ldflags', 'extension-suffix', 'help', 'abiflags', 'configdir',
              'embed']
```
and `--extension-suffix` resolves to `print(getvar('EXT_SUFFIX'))`, read via `sysconfig` — the same value referenced above.

There isn't a dedicated, comprehensive official documentation page for `python-config` as a standalone CLI tool. The closest official mention is in the *Embedding Python in Another Application* guide:
https://docs.python.org/3/extending/embedding.html#compiling-and-linking-under-unix-like-systems

— but it only documents `--cflags` and `--ldflags --embed` (for embedding), not `--extension-suffix`. So the CPython source above, together with CMake's `FindPython` docs, are the most authoritative references for this specific behavior.

hpkfft:
> `python-config` itself is a small script shipped as part of a CPython installation 

Oh, I see.  Thank you.  (I had looked for it in the CMake source code and obviously couldn't find it there.  Sorry.  I didn't expect you to spend time to write such a long reply, but I appreciate it.)

PR #1428, opened 2026-09-01
Fix abi3t detection on Windows

burgholzer:
## Summary

scikit-build-core currently cannot encode `abi3t` in `SKBUILD_SOABI` on Windows because Stable ABI extensions use the untagged `.pyd` suffix. It sets `Py_TARGET_ABI3T=1` instead.
See https://github.com/scikit-build/scikit-build-core/blob/5687cc3a0a6a085f87b3f378d3ed1dfdae2d544e/src/scikit_build_core/builder/builder.py#L447-L455

nanobind currently ignores that setting, leaves `NB_ABI` empty, and disables the requested `FREE_THREADED` split path. 
This PR changes the configuration so that `Py_TARGET_ABI3T` is honored when deriving `NB_ABI`.

Assisted-by: gpt-5.6-sol with Codex

(This change was prompted by us having to work around this in one of our projects; see https://github.com/munich-quantum-toolkit/core/blob/970b79a55c9fd355df5fa4a9eb8c0c029fa0a733/cmake/AddMQTPythonBinding.cmake#L13-L17)

### DIFF
--- a/.github/workflows/split.yml
+++ b/.github/workflows/split.yml
@@ -173,8 +173,12 @@ jobs:
   # abi3t extensions (PEP 803) served by a free-threaded backend module.
   # Python 3.15 is only available as a prerelease at the moment
   abi3t:
-    name: "split mode / Python 3.15t (abi3t) / ubuntu-latest"
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


wjakob:
Thanks!

PR #1434, opened 2026-09-13
Fix the reference type in dict_caster's item loop

o2161405:
Containers like std::flat_map return a prvalue pair when iterated that can't bind to the item loop (which currently binds lvalue references). Changing to `auto&&` should allow both to work.


wjakob:
This looks good. Can you also make this change in `nb_set.h`?

o2161405:
Done. Also now the only file that doesn't iterate with `auto&&` in that folder is `nb_array.h`, I can change that too if you want, but I don't really see the benefit (other than consistency).

wjakob:
I don't think it would make any difference there given that the type is fixed. Thanks for the PR!

PR #1435, opened 2026-09-14
Overhaul Bazel documentation

nicholasjng:
Removes the `nanobind_bazel`-specific documentation in favor of a more general document:
The two presented options for using nanobind in Bazel projects are 1) sourcing prebuilt modules from the BCR, and
2) using local, customized builds with a module extension.

Option 2) was what `nanobind_bazel` was using internally anyway, so the new docs do not reinvent Bazel packaging, but rather give users a more explicit guide on how to customize nanobind libraries.

Removes the `nanobind_bazel` API docs, since users are now expected to use the (slightly) lower-level `@nanobind//...` targets instead of the macros.

------------------------

This is what I promised in the split-mode discussion a while back. Of course, additional reviews by Bazel users (@jwnimmer-tri et al.) are most welcome.

I removed all references to `nanobind_bazel`, since the direct use of `@nanobind` is probably preferable in all cases today.

AI disclosure: I used GPT 5.6 for the first draft of `bazel.rst`, and then edited it substantially in post.


wjakob:
Looks good to me! @jwnimmer-tri ?

docs/bazel.rst
```diff
@@ -3,195 +3,193 @@
...
-    )
+    load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-In local development scenarios, you can clone nanobind-bazel to your machine,
-and then declare it as a ``local_path_override()`` dependency:
+    def _nanobind_impl(_):
+        http_archive(
+            name = "nanobind",
+            urls = ["https://github.com/wjakob/nanobind/archive/<revision>.tar.gz"],
+            strip_prefix = "nanobind-<revision>",
```
jwnimmer-tri, line 43: BTW If you're willing to assume current bazel (8.8, 9.2, 10.0) then `strip_components = 1` instead of `strip_prefix = "nanobind-<revision>"` avoids writing the same revision in two places.

docs/bazel.rst
```diff
@@ -3,195 +3,193 @@
...
-    load(
-        "@nanobind_bazel//:build_defs.bzl",
-        "nanobind_extension",
-        "nanobind_stubgen",
-    )
-
-    nanobind_extension(
-        name = "my_ext",
-        srcs = ["my_ext.cpp"],
+    config_setting(
```
jwnimmer-tri: BTW `config_setting` is an older and dis-preferred configuration mechanism.  Nowadays https://github.com/bazelbuild/bazel-skylib/blob/main/docs/common_settings_doc.md would be standard.

docs/bazel.rst
```diff
@@ -3,195 +3,193 @@
...
--------------------------------------
+Building for the stable ABI (ABI3)
+----------------------------------
 
-You can also use Bazel to generate stubs for an extension directly at build
-time with the ``nanobind_stubgen`` macro. Here is an example of a nanobind
-extension with a stub file generation target declared directly alongside it:
+The current BCR overlay provides a regular CPython build. To build an ABI3
+extension, use the module-extension setup above and customize the copied
+``nanobind.BUILD.bazel`` overlay. Add the following configuration settings:
```
jwnimmer-tri: Besides a custom build file, another way to add `defines` is via a Bazel transition.  That's the way my project does it (https://github.com/RobotLocomotion/drake/blob/master/tools/workspace/nanobind/defs.bzl) using `with_cfg.bzl`.  Possibly it's too expert-level to list it here, though.

In any case, I plan to eventually add direct support in BCR for abi3 and split mode, so I don't mind whatever content we put here in the meantime.

jwnimmer-tri (approved):
LGTM thanks!

docs/bazel.rst
nicholasjng: I thought about transitions here as well. What I wouldn't want is to take on additional dependencies (I think `@with_cfg.bzl` is one of them?), and to blow up the example code, so I lean towards leaving this as is.

jwnimmer-tri: Fine by me.

docs/bazel.rst
nicholasjng: I can make it a `bool_flag`. Do you think it's worth documenting the free-threaded build here too? In my mind, that would probably benefit from an example of a free-threaded toolchain registration in MODULE.bazel.

docs/bazel.rst
nicholasjng: Those releases are at most 6 weeks old, I'd let this sit for a little while longer, at least until Bazel 10 is officially released.

nicholasjng: (But thanks, TIL.)

docs/bazel.rst
jwnimmer-tri: > Do you think it's worth documenting the free-threaded build here too?

No opinion, really.  I don't use free-threaded myself (yet).  I think developers can probably cargo-cult the flags and preprocessor parts for free-threaded from the existing example, but possibly the toolchain deserves more teaching.

nicholasjng: Fair, let's leave this out for now. Changed the setting to a `bool_flag` as discussed.

nicholasjng:
Think this is ready then @wjakob.

wjakob:
Many thanks to both of you!

